PR #18, opened 2025-12-18
fix: Add macOS support for KiCad bundled Python detection

hexatriene:
## Problem

The MCP server fails to start on macOS with Claude Desktop, showing:

```
[ERROR] Python executable not found: python3
```

### Root Cause

The `findPythonExecutable()` function returns the bare command `python3` on macOS, which then fails the `existsSync()` validation check in `validatePrerequisites()`. 

The issue is that `existsSync('python3')` checks if a file literally named `python3` exists in the current directory—it does **not** search the PATH. This works on Windows because the code returns a full path (`C:\Program Files\KiCad\9.0\bin\python.exe`), but macOS had no equivalent platform-specific detection.

Additionally, setting `PATH` in the Claude Desktop config's `env` section doesn't help because `existsSync()` ignores environment variables entirely.

### Option 1: Use `which python3` to resolve the path
- **Pros**: Would find any Python on the system PATH
- **Cons**: Adds subprocess overhead; the resolved Python might not have `pcbnew` available; doesn't match existing code patterns

### Option 2: Skip `existsSync` check for non-absolute paths
- **Pros**: Simple change; would work with any PATH configuration
- **Cons**: Defers error detection; less informative error messages; inconsistent validation behavior

### Option 3: Add macOS-specific detection for KiCad's bundled Python ✓
- **Pros**: Matches existing Windows pattern; uses KiCad's Python which has `pcbnew` built-in; provides full path so validation works; no config needed
- **Cons**: Hardcodes the standard install path (but so does Windows)

## Solution (Option 3)

This PR adds macOS-specific detection that mirrors the existing Windows approach:

1. **Adds macOS bundled Python detection** - Checks `/Applications/KiCad/KiCad.app/Contents/Frameworks/Python.framework/Versions/{version}/bin/python3` for versions 3.9-3.12

2. **Makes `KICAD_PYTHON` cross-platform** - The environment variable override now works on all platforms, not just Windows (provides escape hatch if detection fails)

3. **Adds logging** - Shows which Python was detected for easier debugging

## Testing

Tested on macOS with KiCad installed at the standard location. Server now starts successfully with Claude Desktop using a minimal config:

```json
{
  "mcpServers": {
    "kicad-mcp": {
      "command": "node",
      "args": ["/path/to/KiCAD-MCP-Server/dist/index.js"]
    }
  }
}
```

Logs show:
```
[INFO] Found KiCAD bundled Python at: /Applications/KiCad/KiCad.app/Contents/Frameworks/Python.framework/Versions/3.9/bin/python3
```

### DIFF
--- a/docs/CLIENT_CONFIGURATION.md
+++ b/docs/CLIENT_CONFIGURATION.md
@@ -46,17 +46,29 @@ This guide shows how to configure the KiCAD MCP Server with various MCP-compatib
   "mcpServers": {
     "kicad": {
       "command": "node",
-      "args": ["/Users/YOUR_USERNAME/MCP/KiCAD-MCP-Server/dist/index.js"],
-      "env": {
-        "PYTHONPATH": "/Applications/KiCad/KiCad.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python3.11/site-packages",
-        "NODE_ENV": "production"
-      }
+      "args": ["/Users/YOUR_USERNAME/MCP/KiCAD-MCP-Server/dist/index.js"]
     }
   }
 }
 ```
 
-**Note:** Adjust Python version (3.11) and KiCAD path based on your installation.
+**Note:** For standard KiCad installations in `/Applications/KiCad/`, the server auto-detects KiCad's bundled Python (versions 3.9-3.12). No `PYTHONPATH` configuration is required.
+
+If KiCad is installed in a non-standard location, you can override the Python path:
+
+```json
+{
+  "mcpServers": {
+    "kicad": {
+      "command": "node",
+      "args": ["/Users/YOUR_USERNAME/MCP/KiCAD-MCP-Server/dist/index.js"],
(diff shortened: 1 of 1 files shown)


mixelpixx:
This is excellent work! 🎉

Your analysis of the problem is spot-on - `existsSync()` doesn't check PATH, which is why the bare `python3` command was failing validation on macOS. The solution mirrors our Windows pattern perfectly and unblocks all macOS users.

**What we particularly appreciate:**
- Thorough problem analysis with options considered
- Cross-platform improvement (making `KICAD_PYTHON` work on all platforms)
- Multiple Python version detection (3.9-3.12)
- Clear logging for debugging
- Updated documentation

The formatting changes (single→double quotes) are consistent with our evolving code style, so no worries there.

Merging now - thank you for this critical contribution! 🚀

---

**Credit:** This fix will be mentioned in the next release notes and you're welcome in our contributor community!

PR #88, opened 2026-04-10
Fix/rotate mirror

tnemrap:
## Problem

Three related bugs in schematic mirror/rotate handling:

1. `rotate_schematic_component` with `mirror="x"` or `mirror="y"` always crashed:
   `'NoneType' object has no attribute 'value'`
   (kicad-skip has no API for `(mirror x/y)` on placed symbols)

2. `add_schematic_net_label` injected `(fields_autoplaced yes)` causing
   incorrect visual rendering, and hardcoded `(justify left bottom)`
   regardless of orientation (should be `right` for 180°/270°).

3. `get_schematic_pin_locations` ignored mirror flags — pin coordinates
   were unchanged after mirror_x or mirror_y.

## Fix

**rotate_schematic_component** — rewritten to use `sexpdata` (same
approach as `move_schematic_component`). Two new helpers in `WireDragger`:
- `compute_pin_positions_for_rotation`: computes old/new pin world positions when rotation/mirror changes
- `update_symbol_rotation_mirror`: updates `(at)` rotation and `(mirror x/y)` token correctly
Connected wires are dragged via the existing `drag_wires` infrastructure.

**add_schematic_net_label** — switched from manual sexpdata construction
to `kicad-skip clone()` from an existing label. Eliminates `fields_autoplaced`
and sets justify correctly based on orientation.

**get_schematic_pin_locations** — added `_get_symbol_transform()` which
reads position, rotation, and mirror flags directly from the `.kicad_sch`
file via sexpdata. `get_pin_location` now uses `WireDragger.pin_world_xy`
for correct mirror+rotate+translate. `get_pin_angle` now applies
mirror-induced angle reflection before adding symbol rotation.

## Test results (verified in Claude Desktop)

| Test | Result |
|------|--------|
| `mirror="x"` — no crash | ✅ |
| `mirror="y"` — no crash | ✅ |
| Remove mirror | ✅ |
| Pin positions after `mirror_x` | ✅ |
| Pin positions after `mirror_y` | ✅ |
| Pin positions after remove | ✅ |

### DIFF
--- a/python/commands/pin_locator.py
+++ b/python/commands/pin_locator.py
@@ -25,6 +25,7 @@ class PinLocator:
         """Initialize pin locator with empty cache"""
         self.pin_definition_cache = {}  # Cache: "lib_id:symbol_name" -> pin_data
         self._schematic_cache: Dict[str, object] = {}  # Cache: path -> loaded Schematic
+        self._sexp_cache: Dict[str, Any] = {}  # Cache: path -> parsed sexpdata (mirror-aware)
 
     @staticmethod
     def parse_symbol_definition(symbol_def: list) -> Dict[str, Dict]:
@@ -216,6 +217,35 @@ class PinLocator:
             pass
         return None
 
+    def _get_symbol_transform(
+        self, schematic_path: Path, symbol_reference: str
+    ) -> Optional[Tuple[float, float, float, bool, bool, str]]:
+        """
+        Read symbol position, rotation, mirror flags, and lib_id directly from the
+        .kicad_sch file via sexpdata (authoritative — not kicad-skip cache, which
+        does not reflect mirror/rotation changes made by rotate_schematic_component).
+
+        Returns (x, y, rotation, mirror_x, mirror_y, lib_id) or None.
+        """
+        import sexpdata as _sexpdata
+        from commands.wire_dragger import WireDragger
+
+        sch_key = str(schematic_path)
+        try:
+            if sch_key not in self._sexp_cache:
+                with open(schematic_path, "r", encoding="utf-8") as f:
+                    self._sexp_cache[sch_key] = _sexpdata.loads(f.read())
+        except Exception as e:
+            logger.error(f"_get_symbol_transform: failed to parse {schematic_path}: {e}")
+            return None
+
+        found = WireDragger.find_symbol(self._sexp_cache[sch_key], symbol_reference)
+        if found is None:
+            return None
+
+        _, sym_x, sym_y, rotation, lib_id, mirror_x, mirror_y = found
+        return sym_x, sym_y, rotation, mirror_x, mirror_y, lib_id
+
     def get_pin_angle(
         self, schematic_path: Path, symbol_reference: str, pin_number: str
     ) -> Optional[float]:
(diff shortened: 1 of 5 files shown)


python/kicad_interface.py
```diff
@@ -2123,47 +2123,70 @@ def _handle_move_schematic_component(self, params: Dict[str, Any]) -> Dict[str,
...
-            for symbol in schematic.symbol:
-                if not hasattr(symbol.property, "Reference"):
+            found = WireDragger.find_symbol(sch_data, reference)
+            if found is None:
+                return {"success": False, "message": f"Component {reference} not found"}
+
+            # Determine new mirror state from current + requested
+            _, _, _, _, _, old_mirror_x, old_mirror_y = found
+            new_mirror_x = (mirror == "x")
+            new_mirror_y = (mirror == "y")
```
Mehanik: Hi! Nice work on the sexpdata rewrite — the wire-dragging integration is a great improvement.

I noticed a potential issue with how the existing mirror state is handled when the caller only wants to rotate (i.e. `mirror` param is `None`).

Currently:
```python
new_mirror_x = (mirror == "x")   # False when mirror is None
new_mirror_y = (mirror == "y")   # False when mirror is None
```

If a symbol already has `(mirror x)` in the file and the user calls `rotate_schematic_component` with just `angle=90` (no `mirror` param), both `new_mirror_x` and `new_mirror_y` will be `False`. This flows into:

1. **`compute_pin_positions_for_rotation`** — computes the new pin positions as if the symbol is unmirrored, so wire endpoints get dragged to the wrong locations.
2. **`update_symbol_rotation_mirror`** — receives `new_mirror=None`, which its docstring explicitly says *"removes any existing mirror token"*. So the existing `(mirror x)` is silently stripped.

The net effect: a rotate-only call on a mirrored symbol will **remove the mirror and mis-drag the wires**.

I see that `old_mirror_x` / `old_mirror_y` are already read from `find_symbol` on the line above but never used. A possible fix:

```python
# Preserve existing mirror state when mirror param is not provided
if mirror is None:
    new_mirror_x = old_mirror_x
    new_mirror_y = old_mirror_y
    effective_mirror = "x" if old_mirror_x else ("y" if old_mirror_y else None)
else:
    new_mirror_x = (mirror == "x")
    new_mirror_y = (mirror == "y")
    effective_mirror = mirror
```

Then pass `effective_mirror` to `update_symbol_rotation_mirror` instead of `mirror`.

Would love to hear your thoughts!

python/commands/pin_locator.py
```diff
@@ -228,19 +217,57 @@ def get_pin_angle(
...
+        .kicad_sch file via sexpdata (authoritative — not kicad-skip cache).
+
+        Returns (x, y, rotation, mirror_x, mirror_y, lib_id) or None.
+        """
+        import sexpdata as _sexpdata
+        from commands.wire_dragger import WireDragger
+
+        try:
+            with open(schematic_path, "r", encoding="utf-8") as f:
+                sch_data = _sexpdata.loads(f.read())
```
Mehanik: This method re-reads and re-parses the entire `.kicad_sch` file from disk on every call, but it's called from `get_pin_location` and `get_pin_outward_angle` which are typically invoked in tight loops (once per pin). For a 16-pin IC that's 16 full file reads and `sexpdata.loads()` parses.

The class already has `self._schematic_cache` used by other methods — it would be great to either reuse that cache here or accept the parsed `sch_data` as a parameter to avoid the repeated I/O.

python/kicad_interface.py
tnemrap: Fixed in commit 6b3d1a7. When mirror is None, the handler now reads the component's existing mirror flags (old_mirror_x / old_mirror_y) from the parsed S-expression and uses those as effective_mirror, so a rotation-only call no longer silently clears a mirrored state.

python/commands/pin_locator.py
tnemrap: Fixed in commit 6b3d1a7. PinLocator now has a _sexp_cache dict (keyed by path) that stores the parsed sexpdata tree. _get_symbol_transform reads from the cache on subsequent calls, so the file is parsed at most once per schematic per PinLocator instance regardless of how many pins are looked up.

tnemrap:
Rebased onto current main (resolving conflicts in wire_dragger.py, wire_manager.py, pin_locator.py, and kicad_interface.py).

Also addressed the two review comments from this PR:

Mirror state preservation — when mirror is None, the handler now reads the component's existing mirror flags from the file and preserves them, so a rotation-only call no longer silently clears a mirrored component.

_get_symbol_transform performance — PinLocator now caches the parsed sexpdata tree in _sexp_cache (keyed by path), so the schematic file is parsed at most once per instance instead of once per pin lookup.

One additional test fix: upstream added from annotations import AnnotationLoader and promoted from commands.wire_manager import WireManager to module-level; the smoke test now stubs annotations and adds python/ to sys.path so all imports resolve in a plain environment.

All 10 tests pass.

Mehanik:
Thank you! Merged

PR #89, opened 2026-04-11
feat: Add macOS automated setup script and update platform documentation

nhebling:
Add `setup-macos.sh`, a shell script that automates Claude Desktop MCP configuration on macOS. The script detects KiCad's bundled Python, resolves PYTHONPATH, generates the correct MCP server config, and safely merges it into the existing Claude Desktop configuration with backup support. Supports `--verify`, `--dry-run`, and `--apply` modes.

Update `README.md` with documentation for the automated setup workflow, including usage examples, parameter reference, and post-setup steps.

Extend `docs/PLATFORM_GUIDE.md` with macOS-specific sections covering installation, path handling, Python environment, troubleshooting, best practices, cross-platform migration, and support resources.

Update `docs/STATUS_SUMMARY.md` to reflect the new macOS automation capabilities and bump the last-updated date.

Considerations for potential future changes: 
- Move 'What's new' from README.md to CHANGELOG.md to reduce size of README.md and keep intent clean
- Move 'Archticture' from README.md to ARCHITECTURE.md or docs folder to reduce size of README.md and keep intent clean

### DIFF
--- a/setup-macos.sh
+++ b/setup-macos.sh
@@ -1,0 +1,317 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+MODE=""
+ASSUME_YES=0
+SERVER_NAME="kicad"
+CLAUDE_CONFIG_PATH=""
+
+SCRIPT_NAME="$(basename "$0")"
+
+usage() {
+  cat <<EOF
+Usage:
+  $SCRIPT_NAME --verify [--name NAME] [--claude-config PATH]
+  $SCRIPT_NAME --dry-run [--name NAME] [--claude-config PATH]
+  $SCRIPT_NAME --apply [--name NAME] [--claude-config PATH] [--yes]
+
+Options:
+  --verify               Check prerequisites and print detected paths
+  --dry-run              Show config and merged Claude Desktop config without writing
+  --apply                Write/update Claude Desktop config
+  --yes                  Do not prompt before writing (only with --apply)
+  --name NAME            MCP server name (default: kicad)
+  --claude-config PATH   Path to Claude Desktop config file
+                         (default: ~/.config/Claude/claude_desktop_config.json)
+EOF
+}
+
+# --- Terminal formatting ---
+if [[ -t 1 ]]; then
+  BOLD=$'\033[1m'
+  DIM=$'\033[2m'
+  RESET=$'\033[0m'
+  GREEN=$'\033[32m'
+  YELLOW=$'\033[33m'
+  RED=$'\033[31m'
+  CYAN=$'\033[36m'
+else
+  BOLD="" DIM="" RESET="" GREEN="" YELLOW="" RED="" CYAN=""
+fi
+
+SYM_OK="${GREEN}✓${RESET}"
+SYM_WARN="${YELLOW}⚠${RESET}"
+SYM_FAIL="${RED}✗${RESET}"
+
+fail() { echo "${SYM_FAIL} ${RED}Error:${RESET} $1" >&2; exit 1; }
+info() { echo "${SYM_OK} $1"; }
+warn() { echo "${SYM_WARN} ${YELLOW}$1${RESET}"; }
+
+section() {
+  echo
+  echo "${DIM}────────────────────────────────────────────────────${RESET}"
+  echo "${BOLD}${CYAN}$1${RESET}"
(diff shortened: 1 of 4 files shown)


setup-macos.sh
```diff
@@ -0,0 +1,317 @@
...
+  $SCRIPT_NAME --apply [--name NAME] [--claude-config PATH] [--yes]
+
+Options:
+  --verify               Check prerequisites and print detected paths
+  --dry-run              Show config and merged Claude Desktop config without writing
+  --apply                Write/update Claude Desktop config
+  --yes                  Do not prompt before writing (only with --apply)
+  --name NAME            MCP server name (default: kicad)
+  --claude-config PATH   Path to Claude Desktop config file
+                         (default: ~/.config/Claude/claude_desktop_config.json)
```
Kletternaut: The default Claude Desktop config path shown here is the Linux path. On macOS it should be ~/Library/Application Support/Claude/claude_desktop_config.json.

setup-macos.sh
```diff
@@ -0,0 +1,317 @@
...
+      fail "Unknown argument: $1"
+      ;;
+  esac
+done
+
+[[ -n "$MODE" ]] || { usage; exit 1; }
+[[ -n "$SERVER_NAME" ]] || fail "Server name must not be empty"
+
+if [[ -z "$CLAUDE_CONFIG_PATH" ]]; then
+  CLAUDE_CONFIG_PATH="$HOME/.config/Claude/claude_desktop_config.json"
```
Kletternaut: Same issue here: this should be $HOME/Library/Application Support/Claude/claude_desktop_config.json for macOS. This is the functional bug — the script will write to the wrong location or miss an existing config on most macOS systems.

setup-macos.sh
```diff
@@ -0,0 +1,317 @@
...
+}
+
+if [[ "$MODE" == "verify" ]]; then
+  show_detected
+  section "Proposed Claude Desktop entry ('$SERVER_NAME')"
+  echo "$CONFIG_FRAGMENT_JSON"
+  exit 0
+fi
+
+MERGE_RESULT="$(merge_config)" || {
```
Kletternaut: Small issue: when merge_config fails, $MERGE_RESULT will be empty under set -e, so the error message is lost silently. Suggested fix:

bash
MERGE_RESULT="$(merge_config 2>&1)" || {
  echo "$MERGE_RESULT"
  exit 1
}

setup-macos.sh
```diff
@@ -0,0 +1,317 @@
...
+info "Claude Desktop configuration updated successfully."
+
+echo
+echo "${BOLD}Next steps:${RESET}"
+echo "  1. Fully quit Claude Desktop"
+echo "  2. Reopen Claude Desktop"
+echo "  3. In a new chat, check: + → Connectors"
+echo "  4. Verify with:"
+echo "     Use the ${BOLD}$SERVER_NAME${RESET} MCP server to run ${BOLD}check_kicad_ui${RESET}."
+echo
```
Kletternaut: Minor: the file is missing a trailing newline at the end. POSIX requires text files to end with a newline — most editors add this automatically.

Kletternaut (changes requested):
Hi @nhebling, thanks for contributing this — a macOS setup script is a great addition! 🙏

I've left a few inline comments. The main issue worth fixing before merging is the Claude Desktop config path: the current default (~/.config/Claude/...) is the Linux path. On macOS it should be ~/Library/Application Support/Claude/claude_desktop_config.json. This would cause the script to write to the wrong location on most macOS systems.

The other points are minor. Looking forward to the fix!

setup-macos.sh
nhebling: Yes, you are correct. I struggled to get it running on my machine and then today tried this approach and it worked. But might be that I initially did not restart Claude completly and sth. was still running. Today there was an update to the Desktop app, think this might have then caused to the false assumption. But thanks for pointing it out.

setup-macos.sh
nhebling: Fixed

setup-macos.sh
nhebling: Yes, indeed. Fixed.

setup-macos.sh
nhebling: Fixed.

PR #100, opened 2026-04-14
fix: ERC handler fails on KiCad 9 schematics

LeahArmstrong:
## Summary

`run_erc` silently returns 0 violations on KiCad 9 schematics with sub-sheets. Two root causes:

- **Non-zero exit code treated as failure** — `kicad-cli sch erc` returns a non-zero exit code when violations exist (expected behavior), but the handler treated this as a command error and returned `success: false` before reading the JSON output.
- **KiCad 9 JSON structure change** — KiCad 9 nests violations under `sheets[].violations` instead of the top-level `violations[]` array used by KiCad 8. The handler only read the top-level array, so all sub-sheet violations were missed.

## Changes

**`python/kicad_interface.py`** (14 lines changed)
- Check for output file existence/size instead of exit code to determine failure
- Collect violations from both top-level `violations[]` and `sheets[].violations`
- Both fixes are backward-compatible with KiCad 8

**`tests/test_erc_handler.py`** (new, 248 lines)
- 6 unit tests covering:
  - Non-zero exit code with valid JSON output succeeds
  - Zero exit code with no violations succeeds
  - Missing output file correctly fails
  - KiCad 9 `sheets[].violations` structure parsed correctly
  - KiCad 8 top-level `violations[]` still works
  - Mixed top-level + sheets violations both collected

## Test plan

- [x] `pytest tests/test_erc_handler.py` — 6/6 pass
- [x] Tested against a real KiCad 9 hierarchical schematic (7 sheets, 293 violations detected — previously returned 0)

### DIFF
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -2664,11 +2664,14 @@ class KiCADInterface:
 
                 result = subprocess.run(cmd, capture_output=True, text=True, timeout=120)
 
-                if result.returncode != 0:
-                    logger.error(f"ERC command failed: {result.stderr}")
+                # kicad-cli returns non-zero when ERC violations are found —
+                # this is normal, not an error.  Only fail when no JSON was
+                # produced (genuine CLI failure).
+                if not os.path.exists(json_output) or os.path.getsize(json_output) == 0:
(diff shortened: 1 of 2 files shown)


Mehanik (approved):
Great work, thank you!

PR #101, opened 2026-04-15
chore: align Python version floor with KiCad's macOS bundle (3.9)

OmsAhmad:
Mostly cleanup for running on macOS after #89. KiCad ships Python 3.9 on macOS (both v9 and v10), and sicne `pcbnew.so` is compiled against that interpreter the previous `>=3.10` floor was unreachable on macOS. The docs referenced nonexistent `python3.11/site-packages` paths. Lowered the floor and corrected the docs. No source changes needed and all  files in `python/` already compile clean on 3.9. Also add `.venv/` to `.gitignore`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-24.04, ubuntu-22.04]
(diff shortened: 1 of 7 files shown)


Mehanik (approved):
LGTM

PR #102, opened 2026-04-15
fix: resolve nets when labels are placed directly at pin endpoints

ffindog:
## Problem

`generate_netlist` and `list_schematic_nets` return 0 nets when a schematic has net labels placed directly at component pin endpoints (without wire segments).

This is valid KiCad schematic style — eeschema connects a net label to a pin if the label anchor is at the pin endpoint, no wire required. The MCP server's `get_net_connections()` in `connection_schematic.py` however required wire segments to exist before it would match pins, causing it to silently drop all nets.

## Root cause

`get_net_connections()` built `connected_wire_points` exclusively from wire endpoints, then returned early if that set was empty:

\`\`\`python
if not connected_wire_points:
    return connections  # always empty for label-at-pin schematics
\`\`\`

## Fix

Build `all_match_points` as the union of wire endpoints **and** label positions. Pin matching uses `all_match_points`, so a label placed directly at a pin endpoint is detected whether or not a wire exists.

\`\`\`python
all_match_points = connected_wire_points | {
    (p[0], p[1]) for p in net_label_positions
}
\`\`\`

Also removed the early return when the schematic has no `wire` attribute — we now continue with label positions as the sole match points rather than bailing out.

Wired schematics are completely unaffected.

## Tests added

`tests/test_label_at_pin_net_connections.py` — 11 unit tests covering:

| Class | Tests |
|---|---|
| `TestLabelAtPinNoWire` | Core fix: label at pin finds connection, multiple pins, tolerance boundary |
| `TestLabelViaWire` | Regression: wired connections still work, no false positives |
| `TestMixedStyles` | Both wired and direct labels on the same net |
| `TestEdgeCases` | Unknown net, no labels, no wire attribute, template symbols skipped |

All 11 pass. No regressions in `test_net_connectivity.py` or `test_net_label_pin_snapping.py` (50 total).

## Python code quality

Verified manually (Black, isort, flake8, mypy all clean). The pre-commit `local` npm hook fails to install on Windows due to a MobaXterm shell environment issue unrelated to this change — no TypeScript files were modified.


Mehanik (approved):
Thank you, nice catch!

PR #103, opened 2026-04-15
fix: correct pin location, symbol reference dedup, and ERC violation parsing

tecnovel:
## What                                                   
                                                                                                              
  Three bugs in the schematic component placement and pin location pipeline that caused `connect_to_net`, wire
   stubs, and net labels to be placed at wrong coordinates (or not at all).
                                                                                                              
  ## Root Causes & Fixes                                    

  ### 1. `component_schematic.py` — symbols saved with `R1_` instead of `R1`                                  
   
  kicad-skip's `clone()` already appends the cloned raw element to the schematic tree. The subsequent         
  `schematic.symbol.append(new_symbol)` triggered `NamedCollection.append`, which detected the reference as
  already registered (from the `elementRename` callback fired when setting `property.Reference.value`) and    
  silently renamed it `R1_`. Every downstream pin lookup then failed with *"Symbol R1 not found in 
  schematic"*.

  **Fix:** removed the redundant `append()` call — the clone is already in the tree.                          
   
  ### 2. `pin_locator.py` — pin positions wrong for unrotated symbols, rotation results wrong for rotated ones
                                                            
  `lib_symbols` embedded in `.kicad_sch` files follow the library y-up convention (y increases upward, same as
   `.kicad_sym` files). `get_pin_location` was reading the y coordinate directly as schematic y-down, placing 
  every pin on the wrong side of the symbol body (pin 1 of Device:R at y=103.81 instead of y=96.19 for a
  symbol centered at y=100). This also caused rotated pin positions to be mirrored.                           
                                                            
  **Fix:** negate `pin_rel_y` before applying rotation — `pin_rel_y = -pin_data["y"]` — matching KiCad's own  
  transform order, which is already implemented correctly in `_transform_local_point` in
  `schematic_analysis.py`.                                                                                    
                                                            
  ### 3. `pin_locator.py` — `.rstrip("_")` guard in all reference lookups

  Defensive fix: strips any trailing `_` from `property.Reference.value` before comparing, guarding against   
  edge cases where kicad-skip writes a trailing underscore.

  ### 4. `kicad_interface.py` — ERC violations not parsed in KiCad 8+
                                                                                                              
  `kicad-cli sch erc --format json` changed its output schema in KiCad 8: violations are now nested under 
  `sheets[].violations` rather than at the top-level `violations` key. The old code did                       
  `erc_data.get("violations", [])` and got an empty list on any modern KiCad install, silently reporting zero
  ERC errors regardless of actual issues.                                                                     
                                                            
  **Fix:** collect violations by iterating `erc_data["sheets"][*]["violations"]` first, with a fallback to the
   top-level `violations` key for older kicad-cli versions.
                                                                                                              
  ### 5. `pin_locator.py` self-test — wrong template                                                          
   
  The standalone test was using `template_with_symbols_expanded.kicad_sch`, which only contains `lib_symbols` 
  definitions and has no placed `_TEMPLATE_*` instances to clone from. Switched to                            
  `template_with_symbols.kicad_sch`.
                                                                                                              
  ## Test Plan                                              

  - [x] `python python/commands/pin_locator.py` — pin 1 of R1 (rotation 0°) now at y=96.19 (above center); pin
   1 of C1 (rotation 90°) at x=153.81 (right of center)                                                       
  - [x] `pytest tests/ python/tests/` — 466 tests pass
  - [x] `black --check` and `mypy --ignore-missing-imports` — no issues on changed files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -848,9 +848,9 @@
       }
     },
     "node_modules/brace-expansion": {
-      "version": "1.1.14",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.14.tgz",
-      "integrity": "sha512-MWPGfDxnyzKU7rNOW9SP/c50vi3xrmrua/+6hfPbCS2ABNWfx24vPidzvC7krjU/RTo235sV776ymlsMtGKj8g==",
+      "version": "1.1.13",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.13.tgz",
+      "integrity": "sha512-9ZLprWS6EENmhEOpjCYW2c8VkmOvckIJZfkr7rBW6dObmfgJ/L1GpSYW5Hpo9lDz4D1+n0Ckz8rU7FwHDQiG/w==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
--- a/python/commands/component_schematic.py
+++ b/python/commands/component_schematic.py
@@ -255,8 +255,9 @@ class ComponentManager:
             # Generate new UUID
             new_symbol.uuid.value = str(uuid.uuid4())
 
-            # Append to schematic
-            schematic.symbol.append(new_symbol)
+            # NOTE: clone() already inserts the raw element into the schematic tree.
+            # Calling schematic.symbol.append() again causes NamedCollection to detect
+            # the reference as "taken" and rename it to "R1_" (trailing underscore).
             logger.info(f"Successfully added component {reference} to schematic")
 
             return new_symbol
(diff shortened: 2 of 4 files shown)


package-lock.json
```diff
@@ -848,9 +848,9 @@
       }
     },
     "node_modules/brace-expansion": {
-      "version": "1.1.14",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.14.tgz",
-      "integrity": "sha512-MWPGfDxnyzKU7rNOW9SP/c50vi3xrmrua/+6hfPbCS2ABNWfx24vPidzvC7krjU/RTo235sV776ymlsMtGKj8g==",
+      "version": "1.1.13",
```
Mehanik, line 851: This change probably is not necessary. Could you please revert it?

Mehanik: I fixed this and also fixed merge conflicts in this branch: `fix/pr103-pin-locator-and-component-rebase`
Hope it would help!

tecnovel:
Thanks for you help, should be fixed and ready to merge now

PR #117, opened 2026-04-19
fix: add_label falls back gracefully on sub-sheet schematics

thesamprice:
## Summary

- `WireManager.add_label` returned `False` with a logged error whenever the schematic had no `(sheet_instances)` block
- Sub-sheets in hierarchical KiCad designs never have `(sheet_instances)` — only the top-level schematic does
- This made it impossible to add any net label (local, global, or hierarchical) to any sub-sheet

## Fix

When `sheet_instances_index` is `None`, fall back to `len(sch_data)` as the insertion index, which appends before the final closing paren of `(kicad_sch ...)`.

## Test plan

- [x] Add a net label to a top-level schematic → still works
- [x] Add a net label to a sub-sheet schematic → previously returned False, now succeeds
- [x] Verify the output file is valid KiCad s-expression (parseable by KiCad)

### DIFF
--- a/python/commands/wire_manager.py
+++ b/python/commands/wire_manager.py
@@ -316,8 +316,9 @@ class WireManager:
                     break
 
             if sheet_instances_index is None:
-                logger.error("No sheet_instances section found in schematic")
-                return False
+                # Sub-sheets in hierarchical designs don't have (sheet_instances).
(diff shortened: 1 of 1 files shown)


Mehanik:
Thank you!

PR #124, opened 2026-04-21
fix: locate placed symbols when (lib_name) precedes (lib_id)

vdawger:
## Summary

Fixes a bug where placed schematic components were silently invisible to `get_schematic_component`, `edit_schematic_component`, `delete_schematic_component`, `set_schematic_component_property`, and `remove_schematic_component_property` whenever KiCad serialised the symbol with a `(lib_name "...")` child before `(lib_id "...")`.

This is the format KiCad uses for **rescued** or **locally-customised** library entries, e.g.:

```
(symbol
  (lib_name "RESISTOR_0603_4")
  (lib_id "MF_Passives:RESISTOR_0603")
  (at 132.08 44.45 90)
  ...
  (property "Reference" "R7" ...)
  ...)
```

The user-visible symptom is `Failed to set property 'whatever' on R1: Component 'R1' not found in schematic` (and the analogous "not found" message from `edit` / `delete` / `get`) even though the component is plainly present in the schematic and reachable through the `list_schematic_components` and IPC paths.

## Root cause

Three handlers in `python/kicad_interface.py` (`_handle_get_schematic_component`, `_handle_edit_schematic_component`, `_handle_delete_schematic_component`) all use the same regex to find the start of a placed symbol block:

```python
pattern = re.compile(r'\(symbol\s+\(lib_id\s+"')
```

That pattern requires `(lib_id` _immediately_ after `(symbol`, so any block that opens `(symbol (lib_name "..."))` never matches and the lookup silently fails.

A second, related bug in the parent-position lookup uses an equally-strict regex:

```python
re.search(r'\(symbol\s+\(lib_id\s+"[^"]*"\s*\)\s+\(at\s+...', block_text)
```

On `(lib_name)`-first symbols this regex doesn't match, so the parent origin silently falls back to `(0, 0)` and any newly-added custom property is anchored at the schematic origin instead of next to the component.

The bug pre-dates the custom-properties feature merged in #123: `edit_schematic_component`'s footprint/value/reference rewriting and `delete_schematic_component` were affected too.

## Fix

- Relax the symbol-block opening pattern from `\(symbol\s+\(lib_id\s+"` to `\(symbol\s+\(` — match _any_ opening paren after `(symbol`. The existing `(lib_symbols ...)` range check already excludes library-definition symbols (which use the `(symbol "name" ...)` quoted-string form, not a paren).
- Read the symbol's origin from the first `(at ...)` inside the matched block. KiCad always writes the symbol's own `(at)` before any `(property ...)` children, regardless of `(lib_name)` / `(lib_id)` ordering.

## Test plan

- [x] Added `TestLibNameBeforeLibIdOrdering` (7 regression tests) in `tests/test_schematic_component_properties.py` exercising:
  - `get_schematic_component` finds a `(lib_name)`-first symbol and reports the correct symbol-origin position
  - `set_schematic_component_property` adds a property to the same symbol
  - `edit_schematic_component` with a `properties` dict adds multiple properties
  - `remove_schematic_component_property` removes a property from the same symbol
  - newly-added properties anchor to the symbol origin (`132.08, 44.45`) instead of `(0, 0)`
  - the pre-existing `edit_schematic_component` `value` / `footprint` rewriting works on the same symbol
  - `delete_schematic_component` removes the symbol and a follow-up `get` reports it as not found
- [x] All 39 tests in `tests/test_schematic_component_properties.py` pass
- [x] Full `pytest tests/` is green except for 3 pre-existing failures in `tests/test_pin_locator_and_component.py::TestPinLocatorYAxisNegation` that also fail on `main` (unrelated to this change)
- [x] `black`, `isort`, `prettier`, `flake8`, `mypy`, and `eslint` pre-commit hooks all pass

## Notes

- `isort` also fixed a pre-existing import-ordering issue at the top of `python/kicad_interface.py` (`from annotations import AnnotationLoader` was below `from schemas...`); the diff includes that one-line cleanup. `flake8` and `black` were already happy on `main`.
- This PR only touches the affected handlers in `python/kicad_interface.py`, the regression tests, and a `### Bug Fixes` entry in `CHANGELOG.md` under `[Unreleased]`.

### PR Checklist

- [x] All tests pass locally (modulo pre-existing failures unrelated to this change)
- [x] New tests added for new features
- [x] Documentation updated (CHANGELOG)
- [x] Commit messages follow convention (`fix:` per Conventional Commits)
- [x] No merge conflicts
- [ ] CI/CD pipeline passes (will verify after PR is opened)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- **Schematic symbol lookup**: `get_schematic_component`,
+  `edit_schematic_component`, `set_schematic_component_property`,
+  `remove_schematic_component_property`, and `delete_schematic_component`
+  no longer fail with `Component '<ref>' not found in schematic` when the
+  placed symbol uses KiCad's rescued / locally-customised serialisation
+  form `(symbol (lib_name "...") (lib_id "...") ...)`. The block-matching
+  regex now accepts any opening paren after `(symbol`, and the
+  parent-position lookup uses the first `(at ...)` inside the symbol
+  block, so newly-added properties anchor to the symbol origin instead of
+  silently falling back to `(0, 0)`. Added 7 regression tests reproducing
+  the failure on a real-world user schematic.
+
 ### New MCP Tools
 
 - `set_schematic_component_property` — Add or update a single custom property
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -15,11 +15,11 @@ import traceback
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
+from annotations import AnnotationLoader
 from resources.resource_definitions import RESOURCE_DEFINITIONS, handle_resource_read
 
 # Import tool schemas, resource definitions, and IPC API annotations
 from schemas.tool_schemas import TOOL_SCHEMAS
-from annotations import AnnotationLoader
 
 _annotation_loader = AnnotationLoader()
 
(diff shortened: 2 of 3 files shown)


Mehanik:
Thank you!

PR #127, opened 2026-04-23
fix: improve JLCPCB downloader robustness and disk usage

matthewellis:
## Problem

The downloader was broken: the upstream jlcparts database now spans `cache.z20` (and potentially more volumes in the future), but the script hardcoded the range at `cache.z01`–`cache.z19`. This caused 7z to report a missing volume and fail extraction entirely.

## Changes

- **Dynamic volume discovery**: replaces the hardcoded part list with a loop up to `MAX_PARTS=30` that stops on a 404, so the script automatically handles any number of upstream volumes without code changes.
- **`curl_download()` helper**: extracts the curl invocation into a single reusable function with the `-f` flag so 4xx/5xx responses are treated as errors rather than silently writing error pages to disk.
- **Automatic cleanup**: the `jlcparts_cache/` directory is removed after MCP format conversion completes, freeing ~12 GB of intermediate storage.
- **Type hints**: return type annotations added to all function signatures to comply with project style guidelines.

## Test plan

- [ ] Run `python3 download_jlcpcb.py` from scratch and verify all volumes including `cache.z20` are downloaded
- [ ] Verify archive volumes are removed after extraction
- [ ] Verify `data/jlcparts_cache/` is removed after conversion
- [ ] Re-run and confirm an already-converted DB is detected and skipped

### DIFF
--- a/download_jlcpcb.py
+++ b/download_jlcpcb.py
@@ -39,7 +39,7 @@ def curl_download(url: str, dest: Path) -> bool:
     return result.returncode == 0
 
 
-def download_files():
+def download_files() -> bool:
     """Download split archive volumes (z01..zNN) then cache.zip, stopping volumes at 404 or MAX_PARTS."""
     print("Downloading jlcparts database (~421MB)...")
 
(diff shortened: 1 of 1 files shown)


Mehanik:
Thank you!

PR #128, opened 2026-04-23
chore(jlcpcb): bump cache parts range to include cache.z20

7tobias:
yaqwsx/jlcparts has added cache.z20 to its pre-built cache; the old range(1, 20) stops at cache.z19 and the extracted archive ends up truncated. Extend to range(1, 21).


7tobias:
https://github.com/mixelpixx/KiCAD-MCP-Server/pull/127 seems to fix this nicely

7tobias:
https://github.com/mixelpixx/KiCAD-MCP-Server/pull/127 seems to fix this nicely

PR #129, opened 2026-04-23
fix(routing): migrate create_netclass to NET_SETTINGS API (KiCad 8+)

7tobias:
board.GetNetClasses() returns a netclasses_map in KiCad 8+ that no longer exposes .Find() / .Add(). Calling create_netclass therefore raises AttributeError on any current KiCad install.

Switch to the NET_SETTINGS API:
  ds = board.GetDesignSettings(); ns = ds.m_NetSettings
  ns.SetNetclass(name, netclass)
  ns.SetNetclassPatternAssignment(name, pattern)

Net assignment is now pattern-based and lives in .kicad_pro (as the GUI does it), not directly in the netclass object. Also rename the uVia getters/setters (GetMicroViaDiameter -> GetuViaDiameter, same for the drill and the setters) to match the current SWIG binding.

### DIFF
--- a/python/commands/routing.py
+++ b/python/commands/routing.py
@@ -1044,17 +1044,12 @@ class RoutingCommands:
                     "errorDetails": "name parameter is required",
                 }
 
-            # Get net classes
-            net_classes = self.board.GetNetClasses()
+            # Use NET_SETTINGS API (KiCad 8+)
+            ds = self.board.GetDesignSettings()
+            ns = ds.m_NetSettings
 
-            # Create new net class if it doesn't exist
-            if not net_classes.Find(name):
-                netclass = pcbnew.NETCLASS(name)
-                net_classes.Add(netclass)
-            else:
-                netclass = net_classes.Find(name)
-
-            # Set properties
+            # Create netclass and set properties
+            netclass = pcbnew.NETCLASS(name)
             scale = 1000000  # mm to nm
             if clearance is not None:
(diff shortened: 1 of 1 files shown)


mixelpixx:
Thanks for digging into this — the NET_SETTINGS API is genuinely the right long-term shape for KiCad 10 (pattern assignments stored in `.kicad_pro` is how the GUI does it), and the rename to `GetuViaDiameter` matches the K10 SWIG bindings.

I'm going to close this PR, though, in favour of #144 — which takes a defensive-shim approach (`hasattr` probe for `NETCLASSES.Find/.Add` vs `netclasses_map`, plus `_safe_get`/`_safe_set` helpers that skip missing methods rather than raise). The trade-off is:

- This PR (#129) drops backward compat with KiCad 6/7. We still have users on older KiCad versions, and the rest of the recently-merged K10 compat work (#157, #160, #144) uses the defensive pattern, so I'd like to keep that consistent.
- #144 also fixes an orthogonal `traceWidth`/`trackWidth` schema-alias bug in the same file.

That said, your point about the new pattern API is the right direction for proper K10 persistence — net classes created via the MCP today don't actually save to `.kicad_pro`, so they're effectively session-scoped. I'm filing a follow-up issue for the pattern-assignment persistence work and crediting your PR there. Would love to have you reopen against the new issue when it's time to add proper persistence on K10.

Closing this PR. Thanks again for the analysis.

PR #133, opened 2026-04-23
fix: remove duplicate Y-axis negation in PinLocator.get_pin_location

johndev7000:
## Summary

`PinLocator.get_pin_location()` applied the symbol-to-schematic Y-axis flip
twice in sequence (two identical `pin_rel_y = -...` blocks with matching
comments), cancelling out. Every pin's Y coordinate was mirrored about the
symbol placement Y.

For symmetric passives (R, C) the bug is invisible — pin 1 and pin 2 are
electrically interchangeable, so attributing one to the other's location
still produces a working net. For ICs with non-equivalent pins (power pins,
opamp inputs, etc.), tools that route through PinLocator —
`connect_to_net`, `add_schematic_connection`, `add_schematic_net_label` with
`componentRef`+`pinNumber` — placed connections at the mirror-flipped pin.

This is a regression: PR #103 originally fixed the y-negation; the redundant
second block was added subsequently and cancelled it out.

## Root cause

```python
pin_rel_x = pin_data["x"]
pin_rel_y = -pin_data["y"]       # first negation (correct)

logger.debug(...)

# lib_symbols uses y-up; schematic uses y-down
pin_rel_y = -pin_rel_y           # second negation — cancels the first
```

`python/commands/schematic_analysis.py` (`_transform_local_point`) and
`python/commands/wire_connectivity.py` both do the same transform correctly
with a single negation. This PR makes `pin_locator.py` consistent with them.

## Evidence

Stock `Device:R` placed at `(111.76, 83.82)`, no rotation, no mirror:

| | Symbol pin-y | Pre-fix `get_pin_location` y | Post-fix y | `kicad-cli generate_netlist` ground truth |
|--|--|--|--|--|
| Pin 1 | +3.81 | 87.63 | **80.01** | Pin 1 is electrically at y=80.01 ✓ |
| Pin 2 | −3.81 | 80.01 | **87.63** | Pin 2 is electrically at y=87.63 ✓ |

Same pattern confirmed on an IC (`Power_Management:TPS2592xx`) with pins at
symbol y ±16.51 and −12.7 — the mirrored reports led to a VIN↔GND misroute
when users mixed MCP-driven placement with manual eeschema edits.

## Test

Adds `tests/test_pin_locator_y_flip.py` — places a stock resistor and a
rotated capacitor, asserts pin coordinates match KiCad's render convention.
Both cases fail against pre-fix code.

## Scope

One-line surgical fix (delete the duplicate negation block) plus the test.
No API changes. Safe for all existing callers — every internal consumer of
PinLocator (`connect_to_net`, `add_schematic_connection`,
`add_schematic_net_label`) gains correctness on non-symmetric pins without
any call-site changes.

### DIFF
--- a/python/commands/pin_locator.py
+++ b/python/commands/pin_locator.py
@@ -384,9 +384,6 @@ class PinLocator:
 
             logger.debug(f"Pin {pin_number} relative position: ({pin_rel_x}, {pin_rel_y})")
 
-            # lib_symbols uses y-up; schematic uses y-down
-            pin_rel_y = -pin_rel_y
-
             # Mirror in local coords after y-negate (KiCad transform order)
             # mirror_x = flip across X axis → negate y
             # mirror_y = flip across Y axis → negate x
--- a/tests/test_pin_locator_y_flip.py
+++ b/tests/test_pin_locator_y_flip.py
@@ -1,0 +1,112 @@
+"""
+Regression test for the symbol-to-schematic Y-axis flip in PinLocator.
+
+Before the fix, pin_locator.py's get_pin_location() negated pin_data["y"]
+twice in sequence (two identical blocks, one commented "Negate y here before
+rotation" and a second commented "lib_symbols uses y-up; schematic uses y-down"
+doing the exact same flip). The double-negation cancelled out, leaving pin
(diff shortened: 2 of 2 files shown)


Mehanik:
Thank you!

PR #144, opened 2026-05-01
fix: net-class API compatibility for KiCad 10 and trackWidth alias

inktomi:
## Summary

`create_netclass` and downstream net-class edits crash on KiCad 10 because they call removed APIs. Three closely-related issues bundled because they all live in `routing.py` and surfaced together while building net classes for an aerospace PDB project:

1. **Legacy `NETCLASSES.Find/.Add` calls** — KiCad 10 stores net classes via the `netclasses_map` dict-style API. The old paths simply don't exist on a 10.x BOARD.
2. **Missing NETCLASS getters/setters** — methods like `GetMicroViaDiameter` were removed in KiCad 10. Calling them raised `AttributeError` and aborted any subsequent edit, even on attributes that *did* still exist.
3. **`traceWidth` vs `trackWidth` payload mismatch** — the published schema accepted `traceWidth`, but the handler only read `trackWidth`. Requests using the documented field silently produced no-op netclasses.

## Fixes

- KiCad-version-defensive shim: prefer `netclasses_map`, fall back to `Find/Add` when running against older KiCad.
- `_safe_get` / `_safe_set` helpers so missing getter/setter methods on KiCad 10 NETCLASS objects fail gracefully (skip + log) instead of raising.
- Accept both `traceWidth` and `trackWidth` in the request payload — keeps backwards compat with anyone who started on the old name.

## Known follow-up (out of scope here)

Net classes created via the MCP do **not** persist to `.kicad_pro`. KiCad 10 stores them under `net_settings.classes` in the project file, but the MCP only writes to `.kicad_pcb`. Direct JSON edits to `.kicad_pro` survive because the MCP doesn't touch that section on save. Flagging for a separate follow-up.

## Test plan

- [x] Manual: `create_netclass` on a fresh KiCad 10 PCB no longer raises
- [x] Manual: edits to existing net classes apply where supported on KiCad 10
- [x] Manual: payloads using `traceWidth` produce the expected width
- [ ] Suggest a unit test covering the schema-alias path — happy to add if you want

This is one of three PRs from a working session on a real Class-3 aerospace PDB project; the others are #145 (schematic pin Y-flip) and an upcoming one for multi-line text escaping. Each kept separate so they bisect cleanly.

### DIFF
--- a/python/commands/routing.py
+++ b/python/commands/routing.py
@@ -48,11 +48,20 @@ class RoutingCommands:
                 net = pcbnew.NETINFO_ITEM(self.board, name)
                 self.board.Add(net)
 
-            # Set net class if provided
+            # Set net class if provided — defensive against KiCad 6/7 vs KiCad 9/10 API.
             if net_class:
                 net_classes = self.board.GetNetClasses()
-                if net_classes.Find(net_class):
-                    net.SetClass(net_classes.Find(net_class))
+                resolved = None
+                if hasattr(net_classes, "Find"):
+                    resolved = net_classes.Find(net_class)
+                else:
+                    try:
+                        if net_class in net_classes:
+                            resolved = net_classes[net_class]
+                    except Exception:
+                        resolved = None
+                if resolved is not None:
+                    net.SetClass(resolved)
 
             return {
                 "success": True,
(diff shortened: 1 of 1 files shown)


mixelpixx (approved):
Approving and merging.

This is the right shape of fix for `routing.py`'s net-class handling — defensive `hasattr` probe for `NETCLASSES.Find/.Add` vs the dict-like `netclasses_map`, plus `_safe_get` / `_safe_set` helpers that silently skip missing getters/setters instead of raising. That preserves KiCad 6/7 callers while making KiCad 10 work, matching the same pattern other recently-merged K10 PRs use (#157, #160).

The `traceWidth` / `trackWidth` schema-alias fix is orthogonal but in the same file and clearly the right call — accepting both keeps existing callers working while aligning to the published schema name.

Noted that net classes don't persist to `.kicad_pro` yet — I'm filing that as a separate follow-up issue so the work isn't lost. A different PR (#129) by @7tobias proposed using `m_NetSettings.SetNetclassPatternAssignment` for proper K10 persistence, which is the right long-term direction; I'm closing #129 in favor of this one because the defensive approach preserves backward compat, and capturing the persistence work in the follow-up issue. Thanks!

PR #145, opened 2026-05-01
fix: apply Y-axis flip in WireDragger.pin_world_xy

inktomi:
## Summary

`WireDragger.pin_world_xy` was returning `sym_y + ry` without applying the Y-axis flip that converts library coordinates (Y-up) to schematic coordinates (Y-down). For symmetric two-pin passives (R, C non-polarized) this was invisible because pin 1 and pin 2 are electrically equivalent. For **polarized** parts — electrolytic/polymer caps, TVS diodes, MOSFETs, BJTs — it silently swapped polarity. A label snapped to a polymer cap's positive pin actually landed on the negative pin. That's catastrophic at first power-up.

`tests/test_pin_locator_y_flip.py` was already in the tree with the correct expected coordinates but the matching code fix was never landed; it's been failing on `main`. This PR makes that test pass.

## Fix

Apply the Y-flip before rotation, matching eeschema's actual transformation order:

> `mirror in lib space → Y-flip to screen → rotate → translate`

```python
rx, ry = _rotate(lx, -ly, rotation)   # was: _rotate(lx, ly, rotation)
return sym_x + rx, sym_y + ry
```

## Tests touched

- `tests/test_pin_locator_y_flip.py` — the existing regression now passes.
- `tests/test_move_with_wire_preservation.py` — three tests had baked the *buggy* expected coordinates into their assertions:
  - `TestComputePinPositions::test_resistor_at_origin_no_rotation` — pin-1/pin-2 expected world Ys swap.
  - `TestComputePinPositions::test_resistor_rotated_90` — pin 1 expected at (103.81, 100), not (96.19, 100).
  - `TestSynthesizeTouchingPinWires::test_touching_pin_gap_generates_wire` — the `_make_two_resistors(0, 0, 0, -7.62)` fixture had to flip R2's Y to `+7.62` so the two pins still meet under the corrected formula.

The other ~37 tests in those files pass unchanged.

## End-to-end verification

Verified against a 46-component aerospace PDB schematic outside this repo: all 8 polarized-part pins (4 polymer caps C1/C5/C6/C7, 4 TVS diodes D4/D5/D6/D7) now produce world coordinates that match the labels actually placed in the `.kicad_sch` file. Before the fix, every one of them was off by `2 × lib_pin_y`.

## Test plan

- [x] `pytest tests/test_pin_locator_y_flip.py` — was failing, now passes
- [x] `pytest tests/test_move_with_wire_preservation.py` — 40/40 pass with updated assertions
- [x] Manual end-to-end: snap a label to a polymer cap's pin 1 in a real schematic, confirm KiCad renders it on the correct (positive) lead
- [x] Manual end-to-end: `kicad-cli sch erc` round-trip clean

## Risk

This changes the meaning of `pin_world_xy` outputs project-wide. Every caller (`pin_locator`, `connect_to_net`, `add_schematic_net_label` with snap, `compute_pin_positions` for wire-drag) was internally consistent with the bug, so they all *also* shift to correct positions in lockstep. Worth a careful review for any downstream code path that might have been working against the buggy positions on purpose.

### DIFF
--- a/python/commands/wire_dragger.py
+++ b/python/commands/wire_dragger.py
@@ -152,15 +152,16 @@ class WireDragger:
         """
         Compute the world coordinate of a pin given the symbol transform.
 
-        KiCAD applies mirror first (in local space), then rotation, then translation.
-        mirror_x negates the local X axis; mirror_y negates the local Y axis.
+        Library pins are stored Y-up; the schematic is Y-down. Order matches
+        eeschema: mirror in lib space → Y-flip to screen → rotate → translate.
+        Without the Y-flip, polarized parts get pin 1/pin 2 silently swapped.
         """
         lx, ly = px, py
         if mirror_x:
             lx = -lx
         if mirror_y:
             ly = -ly
-        rx, ry = _rotate(lx, ly, rotation)
+        rx, ry = _rotate(lx, -ly, rotation)
         return sym_x + rx, sym_y + ry
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


Mehanik:
well spotted :+1:

PR #148, opened 2026-05-02
Fix KiCad 10 standard symbol library discovery on Windows

Kulitorum:
## Summary

On a fresh KiCad 10 / Windows install, `search_symbols`, `list_symbol_libraries`, and `add_schematic_component` returned no results for any stdlib symbol (`Device:R`, `RF_Module:ESP32-WROOM-DA`, `Connector_Generic:*`, etc.). Two independent root causes:

- **Hard-coded install paths.** `_find_kicad_symbol_dir` in `library_symbol.py` and `dynamic_symbol_loader.py` only knew about KiCad 8.0/9.0 paths and only consulted `KICAD8_SYMBOL_DIR` / `KICAD9_SYMBOL_DIR`. Added the KiCad 10.0 path and a `KICAD10_SYMBOL_DIR` env var.
- **`sym-lib-table` regex rejected URIs with spaces.** The capture used `[^")\s]+`, but KiCad's default global table on Windows references the stdlib via a quoted URI containing `Program Files`. The regex silently skipped that Table entry, hiding the entire stdlib. Regex now accepts either a quoted value (allowing spaces) or a bare token, applied to `name`, `type`, and `uri`.

## Test plan

- [ ] On Windows + KiCad 10, run `search_symbols` for `Device:R` and confirm a hit.
- [ ] On Windows + KiCad 10, run `list_symbol_libraries` and confirm stdlib libraries appear.
- [ ] On Windows + KiCad 9, confirm prior behavior is unchanged (regex still parses 9.0 tables; old env vars still honored).
- [ ] On Linux/macOS, confirm `/usr/share/kicad/symbols` discovery still works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **KiCad 10 standard symbol libraries are now discoverable**: `search_symbols`,
+  `list_symbol_libraries`, and `add_schematic_component` previously returned no
+  results for stdlib symbols (`Device:R`, `RF_Module:ESP32-WROOM-DA`,
+  `Connector_Generic:*`, etc.) on KiCad 10 installations. Two underlying
+  causes:
+  1. `_find_kicad_symbol_dir` (in `library_symbol.py` and
+     `dynamic_symbol_loader.py`) hard-coded only KiCad 8.0 / 9.0 install
+     paths, so `${KICAD10_SYMBOL_DIR}` and the matching `KICAD10_SYMBOL_DIR`
+     env var never resolved. Added `C:/Program Files/KiCad/10.0/share/kicad/symbols`
+     and `KICAD10_SYMBOL_DIR` env-var support.
+  2. The `sym-lib-table` regex used `[^")\s]+` for the URI capture, which
+     rejects any URI containing a space. KiCad's default global table on
+     Windows references the stdlib via
+     `(lib (name "KiCad") (type "Table") (uri "C:/Program Files/KiCad/10.0/share/kicad/template/sym-lib-table"))`
+     — the space in `Program Files` caused the Table reference to be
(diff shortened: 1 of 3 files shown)


mixelpixx:
Thanks for this — the KiCad 10 stdlib discovery on Windows is a real gap and your two fixes (the K10 paths + env var, and the quoted-URI regex) are exactly right.

I need to ask for a rebase though. PR #164 (merged today) made the same regex fix to `python/commands/dynamic_symbol_loader.py`'s `_resolve_library_from_table` — quoted-URI alternation accepting either `"([^"]+)"` or `([^"\)\s]+)`. That overlap is why GitHub is reporting this PR as CONFLICTING.

What I'd like to merge from this PR after a rebase against current `main`:

1. **`python/commands/library_symbol.py`** — same alternation pattern applied to `_parse_sym_lib_table`. #164 didn't touch this file, so this is unique and valuable. Same fix, different file.
2. **`python/commands/dynamic_symbol_loader.py`** — keep only the `Path("C:/Program Files/KiCad/10.0/share/kicad/symbols")` entry in `find_kicad_symbol_libraries`. Drop the regex hunk in `_resolve_library_from_table` since it's already in `main`.
3. **`python/commands/library_symbol.py::_find_kicad_symbol_dir`** — the K10 path and `KICAD10_SYMBOL_DIR` env var support. Unique and valuable.
4. **CHANGELOG.md** — your entry is still accurate after rebase; just resolve whatever conflict comes from sequential edits.

Once rebased and the diff is just the K10-paths-plus-`library_symbol.py`-regex bits, I'll merge. Thanks!

mixelpixx:
Thanks for this, @Kulitorum — the KiCad 10 symdir discovery was a real gap and your fix was on the right track. To avoid landing two implementations of the same thing, we're consolidating symdir support into #193 (karu2003), which is more complete and actively maintained. Closing this in favor of that — really appreciate you surfacing it and the original fix. If #193 stalls we'll revisit.

PR #149, opened 2026-05-02
Make project-scope sym-lib-table visible to symbol-discovery tools

Kulitorum:
## Summary

`search_symbols`, `list_symbol_libraries`, `list_library_symbols`, and `get_symbol_info` only consulted the global `sym-lib-table`. A library registered with project scope (i.e. an entry in `<project>/sym-lib-table`) was therefore invisible — even right after `open_project` succeeded — so `add_schematic_component` was the only tool that could see it.

The infrastructure to load project-scope tables already existed (`SymbolLibraryManager` accepts `project_path` and `_load_libraries` reads `<project>/sym-lib-table`), but the singleton `SymbolLibraryCommands` was constructed with no project and never refreshed.

## Fix

- `_handle_open_project` / `_handle_create_project` wrappers rebuild `SymbolLibraryCommands.library_manager` scoped to the resolved project directory. Subsequent search/list/info calls see project-scope libraries automatically.
- The four discovery tools also accept an optional `projectPath` parameter (project directory, `.kicad_pro`, `.kicad_pcb`, or `.kicad_sch` path). `SymbolLibraryCommands._derive_project_path` walks up from `schematicPath` / `boardPath` to the project root the same way `_handle_add_schematic_component` does, so stateless callers don't need to call `open_project` first.

## Test plan

- [ ] Register a project-scope library (e.g. `Decenza:NAU7802SGI` via an entry in `<project>/sym-lib-table`).
- [ ] After `open_project`, run `search_symbols` for `NAU7802` and confirm a hit.
- [ ] Without calling `open_project`, call `list_symbol_libraries` with `projectPath=<project dir>` and confirm the project library appears.
- [ ] Without calling `open_project`, call `get_symbol_info` with `symbol=Decenza:NAU7802SGI` and `schematicPath=<project>/foo.kicad_sch` and confirm a hit.
- [ ] Confirm global-only behavior is unchanged when no project context is supplied.
- [ ] Confirm `add_schematic_component` still works (it still constructs its own `DynamicSymbolLoader`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **Project-scope `sym-lib-table` is now visible to symbol-discovery tools**:
+  `search_symbols`, `list_symbol_libraries`, `list_library_symbols`, and
+  `get_symbol_info` previously only consulted the global `sym-lib-table`. A
+  library registered with project scope (i.e. an entry in
+  `<project>/sym-lib-table`) was therefore invisible — even right after
+  `open_project` succeeded — making `add_schematic_component` the only tool
+  that could see it. Two changes:
+  1. `open_project` and `create_project` now rebuild the
+     `SymbolLibraryManager` against the project directory so subsequent
+     search/list/info calls see project-scope libraries automatically.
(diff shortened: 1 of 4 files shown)


Mehanik:
Thanks!

PR #151, opened 2026-05-03
feat: guard SWIG auto-save against external file changes

inktomi:
## Summary

`kicad_interface._auto_save_board()` runs after every board-mutating SWIG command and unconditionally calls `pcbnew.SaveBoard()` with the in-memory `self.board`. If the on-disk `.kicad_pcb` was modified externally between the MCP's `LoadBoard` and `SaveBoard` (KiCad GUI's own save, a `git checkout`, another process), the in-memory state silently overwrites those external changes — and the user has no signal it happened.

This PR adds a content-based guard:

* On `open_project` / `create_project`, record `(mtime_ns, sha256)` of the loaded `.kicad_pcb`.
* Before each auto-save, recompute the signature. If it differs from the recorded one, **refuse the save** and surface a warning on the command result. The in-memory mutation is preserved; the caller is told to reload via `open_project` and re-apply.
* If the signature still matches, copy the existing file to `.mcp-backups/<name>.<timestamp>` (rotating, keeps last 20) before overwriting. Insurance against future failure modes we haven't found yet.
* Refresh the recorded signature after our own writes so subsequent saves aren't falsely flagged.

## Backwards compatibility

* No tool schemas changed — every existing MCP tool keeps its current signature.
* Successful saves return as before, with an additional `autoSave: {saved: true, backup: <path>}` field.
* Refused saves return `success: true` (the in-memory mutation did succeed) plus `warnings: [...]` and `autoSave: {saved: false, diskChangedExternally: true, ...}`. Callers can detect this programmatically.

## Test plan

- [x] 10 new unit tests in `tests/test_auto_save_guard.py`, all passing on Python 3.9. Cover signature math, refusal on external change, backup creation + content, backup rotation, first-save semantics, skip cases.
- [x] Existing test suite: pre-existing failures unchanged (test_get_pin_angle.py PEP 604 syntax, env-specific failures in pin_locator/wire_connectivity/hierarchical_pad_net_map). My change is purely additive — no regressions introduced.
- [x] pre-commit hooks (black, isort, flake8, mypy) all pass.

## Motivation

Caught after a fork-user lost ~480 traces and the entire footprint layout to a silent overwrite during normal MCP use on 2026-05-03. Recovery was only possible because VS Code's local-history extension happened to be running and snapshot the file 3 minutes earlier. A guarded auto-save makes this class of incident loud (warning surfaced to the caller) and locally recoverable (rotating backups in `.mcp-backups/`).

## Notes

* The signature uses `(mtime_ns, sha256)`, not just mtime, so we're robust against filesystems with poor mtime resolution and against tools that touch the file without changing its content.
* `.mcp-backups/` is created next to the `.kicad_pcb`. Users should `.gitignore` it; happy to add a follow-up PR with a sample gitignore snippet in the README if that's preferred.
* Doesn't change the IPC backend code path — that one writes through KiCad's own IPC, which has its own concurrency story.

### DIFF
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -7,11 +7,14 @@ and KiCAD's Python API (pcbnew). It receives commands via stdin as
 JSON and returns responses via stdout also as JSON.
 """
 
+import hashlib
 import json
 import logging
 import os
+import shutil
 import sys
 import traceback
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
@@ -259,6 +262,12 @@ class KiCADInterface:
         """Initialize the interface and command handlers"""
         self.board = None
         self.project_filename = None
+        # On-disk signature (mtime_ns, sha256_hex) of self.board's file as of
+        # last load or successful auto-save.  Used by _auto_save_board() to
+        # detect external modifications and refuse to clobber them.
+        self._board_disk_signature: Optional[Tuple[int, str]] = None
+        # Number of timestamped backups to keep in .mcp-backups/ per board file.
+        self._auto_save_backup_keep = 20
         self.use_ipc = USE_IPC_BACKEND
         self.ipc_backend = ipc_backend
         self.ipc_board_api = None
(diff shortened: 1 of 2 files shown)


Mehanik:
Thank you!

PR #154, opened 2026-05-05
fix(add_mounting_hole): set FPID and restrict NPTH pad to mask layers

inktomi:
## Summary

`BoardOutlineCommands.add_mounting_hole` had two bugs that made the resulting mounting holes both un-editable in the GUI and a source of phantom DRC errors:

1. **Empty FPID** — the footprint was created via `pcbnew.FOOTPRINT(self.board)` with `SetReference`/`SetValue` but no `SetFPID`. It serialised as `(footprint "" ...)` in the .kicad_pcb. KiCad's GUI Move tool refuses to select footprints with no library link, so users couldn't drag the resulting MHs around in the editor.
2. **NPTH pad on copper layers** — the pad inherited the default LSET (`*.Cu` + `*.Mask`) even when `plated:false` was set. With `padDiameter > diameter` that produces phantom copper annular rings on every Cu layer, which fire clearance DRC against neighbouring nets.

## Repro

In any project, call `add_mounting_hole` with `position={x:117,y:84.5,unit:"mm"}, diameter:3.2, padDiameter:3.5`. Open the resulting `.kicad_pcb` in KiCad — the MH is unselectable, and DRC reports phantom `F.Cu` copper pad shorts to neighbouring component pads.

## Changes

- `python/commands/board/outline.py` — set a real FPID via `module.SetFPID(pcbnew.LIB_ID(lib, name))`. Default is `MountingHole:MountingHole_<diameter:g>mm` (e.g. `3.2` → `MountingHole_3.2mm`); a new optional `footprintLibId` param accepts an override (e.g. `MountingHole:MountingHole_3.2mm_M3`). For NPTH, the pad's `LSET` is restricted to `F.Mask` + `B.Mask` only. PTH path is unchanged — the default `*.Cu` + `*.Mask` LSET is correct there.
- `python/schemas/tool_schemas.py` — schema previously advertised `x`/`y`/`diameter` at the top level, but the impl reads `position={x,y,unit}`, `padDiameter`, `plated`. Schema now matches the implementation and exposes the new `footprintLibId` param.
- `tests/test_add_mounting_hole.py` (new) — 7 regression tests covering both fixes.

## Test plan

- [x] `pytest tests/test_add_mounting_hole.py` — 7 / 7 pass
- [x] Manual repro: hand-edited .kicad_pcb to apply the same `(footprint "MountingHole:MountingHole_3.2mm" ...)` and `(layers "*.Mask")` changes — MHs become selectable in the GUI Move tool, and the phantom F.Cu DRC errors disappear.
- [ ] Reviewer: try `add_mounting_hole` in a fresh project to confirm the footprint is selectable in the editor.

## Related (not fixed in this PR)

The IPC backend handler `_ipc_add_mounting_hole` (in `kicad_interface.py`) only creates a `BoardCircle` on `Edge.Cuts` and ignores `plated` / `padDiameter` entirely — it doesn't produce a footprint at all. Different bug, separate PR; the SWIG path is what users actually hit when IPC isn't connected.

### DIFF
--- a/python/commands/board/outline.py
+++ b/python/commands/board/outline.py
@@ -216,6 +216,7 @@ class BoardOutlineCommands:
             diameter = params.get("diameter")
             pad_diameter = params.get("padDiameter")
             plated = params.get("plated", False)
+            footprint_lib_id = params.get("footprintLibId")
 
             if not position or not diameter:
                 return {
@@ -247,6 +248,19 @@ class BoardOutlineCommands:
             module.SetReference(f"MH{next_num}")
             module.SetValue(f"MountingHole_{diameter}mm")
 
+            # Set a real library:name FPID. Without this, the footprint is
+            # written as `(footprint "" ...)` and KiCad's GUI Move tool refuses
+            # to select it (no library link → not draggable in the editor).
+            if not footprint_lib_id:
+                # Strip trailing zeros so 3.2 → "3.2" not "3.20"
+                footprint_lib_id = f"MountingHole:MountingHole_{diameter:g}mm"
+            if ":" in footprint_lib_id:
+                lib_name, fp_name = footprint_lib_id.split(":", 1)
+            else:
+                lib_name = "MountingHole"
+                fp_name = footprint_lib_id
+            module.SetFPID(pcbnew.LIB_ID(lib_name, fp_name))
+
             # Create the pad for the hole
             pad = pcbnew.PAD(module)
             pad.SetNumber(1)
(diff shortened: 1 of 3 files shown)


mixelpixx (approved):
Approving and merging.

Both fixes are real and surgical: `SetFPID(pcbnew.LIB_ID(lib, name))` makes the mounting hole selectable in the GUI Move tool, and restricting the NPTH pad's `LSET` to `F.Mask`/`B.Mask` eliminates the phantom Cu annular rings that were tripping clearance DRC. The schema update is technically a shape change but it's *correcting* a schema that was already inconsistent with the implementation (the impl reads `params["position"]`, not top-level `x`/`y`) — so callers using the old schema were already broken. Good catch on the `%g` formatting for the diameter suffix (`3.0` → `"3"`).

7 regression tests with explicit assertions on `SetAttribute`, `SetLayerSet` construction, and `SetFPID` invocations. The PTH-keeps-default-layers test is exactly the right negative assertion.

Noted that the IPC-backend `_ipc_add_mounting_hole` is a separate bug — leaving that for a follow-up. Thanks!

PR #157, opened 2026-05-06
fix(ipc): guard missing library_link attribute on Footprint (KiCad 9)

gcolonese:
On KiCad 9.0 IPC, some `Footprint.definition` objects lack the
`library_link` attribute, so `list_components` raises `AttributeError`
and returns nothing. Guard the access and fall back to `definition.id`
when present.

## Test plan
- Existing test suite (678 passing).
- Manual: `get_component_list` against a KiCad 9 board no longer raises
  on footprints loaded from the project library.


mixelpixx (approved):
Approving. Defensive `hasattr(definition, "library_link")` guard with fallback to `definition.id` is exactly the right shape for the KiCad 9 IPC variance, and the regression test in `tests/test_ipc_footprint_library_link_guard.py` pins the behavior. Thanks!

PR #158, opened 2026-05-06
fix(ipc): convert inch units to mm in place_component / move_component

gcolonese:
The `_ipc_place_component` and `_ipc_move_component` handlers ignore
the `unit` field in position parameters and pass values straight to
`ipc_board_api`, which expects mm. Inches were placed at 1/25.4 of the
intended position.

Read `position.unit` and convert to mm before calling the IPC backend.

## Test plan
- Existing test suite passes.
- Manual: place a footprint at `{x:1, y:1, unit:"inch"}` and confirm
  the resulting position in pcbnew is 25.4 mm, not 1 mm.

### DIFF
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -4816,9 +4816,15 @@ print("ok")
             position = params.get("position", {})
             x = position.get("x", 0) if isinstance(position, dict) else params.get("x", 0)
             y = position.get("y", 0) if isinstance(position, dict) else params.get("y", 0)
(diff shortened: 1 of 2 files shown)


mixelpixx (approved):
Approving. Small, focused conversion (`x * 25.4`) in `_ipc_place_component` and `_ipc_move_component`, with coverage in `tests/test_ipc_position_inch_conversion.py`. No scope creep. Thanks!

PR #159, opened 2026-05-06
fix(ipc): rotate_component uses absolute angle (matches schema)

gcolonese:
`_ipc_rotate_component` adds the supplied angle to the current rotation,
but the tool schema documents `rotation` as the absolute target angle.
Two consecutive `rotate_component(angle=90)` calls were rotating the
part to 180° instead of leaving it at 90°.

Use `angle % 360` directly. Matches the documented schema and matches
the SWIG-path behavior.

## Test plan
- Existing tests pass.
- Manual: `rotate_component(R1, 90)` twice in a row leaves the part at
  90°, not 180°.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **`rotate_component` now treats `angle` as an absolute target rotation**,
+  matching its schema description. Previously the IPC backend added the
+  supplied angle to the current rotation, so two consecutive
+  `rotate_component(angle=90)` calls would rotate the part to 180° instead
+  of leaving it at 90°. Workflows that relied on the additive behavior will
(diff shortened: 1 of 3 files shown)


mixelpixx (approved):
Approving — the implementation was diverging from the documented schema, and the new tests pin the absolute-angle contract.

One ask before merging: please add a line to `CHANGELOG.md` under the unreleased section noting the behavior change, e.g.:

> Fixed: `rotate_component` now treats `angle` as an absolute target rotation, matching its schema description. Previously the angle was added to the current rotation. Workflows that relied on the additive behavior will need to be updated.

This is technically a fix, but it is observable to existing callers, so we want it called out explicitly.

gcolonese:
Added the CHANGELOG.md entry under \[Unreleased] > Bug Fixes\ in commit 11f66bd, using the wording you suggested. Branch has also been rebased onto upstream/main.

PR #160, opened 2026-05-06
fix(windows): preload cairo-2.dll for 2D board view rendering

gcolonese:
On Windows, `cairocffi` calls `ffi.dlopen('cairo-2')` which only
searches `PATH`. KiCad ships `cairo-2.dll` in `C:\Program Files\KiCad\
9.0\bin` but that directory isn't necessarily on the user's `PATH` when
the MCP server's Python child starts, so 2D view rendering fails with a
DLL load error.

Prepend the KiCad `bin` directory to `PATH` early in
`kicad_interface.py`, before any `cairocffi` import. Falls back through
`PYTHONPATH`, the Python executable's directory, and the default 9.0 /
8.0 install paths.

## Test plan
- Linux/macOS: `sys.platform == "win32"` guard skips the block, no
  behavior change.
- Windows: `get_board_2d_view` no longer fails with `cairo-2` DLL load
  error in a fresh shell.

### DIFF
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -15,6 +15,21 @@ import traceback
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
+# Fix cairo DLL loading on Windows before any cairocffi import.
+# cairocffi uses cffi's ffi.dlopen('cairo-2') which needs the DLL on PATH.
+if sys.platform == "win32":
+    for _bin_dir in [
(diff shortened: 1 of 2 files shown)


mixelpixx (approved):
Approving. Windows-only PATH prepend, gated on `cairo-2.dll` actually existing in a candidate KiCad bin dir, with no remote sources and a regression test. Clean fix for a real Windows pain point. Thanks!

PR #161, opened 2026-05-06
feat(view): save 2D board view to file instead of base64

gcolonese:
`get_board_2d_view` returns the rendered image as base64 in JSON, which
often blows past tool/message size limits on larger boards.

Save the rendered image (PNG/JPG/SVG) next to the `.kicad_pcb` file as
`<board>_2d_view.<ext>` and return the file path. Image-aware tools can
then open the file directly.

This is a response-shape change: callers receive `filePath` instead of
`imageData`. Happy to gate behind an opt-in flag if maintaining the
base64 path is preferred.

## Test plan
- Manual: `get_board_2d_view` with `format="png"` writes
  `<board>_2d_view.png` next to the PCB and returns its path.
- Same for `format="svg"` and `format="jpg"`.

### DIFF
--- a/python/commands/board/view.py
+++ b/python/commands/board/view.py
@@ -124,36 +124,44 @@ class BoardViewCommands:
 
             plotter.ClosePlot()
 
+            # Determine output path next to the PCB file
+            board_dir = os.path.dirname(self.board.GetFileName())
+            board_name = os.path.splitext(os.path.basename(self.board.GetFileName()))[0]
+
             # Convert SVG to requested format
             if format == "svg":
-                with open(temp_svg, "r") as f:
-                    svg_data = f.read()
-                os.remove(temp_svg)
-                return {"success": True, "imageData": svg_data, "format": "svg"}
+                output_path = os.path.join(board_dir, f"{board_name}_2d_view.svg")
+                if os.path.exists(output_path):
+                    os.remove(output_path)
+                os.rename(temp_svg, output_path)
+                return {
+                    "success": True,
+                    "format": "svg",
+                    "filePath": output_path,
+                    "message": f"2D view saved to {output_path}",
+                }
             else:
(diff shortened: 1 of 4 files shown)


mixelpixx (changes requested):
Thanks for tackling this — the inline base64 payload is genuinely a problem for large boards, and saving to disk is the right escape hatch.

I do need to ask for a non-breaking version before merging. As written, the PR removes the `imageData` field from the `get_board_2d_view` response and replaces it with `filePath`. Any existing client expecting the base64 payload (including AI clients integrated against the current schema) will silently break the next time they call this tool.

Could we make this additive instead? Two options that would both work:

1. **Add an opt-in parameter** — e.g. `responseMode: "inline" | "file"`, defaulting to `"inline"` (the current behavior). When set to `"file"`, the response includes `filePath` and omits `imageData`. This preserves the existing contract and lets callers opt into the new behavior explicitly.
2. **Return both fields when small, file-only when large** — keep `imageData` for responses under some size threshold (e.g. 1 MB), and switch to `filePath` only above it. Document the threshold in the tool description.

Option 1 is cleaner and easier to reason about. Happy with either, just not a silent shape change.

Once that's in, the rest of the diff (the `cairosvg.convert(..., write_to=...)` flow, the JPEG save path next to the PCB) looks good.

gcolonese:
Implemented Option 1: added an opt-in `responseMode` parameter (`inline` | `file`, default `inline`). The default restores the original base64 `imageData` response (backward-compatible), while `responseMode=file` writes the image next to the `.kicad_pcb` and returns `filePath`. Both modes share the same rendering path; only response packaging differs. Schema updated in Python and TypeScript, and 5 unit tests cover inline PNG/SVG, file PNG/SVG, and the default-is-inline contract.

PR #163, opened 2026-05-06
feat(components): include bounding box and courtyard in queries

gcolonese:
> Stacked on top of #157. The first commit on this branch is the `library_link` guard from that PR; once #157 merges I'll rebase this onto `main` so the diff drops to just the bounding-box / courtyard commit.

`get_component_list` and `get_component_properties` don't expose any
extent information for placed footprints. Adds:

- `boundingBox` (`min_x`, `min_y`, `max_x`, `max_y`, `width`, `height`,
  `unit`) on every returned component.
- `courtyard` (same shape) on `get_component_properties` SWIG path,
  pulled from `F_CrtYd` / `B_CrtYd` via `GetCourtyard()`.

SWIG path uses `GetBoundingBox()` / `GetCourtyard()` directly. IPC path
tries `board.get_item_bounding_box(fp)` first, then falls back to a
pad-extent-with-margin computation, then falls back to enriching from
the SWIG result. This way bbox is present regardless of which backend
served the query.

Useful for placement clearance checks, auto-layout heuristics, and
collision detection.

## Test plan
- `pytest tests/` passes.
- Manual: `get_component_list` returns `boundingBox` for each component
  on a board.
- Manual: `get_component_properties(R1)` returns both `boundingBox` and
  `courtyard` (when the footprint defines a courtyard layer).

### DIFF
--- a/python/commands/component.py
+++ b/python/commands/component.py
@@ -414,6 +414,38 @@ class ComponentCommands:
             x_mm = pos.x / 1000000
             y_mm = pos.y / 1000000
 
+            # Get bounding box
+            bbox = module.GetBoundingBox()
+            bbox_data = {
+                "min_x": bbox.GetLeft() / 1000000,
+                "min_y": bbox.GetTop() / 1000000,
+                "max_x": bbox.GetRight() / 1000000,
+                "max_y": bbox.GetBottom() / 1000000,
+                "width": (bbox.GetRight() - bbox.GetLeft()) / 1000000,
+                "height": (bbox.GetBottom() - bbox.GetTop()) / 1000000,
+                "unit": "mm",
+            }
+
+            # Try to get courtyard bounds (preferred for placement clearance)
+            courtyard_data = None
+            try:
+                for layer_id in [pcbnew.F_CrtYd, pcbnew.B_CrtYd]:
+                    courtyard = module.GetCourtyard(layer_id)
+                    if courtyard and courtyard.OutlineCount() > 0:
+                        cbox = courtyard.BBox()
+                        courtyard_data = {
+                            "min_x": cbox.GetLeft() / 1000000,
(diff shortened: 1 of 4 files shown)


mixelpixx:
Reviewed — the additive `boundingBox` / `courtyard` fields on `get_component_properties` and `get_component_list` look good (additive shape change, SWIG-then-IPC fallback, with regression tests), and I'm planning to approve it.

One housekeeping item: this PR is stacked on top of #157 — the `library_link` `hasattr` guard hunk in `python/kicad_api/ipc_backend.py` and the `tests/test_ipc_footprint_library_link_guard.py` file are both present in this diff as well. I'm going to merge #157 first, then could you rebase this branch onto current `main` so the duplicate hunks drop out and this PR only contains the bbox/courtyard changes? Once that's clean, this is ready to go in. Thanks!

gcolonese:
Done — rebased onto current `main` (95e23c7). The `library_link` `hasattr` guard hunk and `tests/test_ipc_footprint_library_link_guard.py` are both absent from the diff now; the PR contains only the bbox/courtyard changes across the four files. Black formatting was cleaned up in `component.py` and all pre-commit hooks (black, isort, flake8, mypy, eslint) pass cleanly.

mixelpixx (approved):
Approving and merging — thanks for the clean rebase.

Diff is now exclusively the bbox/courtyard additions (4 files) with the `library_link` `hasattr` guard hunk and its test file dropped, as expected after #157 merged. The additive output-shape change (new `boundingBox` and `courtyard` fields on `get_component_properties` and `get_component_list`, with SWIG-then-IPC fallback) is non-breaking — existing callers see only added keys. The SWIG-courtyard try/except for older pcbnew bindings is the right level of defence.

Black/isort/flake8/mypy/eslint all passing on the rebase is appreciated.

PR #164, opened 2026-05-07
fix(loader): read global sym-lib-table; quoted URIs with spaces

gcolonese:
`dynamic_symbol_loader` only consulted the project-local `sym-lib-table` and a hardcoded list of bundled symbol directories, so libraries registered via the user-global `sym-lib-table` (`Preferences > Manage Symbol Libraries > Global`) were invisible to `add_schematic_component`. Common case: company libraries that live under OneDrive / a network share / any other custom path the user added through the GUI.

Also widened the `sym-lib-table` parser regex to accept quoted URIs (and quoted names) that contain spaces — required for paths like `C:/Users/.../OneDrive - Company/Documents/KiCad/...`. The old bareword capture stopped at the first space.

New search order:
1. Project `sym-lib-table`
2. User-global `sym-lib-table`
   - Windows: `~/AppData/Roaming/kicad/<ver>/sym-lib-table`
   - Linux: `~/.config/kicad/<ver>/sym-lib-table`
   - macOS: `~/Library/Preferences/kicad/<ver>/sym-lib-table`
3. Bundled / well-known symbol directories

## Test plan
- `pytest tests/` passes (678 passing).
- Manual: `add_schematic_component` resolves a symbol from a globally-registered library whose `.kicad_sym` lives under a path with spaces (verified against `FAST_PCB_Library` at `C:/Users/.../OneDrive - .../Documents/KiCad/9.0/symbols/FAST_PCB_Library.kicad_sym`).

### DIFF
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -62,7 +62,12 @@ class DynamicSymbolLoader:
 
         Search order:
         1. Project-specific sym-lib-table (if project_path is set)
-        2. Global KiCad symbol library directories
+        2. Global KiCad sym-lib-table (~/AppData/Roaming/kicad/<ver>/sym-lib-table on
+           Windows, ~/.config/kicad/<ver>/sym-lib-table on Linux,
+           ~/Library/Preferences/kicad/<ver>/sym-lib-table on macOS) — covers user-
+           registered libraries that live outside the bundled symbol directories
+           (e.g. company libraries in OneDrive, network shares, custom paths).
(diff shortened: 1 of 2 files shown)


mixelpixx (approved):
Approving. The three-tier resolution (project-local → global user `sym-lib-table` → bundled) is the right precedence, the regex update correctly handles both quoted and bare `name`/`uri` forms (good fix for OneDrive paths with spaces), and the platform-specific config paths cover KiCad 8/9/10 on Windows/Linux/macOS. Four regression tests is solid coverage. Thanks!

PR #165, opened 2026-05-07
Add route_arc_trace for native KiCad arc track

karu2003:
Added a new MCP routing tool route_arc_trace to create true KiCad arc tracks from start/mid/end points, instead of polyline approximations.
Implemented end-to-end support across the stack:
MCP tool registration and schema in src/tools/routing.ts
SWIG command handler in python/commands/routing.py
Command routing and IPC dispatch in python/kicad_interface.py
IPC backend arc creation via ArcTrack in python/kicad_api/ipc_backend.py
Kept behavior consistent with existing routing tools (layer, width, optional net, success/error payloads), while enabling real curved geometry in PCB.

### DIFF
--- a/python/commands/routing.py
+++ b/python/commands/routing.py
@@ -342,6 +342,81 @@ class RoutingCommands:
                 "errorDetails": str(e),
             }
 
+    def route_arc_trace(self, params: Dict[str, Any]) -> Dict[str, Any]:
+        """Route a copper arc trace from start/mid/end points."""
+        try:
+            if not self.board:
+                return {
+                    "success": False,
+                    "message": "No board is loaded",
+                    "errorDetails": "Load or create a board first",
+                }
+
+            start = params.get("start")
+            mid = params.get("mid")
+            end = params.get("end")
+            layer = params.get("layer", "F.Cu")
+            width = params.get("width")
+            net = params.get("net")
+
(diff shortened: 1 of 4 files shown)


mixelpixx (approved):
Approving and merging.

Clean end-to-end addition of a real KiCad arc primitive — SWIG path uses `pcbnew.PCB_ARC` with `SetStart`/`SetMid`/`SetEnd`, IPC path uses `kipy.board_types.ArcTrack` with the standard `begin_commit`/`push_commit` pattern, and the `hasattr(self.ipc_board_api, "add_arc_track")` guard gives graceful degradation when the IPC backend version doesn't support arcs.

Two nits I'm noting but not blocking on — happy with a follow-up PR or leave them be:

1. The IPC `layer_map` covers F.Cu/B.Cu/In1.Cu/In2.Cu and silently falls back to F.Cu for unknown layers (e.g. In3.Cu/In4.Cu on a 6+ layer board). Worth either expanding to enumerate `BL_*_Cu` variants or returning an error for unknown layers rather than silently routing on the wrong layer.

2. No unit test for the new SWIG handler — given the IPC `hasattr` guard already provides the runtime degradation path, this is low-risk to merge as-is, but a test covering the start/mid/end coordinate scaling (mm → nm) would be welcome.

Thanks for the work!

PR #167, opened 2026-05-08
fix(jlcpcb): use platform user data dir for parts database

seanybaggins:
JLCPCBPartsManager defaulted db_path to a "data/" directory computed relative to __file__, which fails with read-only filesystems when the package is installed to a system-managed prefix (e.g. /nix/store, an immutable container image, or /usr/lib). The same pattern in download_jlcpcb.py would silently scatter the ~1.5 GB JLCPCB cache inside the install tree even when it is writable.

The original integration plan (docs/archive/JLCPCB_INTEGRATION_PLAN.md) called for a per-user database under ~/.kicad-mcp/. This change moves the default to the platform-appropriate user data directory by adding a new PlatformHelper.get_data_dir() helper that mirrors the existing get_config_dir() / get_cache_dir() conventions:

  - Linux:   XDG_DATA_HOME/kicad-mcp or ~/.local/share/kicad-mcp
  - macOS:   ~/Library/Application Support/kicad-mcp
  - Windows: %USERPROFILE%\.kicad-mcp\data

Both JLCPCBPartsManager and download_jlcpcb.py now resolve their database paths through this helper. ensure_directories() and detect_platform() include the new directory. Unit tests parallel to the existing config_dir/cache_dir cases cover platform-appropriate paths and the relative-XDG_DATA_HOME edge case.

### DIFF
--- a/download_jlcpcb.py
+++ b/download_jlcpcb.py
@@ -19,11 +19,14 @@ import sys
 import time
 from pathlib import Path
 
-DATA_DIR = Path(__file__).parent / "data"
-DATA_DIR.mkdir(exist_ok=True)
+sys.path.insert(0, str(Path(__file__).parent / "python"))
+from utils.platform_helper import PlatformHelper  # noqa: E402
+
+DATA_DIR = PlatformHelper.get_data_dir()
+DATA_DIR.mkdir(parents=True, exist_ok=True)
 
 CACHE_DIR = DATA_DIR / "jlcparts_cache"
-CACHE_DIR.mkdir(exist_ok=True)
+CACHE_DIR.mkdir(parents=True, exist_ok=True)
 
 BASE_URL = "https://yaqwsx.github.io/jlcparts/data"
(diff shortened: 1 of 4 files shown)


mixelpixx (approved):
Approving and merging.

Right shape — `get_data_dir()` parallels the existing `get_config_dir()` / `get_cache_dir()` / `get_log_dir()` conventions in `PlatformHelper`, follows OS standards (XDG_DATA_HOME, Application Support, %USERPROFILE%\.kicad-mcp\data), and the relative-XDG_DATA_HOME edge case is explicitly handled with a warning + fallback. Both `JLCPCBPartsManager` and `download_jlcpcb.py` now route through it, and `ensure_directories()` / `detect_platform()` are updated so the new directory shows up in diagnostics. Four tests parallel to the existing config/cache cases.

Solves a real problem for Nix / immutable-container / system-prefix installs where the install tree is read-only. Thanks!

PR #168, opened 2026-05-09
fix(board-view): use kicad-cli instead of pcbnew/cffi, return image to Claude

jflaflamme:
## Problem

`get_board_2d_view` has two bugs that make it unusable in practice:

1. **cffi crash** — uses `pcbnew.PLOT_CONTROLLER` which loads KiCAD's C++ bindings via cffi. On Linux/macOS where the Python env cffi version doesn't match KiCAD's compiled version, the call crashes with a cffi version mismatch error.

2. **Claude can't see the image** — the TypeScript handler wraps the result in `JSON.stringify()` and returns it as `type: "text"`. Claude receives a wall of base64 text instead of a rendered image.

## Fix

**Python (`python/commands/board/view.py`)** — rewrite `get_board_2d_view` to shell out to `kicad-cli pcb export svg`, exactly the same pattern already used by `get_schematic_view`. No pcbnew/cffi import needed. Supports `--layers` filter for targeted renders (e.g. copper-only for island inspection). Falls back to returning SVG if `cairosvg` is not installed.

**TypeScript (`src/tools/board.ts`)** — return `type: "image"` with `mimeType: "image/png"` so Claude can render it visually, matching what `get_schematic_view` already does. Added a proper tool description so the router knows when to call it. Added `pcbPath` as an explicit parameter (mirrors schematic tools pattern).

## Test

```
# verify kicad-cli is in PATH
kicad-cli pcb export svg --help

# call the tool
get_board_2d_view({
  pcbPath: "/path/to/project.kicad_pcb",
  layers: ["F.Cu", "B.Cu", "Edge.Cuts"],
  format: "png"
})
```

Claude will receive and display the board image directly.

### DIFF
--- a/python/commands/board/view.py
+++ b/python/commands/board/view.py
@@ -14,6 +14,64 @@ from PIL import Image
 logger = logging.getLogger("kicad_interface")
 
 
+def _svg_to_png(svg_path: str, width: int, height: int) -> Optional[bytes]:
+    """Convert SVG to PNG. No cffi dependency.
+
+    Priority:
+      1. pymupdf (fitz) — bundled MuPDF renderer, pure Python, no system deps
+      2. Inkscape CLI — accurate KiCAD SVG rendering
+      3. ImageMagick convert — broad availability fallback
+    Returns PNG bytes or None if all converters fail.
+    """
+    import subprocess
+    import tempfile
+
+    try:
+        import fitz
+
+        doc = fitz.open(svg_path)
+        page = doc[0]
+        mat = fitz.Matrix(width / page.rect.width, height / page.rect.height)
+        return page.get_pixmap(matrix=mat).tobytes("png")
+    except Exception:
+        pass
+
+    out_path = os.path.join(tempfile.mkdtemp(), "out.png")
+
+    try:
+        r = subprocess.run(
+            [
+                "inkscape",
+                svg_path,
+                "--export-type=png",
+                f"--export-width={width}",
+                f"--export-height={height}",
+                f"--export-filename={out_path}",
+            ],
+            capture_output=True,
+            timeout=60,
+        )
+        if r.returncode == 0 and os.path.exists(out_path):
+            with open(out_path, "rb") as f:
+                return f.read()
+    except (FileNotFoundError, subprocess.TimeoutExpired):
+        pass
+
+    try:
+        r = subprocess.run(
+            ["convert", "-density", "150", svg_path, "-resize", f"{width}x{height}", out_path],
+            capture_output=True,
+            timeout=60,
+        )
+        if r.returncode == 0 and os.path.exists(out_path):
+            with open(out_path, "rb") as f:
+                return f.read()
+    except (FileNotFoundError, subprocess.TimeoutExpired):
+        pass
+
+    return None
+
+
 class BoardViewCommands:
     """Handles board viewing operations"""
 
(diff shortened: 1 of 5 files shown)


mixelpixx (changes requested):
Thanks for tackling this — the cffi mismatch on `pcbnew.PLOT_CONTROLLER` has been a real pain point, and switching to `kicad-cli pcb export svg` is the right escape hatch. Returning the result as an MCP `image` content block instead of JSON-wrapped base64 is also the right call — that's what makes the rendered board actually visible to Claude.

I do need to ask for one change before merging. The TS schema makes `pcbPath` a required `z.string()`:

```ts
pcbPath: z.string().describe("Absolute path to the .kicad_pcb file"),
```

But the Python implementation still has a `self.board.GetFileName()` fallback when `pcbPath` is omitted:

```python
pcb_path = params.get("pcbPath")
if not pcb_path:
    if self.board:
        pcb_path = self.board.GetFileName()
```

Any existing caller that was relying on the "use the currently loaded board" behaviour will now be rejected by the TS layer before the Python fallback ever runs. Could you make `pcbPath` optional in the TS schema so both paths work?

```ts
pcbPath: z.string().optional().describe("Absolute path to the .kicad_pcb file (defaults to currently loaded board)"),
```

A few smaller things you can fix in the same push or leave for follow-up — I'm not blocking on these:

1. **Duplicate `_svg_to_png`** — the same function appears in both `python/commands/board/view.py` and `python/kicad_interface.py`. Worth extracting to a shared util (e.g. `python/utils/svg_render.py`) so the priority chain stays in one place.

2. **`pymupdf` as a soft dep** — the `import fitz` priority-1 path requires pymupdf, which isn't in `requirements.txt`. It's properly wrapped in try/except so it silently falls through, but worth mentioning in the tool description or a docs note that pymupdf produces the best output if installed.

3. **`shutil.which("convert")`** — ImageMagick 7 split `convert` into `magick convert`; on systems with only IM7 the convert fallback won't find anything. Low priority; the Inkscape path is the realistic primary fallback.

Once `pcbPath` is optional this is ready to go in. Thanks!

mixelpixx:
Thanks for the Copilot follow-up — those changes look reasonable. There's still one outstanding item from my earlier review though: the `pcbPath` parameter is still required in the TS schema:

```ts
pcbPath: z.string().describe("Absolute path to the .kicad_pcb file"),
```

Could you make it optional so the Python fallback to `self.board.GetFileName()` actually gets exercised when callers don't supply a path?

```ts
pcbPath: z.string().optional().describe("Absolute path to the .kicad_pcb file (defaults to currently loaded board)"),
```

Once that's in, this is ready to go. Thanks!

mixelpixx:
Pushed the `pcbPath` fix directly to the branch (thanks for enabling maintainer edits) — `get_board_2d_view` now declares `pcbPath` as `.optional()`, so the Python `self.board.GetFileName()` fallback for the currently-loaded board is reachable again. Left `import_svg_logo`'s `pcbPath` required as before, since that tool needs an explicit target.

Verified: `tsc` builds clean, and a unit check confirms that calling `get_board_2d_view` with no `pcbPath` now falls back to the loaded board (previously rejected at the schema layer).

The feature itself looks good and is security-clean (subprocess list-args, no shell, timeouts). Merging shortly. Two tiny non-blocking follow-ups for whenever: `_svg_to_png` is duplicated in `view.py` and `kicad_interface.py` (could share a util), and it leaks a `mkdtemp` dir on the Inkscape/ImageMagick path.

mixelpixx:
Heads up: this now conflicts with main and needs a rebase.

Since you opened this, main picked up a separate change to the same function. get_board_2d_view on main gained a responseMode option (inline vs file) so large boards can be written to disk instead of returned as base64 inline. That change was made on the existing pcbnew/PLOT_CONTROLLER path, which is exactly the path this PR replaces with kicad-cli. So the two versions of the function collide in both view.py and src/tools/board.ts.

What I'd like from a rebase onto current main: keep your kicad-cli approach (that's the part that fixes the cffi crash), but fold in the responseMode inline/file behavior that's now on main so we don't lose it. A couple of smaller things while you're in there: main still supports jpg output and this PR drops it to png/svg only, so either keep jpg or call out that it's intentionally going away; and the pcbPath optional fix I pushed earlier should carry through.

If the rebase gets messy, say the word and I can take a pass at reconciling the two myself. No rush either way.

PR #169, opened 2026-05-09
fix(add_schematic_component): support hierarchical sub-sheets

inktomi:
## Summary

`DynamicSymbolLoader.create_component_instance` only handled root schematics: it located its insertion point by searching for `(sheet_instances`, which exists only in the root `.kicad_sch`. Adding a component to any hierarchical sub-sheet raised `Could not find insertion point in schematic` and aborted the call — no symbol was inserted, every `add_schematic_component` request into a child sheet was a hard fail.

## Repro

In any hierarchical project (root sheet + sub-sheet via `(sheet ...)` reference), call `add_schematic_component` against the sub-sheet:

```
add_schematic_component(
  schematicPath = "<project>/sub-sheet.kicad_sch",
  symbol = "Device:R", reference = "R_TEST", value = "100k",
  position = {x: 50, y: 50},
  footprint = "Resistor_SMD:R_0603_1608Metric"
)
```

Expected: success. Actual (before this PR): `Failed to add component: Could not find insertion point in schematic`.

## Fix

`WireManager.add_label` already handles the hierarchical case by falling back to inserting before the closing paren of the outer `(kicad_sch ...)` form when `(sheet_instances` is absent. Apply the same fallback in `create_component_instance`:

```python
insert_at = content.rfind("(sheet_instances")
if insert_at == -1:
    stripped = content.rstrip()
    if not stripped.endswith(")"):
        raise ValueError("Could not find insertion point in schematic")
    insert_at = len(stripped) - 1   # final ')' of the outer (kicad_sch ...) form
```

Root-sheet behaviour is unchanged — the existing `(sheet_instances` path runs first.

## Test plan

- [x] `pytest tests/test_add_schematic_component.py` — 14/14 pass (3 new in `TestCreateComponentInstanceSubSheet` + 11 existing). The pre-existing `TestAddComponentMirrorParam::test_mirror_x_arg_is_silently_dropped` failure is unrelated and present on `main`.
- [x] Manual: aircam-pdb hierarchical schematic — adding `Device:R` to a sub-sheet now succeeds, paren count balances, sexpdata loads the result, and the new symbol round-trips through `kicad-cli sch erc` cleanly.
- [ ] Reviewer: try `add_schematic_component` against any of your own hierarchical projects to confirm.

### DIFF
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -445,9 +445,16 @@ class DynamicSymbolLoader:
         insert_marker = "(sheet_instances"
         insert_at = content.rfind(insert_marker)
         if insert_at == -1:
-            raise ValueError("Could not find insertion point in schematic")
-
-        content = content[:insert_at] + instance_block + "\n  " + content[insert_at:]
+            # Hierarchical sub-sheets don't carry (sheet_instances ...) — only the
+            # root .kicad_sch does. Fall back to inserting just before the final
+            # closing paren of the outer (kicad_sch ...) form.
+            stripped = content.rstrip()
+            if not stripped.endswith(")"):
+                raise ValueError("Could not find insertion point in schematic")
+            insert_at = len(stripped) - 1
+            content = content[:insert_at] + instance_block + "\n" + content[insert_at:]
+        else:
+            content = content[:insert_at] + instance_block + "\n  " + content[insert_at:]
 
         with open(schematic_path, "w", encoding="utf-8") as f:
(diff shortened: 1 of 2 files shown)


mixelpixx (approved):
Approving and merging.

Clean fallback that mirrors the existing pattern in `WireManager.add_label`: when `(sheet_instances ...)` isn't found, insert before the final closing paren of the outer `(kicad_sch ...)` form. Root-sheet behaviour is unchanged. Three regression tests with paren balance + sexpdata round-trip checks. Thanks!

PR #170, opened 2026-05-09
fix(add_schematic_wire): support hierarchical sub-sheets

inktomi:
## Summary

`WireManager.add_wire` and `add_polyline_wire` bailed out with `No sheet_instances section found in schematic` whenever they were called against a hierarchical sub-sheet — that block is only ever emitted on the root `.kicad_sch`. The handler in `kicad_interface.py` flattened the `False` return into a generic `Failed to add wire` message, so callers got no diagnostic about why their wire didn't land.

## Repro

In any hierarchical project, call `add_schematic_wire` against a sub-sheet:

```
add_schematic_wire(
  schematicPath = \"<project>/sub-sheet.kicad_sch\",
  waypoints = [[40, 40], [40, 50]],
  snapToPins = false
)
```

Expected: a wire is inserted. Actual (before this PR): `Failed to add wire`, file unchanged (verified by md5).

## Fix

`WireManager.add_label` already uses the right fallback for hierarchical sub-sheets — when `(sheet_instances ...)` is absent, append the new item at the end of the outer `(kicad_sch ...)` form. Apply the same pattern to both `add_wire` and `add_polyline_wire`:

```python
if sheet_instances_index is None:
    sheet_instances_index = len(sch_data)
sch_data.insert(sheet_instances_index, wire_sexp)
```

Root-sheet behaviour is unchanged.

## Test plan

- [x] `pytest tests/test_add_wire_sub_sheet.py` — 6/6 pass (new file). Covers `add_wire` and `add_polyline_wire` on a sub-sheet, paren balance, sexpdata round-trip, and segment count for an N-point polyline.
- [x] `pytest tests/test_wire_connectivity.py tests/test_wire_junction_changes.py tests/test_add_label_empty_schematic.py` — 133/133 pass (no regressions on existing wire tests).
- [x] Manual: aircam-pdb hierarchical sub-sheet — `add_wire` now produces a properly-balanced file and the wire round-trips through `kicad-cli sch erc`.

### DIFF
--- a/python/commands/wire_manager.py
+++ b/python/commands/wire_manager.py
@@ -174,7 +174,9 @@ class WireManager:
                 start_point, end_point, stroke_width, stroke_type
             )
 
-            # Find insertion point (before sheet_instances)
+            # Find insertion point (before sheet_instances on the root sheet,
+            # or appended to the end on a hierarchical sub-sheet which has no
+            # sheet_instances block).
             sheet_instances_index = None
             for i, item in enumerate(sch_data):
                 if isinstance(item, list) and len(item) > 0 and item[0] == _SYM_SHEET_INSTANCES:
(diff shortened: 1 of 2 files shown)


mixelpixx (approved):
Approving and merging.

Same pattern as #169 applied to `add_wire` and `add_polyline_wire`: fall through to "insert at end" when `(sheet_instances ...)` is absent rather than logging an error and returning False. Six regression tests cover both methods, the segment-count invariant for polylines, paren balance, and sexpdata round-trip. Thanks!

PR #171, opened 2026-05-09
fix(sync_schematic_to_board): add missing footprints, not just nets

inktomi:
## Summary

`_handle_sync_schematic_to_board` walked `board.GetFootprints()` and assigned nets to existing pads, but had no path to *add* footprints for schematic symbols whose Reference was not yet on the board. New parts placed on the schematic landed in the net list with no PCB representation — the rats nest had nowhere to terminate, and `place_component` / `route_pad_to_pad` would fail because the target footprint did not exist.

KiCad's \"Update PCB from Schematic\" (F8) implicitly adds the missing footprints; this PR brings the MCP into line.

## Repro

1. Add a new symbol to a schematic (any path — direct file edit if needed) with a `Footprint` property that exists in your `fp-lib-table`.
2. Run `sync_schematic_to_board`.
3. Diff the `.kicad_pcb` before/after.

Expected: a new `(footprint …)` block in the PCB for the new symbol, plus net assignments. Actual (before this PR): only net assignments updated; no footprint added.

## Fix

Two new helpers in `kicad_interface.py`:

* `_extract_components_from_schematic` — runs `kicad-cli sch export netlist --format kicadxml` (the same pattern `_handle_generate_netlist` already uses) and returns a flat `[{reference, value, footprint}]` list. Walks hierarchical sub-sheets transparently because kicad-cli does.
* `_add_missing_footprints_from_schematic` — diffs the netlist's references against `board.GetFootprints()`, resolves each missing component against the project's `fp-lib-table` via `LibraryManager`, and adds it via `pcbnew.FootprintLoad`. Reference / value / FPID are set on the loaded footprint; position defaults to the board origin so the user (or autoplacer) can move it. `#PWR…` / `#FLG…` are excluded — they have no PCB footprint.

The pad-net assignment loop now runs *after* the add path, so newly placed footprints get their nets assigned in the same call.

The response payload gains `footprints_added` and `footprints_skipped` diagnostic lists, and the `message` field reports the new footprint count alongside the existing net / pad counts.

## Test plan

- [x] `pytest tests/test_sync_schematic_to_board_footprints.py` — 9/9 pass (new file). Covers: add path for a missing reference, skipping refs already on the board, `#PWR` / `#FLG` exclusion, missing footprint property, unknown library nickname, kicad-cli failure modes, and XML netlist parsing.
- [x] `pytest tests/test_hierarchical_pad_net_map.py` — 15/15 pass (no regressions on the underlying schematic walk).
- [x] Manual: ran kicad-cli netlist export against the aircam-pdb hierarchical schematic — 55 components, all with footprints, library breakdown matches the project's `fp-lib-table`.
- [ ] Reviewer: in a project where you add a new symbol to the schematic and run `sync_schematic_to_board`, verify the PCB now contains the corresponding footprint at the origin.

## Out of scope

* Removal of footprints that are no longer in the schematic. F8 in KiCad asks the user; this PR keeps the MCP behaviour conservative — never delete footprints. Would need a separate PR with explicit user opt-in.
* Smart positioning of new footprints (e.g. an off-board grid). Today they all land at `(0, 0)`; that's deterministic and easy to find. A future PR could compute the board's outline and stage new footprints just outside it.

### DIFF
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -3979,6 +3979,15 @@ class KiCADInterface:
             # Build hierarchical pad→net map (walks all sub-sheets)
             pad_net_map, net_names = self._build_hierarchical_pad_net_map(schematic_path)
 
+            # Add missing footprints from the schematic to the board *before*
+            # we add nets and assign pads — F8 in KiCad does this implicitly
+            # ("Update PCB from Schematic"), but our previous implementation
+            # only mutated nets, leaving newly-added schematic symbols with no
+            # PCB footprint at all.
+            added_footprints, skipped_footprints = self._add_missing_footprints_from_schematic(
+                board, schematic_path
+            )
+
             # Add all nets to board
             netinfo = board.GetNetInfo()
             nets_by_name = netinfo.NetsByName()
@@ -3993,7 +4002,7 @@ class KiCADInterface:
             netinfo = board.GetNetInfo()
             nets_by_name = netinfo.NetsByName()
 
-            # Assign nets to pads
+            # Assign nets to pads (now also covers any footprints we just added)
             assigned_pads = 0
             unmatched = []
             for fp in board.GetFootprints():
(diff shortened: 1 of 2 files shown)


mixelpixx (approved):
Approving and merging.

This brings `sync_schematic_to_board` into line with F8 ("Update PCB from Schematic") — adding missing footprints, not just nets, was the obvious gap and the implementation is exactly right: extract via `kicad-cli sch export netlist --format kicadxml` (same pattern `_handle_generate_netlist` uses), diff against `board.GetFootprints()`, resolve via `LibraryManager`, add via `pcbnew.FootprintLoad`. The subprocess invocation is safe — argv list (no shell), 60-second timeout, tempfile cleaned up in a finally block, and the schematic path is a positional argv element with no injection vector. Skipping `#PWR*` / `#FLG*` is correct.

The `footprints_added` / `footprints_skipped` diagnostic lists are great for surfacing fp-lib-table misses to the AI client instead of silently dropping components.

Nine regression tests cover the happy path, already-on-board, power exclusion, missing-Footprint-property, unknown library, the XML parser, and the kicad-cli failure modes. Thanks!

PR #172, opened 2026-05-09
fix(auto-save-guard): refuse only on content divergence, not mtime

inktomi:
## Stacked on #151

This PR is **stacked on top of #151** (\`feat/auto-save-guard\`). The diff against \`main\` includes #151's commit because that's the only way to express the dependency in a single PR; review just the second commit (\`fix(auto-save-guard): refuse only on content divergence, not mtime\`) here. If #151 lands first, this PR's diff against \`main\` collapses to only the second commit.

## Summary

The guard added in #151 records \`(mtime_ns, sha256)\` as the file's disk signature and refuses auto-save when the recorded tuple no longer matches the current one. Comparing the **full tuple** means any mtime delta fires the refusal — including a bare \`touch\` of the file, an atime-style backup tool, or any MCP read path that opened the \`.kicad_pcb\` between load and save. Users get trapped: every write needs an explicit \`save_project\` call to bypass the false positive (already documented as a workaround in fork users' notes).

## Repro

1. \`open_project\` to load a board.
2. \`touch <project>/board.kicad_pcb\` in another shell (mtime advances, content unchanged).
3. Try any board-mutating tool (e.g. \`add_via\`).

Expected (against #151 as-is): \`autoSave.saved: false\` with \`diskChangedExternally: true\`. The mutation is dropped.

## Fix

Compare on \`sha256\` only; mtime is incidental. The actual data-loss scenario the guard is meant to catch — an external write that genuinely changed the file — produces a different content hash, which is what the guard now keys off. After a touch-only mtime advance with content unchanged, refresh the recorded signature so we don't re-hash on every subsequent call.

\`\`\`python
# Before:
if expected != current:               # tuple equality — fires on mtime alone
    refuse(...)

# After:
if expected[1] != current[1]:         # sha256 only — actual content
    refuse(...)
elif expected != current:             # touch-only: refresh recorded mtime
    self._board_disk_signature = current
\`\`\`

Drops the mtime-equality fast path on \`_disk_signature\`: a filesystem with coarse mtime resolution (FAT32, some network mounts) could accept two writes inside one mtime tick; trusting mtime as a hash cache key would re-introduce a class of silent overwrites the guard exists to prevent. The hash itself is cheap (sha256 over a typical \`.kicad_pcb\` completes in tens of milliseconds).

## Test plan

- [x] \`pytest tests/test_auto_save_guard.py\` — 14/14 pass (10 from #151 + 4 new). The new ones cover:
  - touch-only mtime advance proceeds and refreshes the signature
  - content change at the same mtime is **still** refused (hash divergence drives the decision, not tuple equality)
  - the user-facing warning calls out \"contents\", not \"mtime\"
  - \`_disk_signature\` returns the same hash when content is unchanged even after the file's mtime advances
- [x] No regressions in the existing #151 tests — the guard still refuses on real external content changes, still snapshots backups before overwrite, still handles first-save / no-board / no-path cases.

## Notes

This PR also unblocks the \`open_project → save_project\` workaround documented in downstream users' notes (the \"call \`save_project\` to bypass the mtime check\" pattern). With this fix in place, the workaround is no longer necessary — auto-save just works after a \`touch\` or other no-content-change file event.

### DIFF
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -7,11 +7,14 @@ and KiCAD's Python API (pcbnew). It receives commands via stdin as
 JSON and returns responses via stdout also as JSON.
 """
 
+import hashlib
 import json
 import logging
 import os
+import shutil
 import sys
 import traceback
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
@@ -259,6 +262,12 @@ class KiCADInterface:
         """Initialize the interface and command handlers"""
         self.board = None
         self.project_filename = None
+        # On-disk signature (mtime_ns, sha256_hex) of self.board's file as of
+        # last load or successful auto-save.  Used by _auto_save_board() to
+        # detect external modifications and refuse to clobber them.
+        self._board_disk_signature: Optional[Tuple[int, str]] = None
+        # Number of timestamped backups to keep in .mcp-backups/ per board file.
+        self._auto_save_backup_keep = 20
         self.use_ipc = USE_IPC_BACKEND
         self.ipc_backend = ipc_backend
         self.ipc_board_api = None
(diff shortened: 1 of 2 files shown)


mixelpixx (approved):
Approving and merging.

The actual fix is a two-character change (`expected != current` → `expected[1] != current[1]`) plus a refresh-the-recorded-signature path for the touch-only case, but the reasoning is exactly right: mtime alone is not a conflict signal, and the underlying invariant the guard exists to enforce is content equality (sha256). Dropping the mtime-equality fast path on `_disk_signature` is the correct call — a coarse-resolution filesystem (FAT32, network mounts) could otherwise mask a real content change within one mtime tick.

`test_auto_save_refuses_when_content_differs_even_at_same_mtime` is the critical regression test — it pins that the guard still fires for the actual data-loss scenario, not just the false-positive case we're fixing.

Since #151 is already in `main`, the squash merge will collapse to just the fix commit. Thanks!

PR #173, opened 2026-05-10
fix(swig): detect and recover from BOARD proxy dehydration

inktomi:
## Summary

Fixes a long-standing failure mode on KiCAD main builds where, after a sequence like `delete_trace` + auto-save, every subsequent board operation fails with `AttributeError` on `GetDesignSettings` / `GetBoardEdgesBoundingBox` / `GetCurrentViaSize`. The MCP keeps reporting `"Opened project: foo.kicad_pcb"` while the board is actually unusable, with no path to recovery short of restarting the MCP server.

Root cause is in pcbnew's SWIG bindings: `pcbnew.LoadBoard` and `pcbnew.SaveBoard` can leave `self.board` as a raw `SwigPyObject` with no method dispatch table after certain in-process call sequences. This PR adds a defensive recovery layer at the wrapper level — it does not (cannot) fix the underlying pcbnew issue.

## Symptom users hit (real incident)

Sequence on a 21MB PCB, KiCad 10 `(version 20260306)`:

1. `open_project` ✓
2. `add_via` (~20×) ✓
3. `refill_zones` ✓ (auto-save refused once due to external edit, re-opened)
4. `run_drc` (timed out at MCP transport layer)
5. `open_project` claims success
6. `delete_trace` ✓
7. `add_via` → `'SwigPyObject' object has no attribute 'GetCurrentViaSize'`
8. `open_project` claims success
9. `add_via` → `'SwigPyObject' object has no attribute 'GetDesignSettings'`
10. State is permanently stuck until MCP restart

## What this PR adds

**Dehydration helpers in `KiCADInterface`:**

- `_is_board_healthy(board=None)` — probes for stable BOARD methods (`GetDesignSettings`, `GetBoardEdgesBoundingBox`, `GetFileName`). Missing on a dehydrated SwigPyObject — `hasattr()` catches the state without segfaulting.
- `_safe_load_board(path)` — wraps `pcbnew.LoadBoard`, checks health, and on dehydration reloads the `pcbnew` module via `importlib.reload` and retries once. Returns `None` when recovery is impossible so callers surface real failure rather than fake success.

**Wired into:**

- `handle_command` for `open_project` / `create_project` — validates the loaded board and either recovers (with a `warnings[]` entry) or returns `success=False` with explicit "restart the MCP server" `errorDetails`. Never claims success when the board is unusable.
- `_auto_save_board` — detects dehydration introduced by `SaveBoard` itself and reloads from disk so the next command sees a usable proxy. This is the post-`delete_trace` failure mode users hit.
- `_handle_place_component`, `_handle_sync_schematic_to_board`, `_handle_import_svg_logo`, `_handle_refill_zones` — all go through `_safe_load_board` instead of bare `LoadBoard`, surfacing real errors consistently.

**Adjacent fixes from the same incident:**

- `_handle_check_kicad_ui` — used to call `is_running()` and `get_process_info()` separately with different detection methods. They could disagree, producing the confusing `running=true, processes=[]` state users hit after manually quitting KiCAD. `processes` is now the single source of truth and `running` is derived from `len()`.
- `run_drc` — accepts a `timeoutSec` param (default 600s, clamped to `[10, 1800]`) so callers with smaller MCP transport budgets can bound the kicad-cli subprocess. Same timeout is applied to the optional report-generation subprocess.

## Relationship to other PRs

Complementary to **#151 (auto-save-guard)**. PR #151 expands the `LoadBoard` call rate by refusing saves on external file change and forcing re-`open_project` cycles. The SWIG dehydration is triggered by repeated `LoadBoard` in a single Python process, so #151 makes users hit this bug more often — but the bug is in pcbnew on `main` regardless. This PR is the recovery layer; #151 is unrelated value.

I verified by code-auditing every open PR for new pcbnew/SWIG calls in the user's repro hot path: only #151/#172 touch code in that path, and they add only Python-level file I/O around the existing `pcbnew.SaveBoard` — no new SWIG calls.

## Test plan

- [x] `tests/test_swig_dehydration.py` — 17 new unit tests covering detection, recovery, the `open_project` surfacing path, the auto-save post-recovery path, the `check_kicad_ui` consistency, and `run_drc` timeout clamping.
- [x] Full test suite: same 12 pre-existing failures both before and after this change (none touched by this PR), +17 new tests passing.
- [ ] Manual verification against the reproducing PCB on KiCad nightly (`/Users/inktomi/Projects/aircam-pcb/Hardware-V3/aircam-pdb-v3.kicad_pcb`).

## Notes

- The fix is **defensive against the symptom**, not a fix to the underlying pcbnew SWIG issue. Restart-after-failure was the only previous recovery; now the recovery happens automatically (most cases) or surfaces a real error message (worst case).
- `delete_trace`'s SWIG ownership pattern (`Remove(track); track = None`) is unchanged — altering SWIG ownership semantics without a real KiCAD test environment is too risky. The recovery layer catches the symptom regardless of which call triggered it.

### DIFF
--- a/python/commands/design_rules.py
+++ b/python/commands/design_rules.py
@@ -187,6 +187,13 @@ class DesignRuleCommands:
                 }
 
             report_path = params.get("reportPath")
+            # Caller-overridable timeout (seconds). Defaults to 600s for big boards
+            # but smaller MCP transport budgets (e.g. 120s) can lower it explicitly.
+            try:
+                timeout_sec = int(params.get("timeoutSec", 600))
+            except (TypeError, ValueError):
+                timeout_sec = 600
+            timeout_sec = max(10, min(timeout_sec, 1800))  # clamp to [10, 1800]
 
             # Get the board file path
             board_file = self.board.GetFileName()
@@ -225,14 +232,14 @@ class DesignRuleCommands:
                     board_file,
                 ]
 
-                logger.info(f"Running DRC command: {' '.join(cmd)}")
+                logger.info(f"Running DRC command (timeout={timeout_sec}s): {' '.join(cmd)}")
 
-                # Run DRC
+                # Run DRC. subprocess.run kills the child on TimeoutExpired.
                 result = subprocess.run(
                     cmd,
                     capture_output=True,
                     text=True,
-                    timeout=600,  # 10 minute timeout for large boards (21MB PCB needs time)
+                    timeout=timeout_sec,
                 )
 
                 if result.returncode != 0:
@@ -318,7 +325,7 @@ class DesignRuleCommands:
                         "mm",
                         board_file,
                     ]
-                    subprocess.run(cmd_report, capture_output=True, timeout=600)
+                    subprocess.run(cmd_report, capture_output=True, timeout=timeout_sec)
 
                 # Return summary only (not full violations list)
                 return {
@@ -339,11 +346,14 @@ class DesignRuleCommands:
                     os.unlink(json_output)
 
         except subprocess.TimeoutExpired:
-            logger.error("DRC command timed out")
+            logger.error(f"DRC command timed out after {timeout_sec}s")
             return {
                 "success": False,
                 "message": "DRC command timed out",
-                "errorDetails": "Command took longer than 600 seconds (10 minutes)",
+                "errorDetails": (
+                    f"Command took longer than {timeout_sec} seconds; "
+                    "raise timeoutSec param for very large boards"
+                ),
             }
         except Exception as e:
             logger.error(f"Error running DRC: {str(e)}")
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -567,6 +567,47 @@ class KiCADInterface:
                         logger.info("Updating board reference...")
                         # Get board from the project commands handler
                         self.board = self.project_commands.board
+
+                        # Detect SWIG dehydration before claiming success.
+                        # Without this, every later board op sees a raw
+                        # SwigPyObject and raises AttributeError, while the
+                        # MCP keeps reporting "Opened project" — the exact
+                        # symptom users hit on KiCAD nightlies.
+                        if not self._is_board_healthy():
+                            board_path = (result.get("project") or {}).get("boardPath")
+                            recovered = None
+                            if board_path:
+                                logger.warning(
+                                    "Board after %s is SWIG-dehydrated; attempting recovery",
+                                    command,
+                                )
+                                recovered = self._safe_load_board(board_path)
+                            if recovered is not None:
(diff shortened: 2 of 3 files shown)


inktomi:
Note - the real error message that you get is a notification that python has crashed because of a c++ error. Not super freindly, but at least it's not silently crashing in the background anymore and getting stuck,

mixelpixx (approved):
Approving and merging.

This is a large PR but every line is defensive and well-justified. The dehydration symptom (`'SwigPyObject' object has no attribute 'GetDesignSettings'` while `open_project` still reports success) is a genuine and severe failure mode — users were stuck restarting the MCP server with no path to recovery. The `_is_board_healthy()` / `_safe_load_board()` pair is the right level of defence: probe stable methods via `hasattr()` (safe against the dehydrated proxy), and if pcbnew really has lost its dispatch table, `importlib.reload(pcbnew)` is a last-resort hammer that's properly gated and tested.

Two things I want to call out explicitly because they're worth being honest about:

1. **`importlib.reload(pcbnew)` and the `global pcbnew` rebind** is a notable maneuver. It can have ripple effects on other modules that hold references to the old pcbnew classes. The PR handles this responsibly — it's only invoked when the board is already unhealthy, the retry is gated, and recovery failure surfaces a clear "restart the MCP server" error rather than pretending to succeed. The right level of caution for a recovery-of-last-resort.

2. **Bundled scope.** The `run_drc` `timeoutSec` param and the `_handle_check_kicad_ui` consistency fix are bundled with the dehydration recovery — both came from the same incident triage. Each is small and backward-compatible (`timeoutSec` defaults to the existing 600s; `check_kicad_ui` now derives `running` from `len(processes)` so the two fields can no longer disagree). Disclosed honestly in the PR body. Acceptable.

17 regression tests covering detection, recovery, the `open_project` surfacing path, the auto-save post-recovery path, and the adjacent fixes. The audit of every open PR for new pcbnew/SWIG calls in the hot path was the right discipline.

Thanks for the thorough work.

mixelpixx:
Heads up — this is approved and ready to merge once a small rebase clears.

I merged #172 just before this one, which touches `_auto_save_board` (changed the conflict check from full-tuple comparison to sha256-only). That collides with this PR's restructure of the same function (the post-save dehydration check). GitHub now reports a merge conflict on this branch.

Could you rebase onto current `main` and reconcile? The semantic merge should be straightforward since you wrote both:

- Keep #172's `expected[1] != current[1]` comparison and the new "contents" warning string.
- Layer this PR's post-save `_is_board_healthy(...)` check and `_safe_load_board(...)` fallback on top.
- The `expected != current` mtime-only-refresh path from #172 should land just before this PR's dehydration check.

Once you've pushed the rebase I'll squash-merge. Thanks!

PR #140, opened 2026-04-30
fix: reconnect IPC backend after KiCad starts

scorp508:
## Summary
- Retry IPC at runtime for IPC-capable board tools in `auto` backend mode, allowing MCP to fall up from SWIG after KiCad starts.
- Add live backend status to `check_kicad_ui`, `launch_kicad_ui`, and `get_backend_info` responses.
- Route `query_traces` through IPC when available.
- Detect Windows per-user KiCad installs under `%LOCALAPPDATA%\Programs\KiCad` in addition to machine-wide installs.
- Handle KiCad 10 IPC `Box2` board-size responses that expose `pos` / `size` instead of `min` / `max`.

## Documentation
- Updated README, changelog, IPC backend status docs, UI auto-launch docs, visual feedback docs, and Windows troubleshooting docs.

## Tests
- Added regression coverage in `tests/test_backend_metadata.py` for backend metadata, runtime reconnect, fallback behavior, IPC `query_traces`, and KiCad 10 `Box2` board-size compatibility.

## Local validation
- [x] Code follows style guidelines: `pre-commit run --files CHANGELOG.md README.md docs\IPC_BACKEND_STATUS.md docs\UI_AUTO_LAUNCH.md docs\VISUAL_FEEDBACK.md docs\WINDOWS_TROUBLESHOOTING.md python\kicad_api\ipc_backend.py python\kicad_interface.py src\server.ts tests\test_backend_metadata.py`
- [x] All tests pass locally: `npm run build && npm test` — 654 passed, 2 skipped
- [x] New tests added for new behavior
- [x] Documentation updated
- [x] Commit message follows convention
- [x] No merge conflicts at PR creation time

## Manual validation
- Started MCP with KiCad closed and confirmed SWIG fallback.
- Restarted MCP with KiCad still closed and confirmed SWIG fallback.
- Started KiCad afterward and confirmed `check_kicad_ui` switched to IPC.
- Confirmed `get_board_info` and `query_traces` returned `_backend: "ipc"` / `_realtime: true`.

PR #174, opened 2026-05-12
Add query_zones tool for auditing copper pours

kevargaso:
## Summary

- `query_traces` silently omits `PCB_ZONE_T` objects, so layer-usage audits miss power planes and GND pours entirely on any board that uses copper zones.
- This PR adds a new `query_zones` tool that complements `query_traces` by iterating `board.Zones()` and returning each zone's net, layers, priority, fill state, min thickness, bounding box, and filled area.
- The filter surface (`net`, `layer`, `boundingBox`) is symmetric with `query_traces`, so callers can mix the two for a complete picture of a layer's usage.

### Why

Hit this while auditing a 6-layer board with In1/In4 as GND planes and In2 as a power layer. `query_traces` reported In1 and In4 as empty because all the copper there lives in zones, leading to incorrect recommendations. With `query_zones` the audit completes correctly without having to parse `.kicad_pcb` S-expressions manually.

### Implementation

- `python/commands/routing.py` — `RoutingCommands.query_zones()` using `pcbnew` API (`Zones()`, `GetLayerSet().CuStack()`, `GetBoundingBox()`, `IsFilled()`, `GetFilledArea()`). Defensive fallbacks for older pcbnew bindings.
- `python/kicad_interface.py` — command dispatch entry.
- `python/schemas/tool_schemas.py` — MCP tool schema.
- `src/tools/routing.ts` — MCP tool registration mirroring `query_traces`.
- `CHANGELOG.md` — entry under Unreleased.

### Output shape

```json
{
  "success": true,
  "zoneCount": 1,
  "zones": [
    {
      "uuid": "bbad52c1-...",
      "net": "GND",
      "netCode": 1,
      "layers": ["In1.Cu"],
      "priority": 0,
      "isFilled": true,
      "minThickness": 0.25,
      "boundingBox": { "x1": 150.56, "y1": 24.35, "x2": 230.56, "y2": 118.35, "unit": "mm" },
      "filledArea": 5720.23
    }
  ]
}
```

## Test plan

- [x] Manual test against a real 6-layer KiCad 9 board (44 zones detected across F.Cu/In1.Cu/In2.Cu/In3.Cu/In4.Cu/B.Cu); results match S-expression parsing of the same `.kicad_pcb`.
- [x] `layer=In1.Cu` filter returns only the matching GND pour.
- [x] `npm run build` passes with no TS errors.
- [ ] Add a regression test once maintainers point me at the preferred fixture board.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### New MCP Tools
 
+- `query_zones` — Query copper zones (filled pours) on the board with optional
+  filters by net, layer, or bounding box. Returns one entry per zone with its
+  net, layers, priority, fill state, min thickness, bounding box, and filled
+  area. Complements `query_traces`, which only reports tracks/vias and silently
+  omits power-plane and GND pours — making layer-usage audits incomplete on any
+  board that uses copper zones.
+
 - `set_schematic_component_property` — Add or update a single custom property
   (BOM / sourcing field) on a placed schematic symbol. Convenience wrapper
   around `edit_schematic_component` for the common case of attaching one MPN /
--- a/python/commands/routing.py
+++ b/python/commands/routing.py
@@ -707,6 +707,113 @@ class RoutingCommands:
                 "errorDetails": str(e),
             }
 
+    def query_zones(self, params: Dict[str, Any]) -> Dict[str, Any]:
+        """Query copper zones (filled pours) by net, layer, or bounding box.
+
+        Returns one entry per zone with its net, layers, priority, fill state,
+        and bounding box. Useful for auditing power planes / GND pours that
+        ``query_traces`` does not report (zones are PCB_ZONE_T, not tracks).
+        """
+        try:
+            if not self.board:
+                return {
+                    "success": False,
+                    "message": "No board is loaded",
+                    "errorDetails": "Load or create a board first",
+                }
+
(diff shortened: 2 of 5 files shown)


Mehanik:
Thank you!

mixelpixx (approved):
Approving and merging.

Clean addition mirroring `query_traces`'s filter surface (`net` / `layer` / `boundingBox`) so the two compose naturally — exactly the right shape for a layer-usage audit on boards that mix zones and traces. Implementation uses the standard pcbnew Zones API with defensive fallbacks where it matters: `hasattr(layer_set, "CuStack")` before falling back to `Seq()`, `hasattr(zone, "GetAssignedPriority")` for older bindings, and per-zone try/except so a single malformed zone doesn't kill the whole query. Output shape (`uuid`, `net`, `netCode`, `layers`, `priority`, `isFilled`, `minThickness`, `boundingBox`, `filledArea`) reads cleanly and gives the AI exactly what it needs.

Two follow-up notes (not blockers):

1. The new tool is SWIG-only — `query_zones` isn't in `IPC_CAPABLE_COMMANDS`. That's consistent with how `query_traces` started before #140 added IPC support, so it's a reasonable trajectory. Worth filing as a follow-up.

2. The unit-test slot is open — happy to point at `tests/test_query_traces.py` (or equivalent) as a fixture template if you want to add coverage. The manual 6-layer board validation in your PR body is solid for now.

Thanks for the work!

PR #177, opened 2026-05-14
fix(wire_connectivity): pwr-flags must not bridge unrelated power rails

### LINKED ISSUE #176: get_net_connections over-merges every power rail that has a PWR_FLAG into one mega-net
# `get_net_connections` over-merges every power rail that has a PWR_FLAG into one mega-net

## Summary

After commit 7f3a379 ("Treat PWR_FLAG anchors as connected in orphan-wire detection"), every power rail in a schematic that has a `PWR_FLAG` ERC marker on it is virtually bridged into a single mega-net by `get_net_connections`. Querying any one of those rails returns the union of pins on every rail.

## Reproduction

Any schematic with two or more power rails each carrying a `PWR_FLAG` reproduces this. A minimal example: two rails `VCC` and `GND`, each with a `power:PWR_FLAG` symbol on a stub wire connected to the rail label, and a resistor `R1` whose pin 1 is on `VCC` and pin 2 is on `GND`. Calling `get_net_connections(schematic, "VCC")` returns both `R1/1` *and* `R1/2`.

On a real-world schematic with seven `PWR_FLAG` symbols (one per named rail) this means that querying any one rail returns pins from all seven.

The official `kicad-cli sch export netlist` output shows the correct, distinct nets — so the bug is in the MCP server's wire-graph walking, not in the schematic.

## Root cause

`python/commands/wire_connectivity.py::_parse_virtual_connections` (lines 232–254 before this fix) treats `#FLG` symbols the same as `#PWR` symbols:

```python
ref = symbol.property.Reference.value
if not (ref.startswith("#PWR") or ref.startswith("#FLG")):
    continue
...
name = symbol.property.Value.value
...
point_to_label[pt] = name
label_to_points.setdefault(name, []).append(pt)
```

For a power-port (`#PWR*`) symbol, `Value` is the rail's net name (`+BATT`, `GND`, `+3V3`, ...) — correct.

For a power-flag (`#FLG*`) symbol, `Value` is always the literal string `"PWR_FLAG"`. So every pwr-flag in the schematic gets appended to `label_to_points["PWR_FLAG"]`.

`PWR_FLAG` is an ERC marker, not a net name — it inherits its name from whatever wire/label it sits on. Treating it as a label name lets `_find_connected_wires` use it as a BFS-via-label-jump bridge:

```python
# in _find_connected_wires
label_name = point_to_label.get(pt)
if label_name and label_name not in seen_labels:
    seen_labels.add(label_name)
    for other_pt in label_to_points.get(label_name, []):
        # virtually jump BFS to other_pt
```

When BFS for `+BATT` reaches its pwr-flag pin, `point_to_label[pt] == "PWR_FLAG"`. The BFS then jumps to every other pwr-flag pin position, walks across each one's stub wire, and reaches every other power rail.

## Fix

Pwr-flag pin positions must still register as anchors in `point_to_label` so `find_orphaned_wires` keeps accepting them (the original intent of commit 7f3a379) — but they must not enter `label_to_points`, which is what the BFS uses for virtual jumps. The pwr-flag remains electrically connected to its rail by the wire-graph BFS through the wire it sits on; the label-jump mechanism is unnecessary for that path.

See PR #177 for the patch and regression tests.

## Affected functions

- `get_net_connections` (visibly wrong via MCP)
- `generate_netlist` (calls `get_connections_for_net` for each net)
- Anything else that uses `_process_single_sheet` / `_find_connected_wires`

`find_orphaned_wires` is unaffected because it only reads `point_to_label`, not `label_to_points`.

## Real-world example

Schematic: a single-sheet board with seven distinct power rails (`+BATT`, `+3V3`, `GND`, `BATT_RAW`, `GPS_VBACKUP`, `V_A`, `V_B`) each with a `PWR_FLAG`.

Before the fix:
```
get_net_connections("+BATT") → 19 pins  (union of all seven rails)
get_net_connections("+3V3") → 56 pins  (same set)
get_net_connections("BATT_RAW") → 11 pins
```

After the fix:
```
get_net_connections("+BATT") → 8 pins   ✓ matches kicad-cli netlist (modulo a separate library-symbol pin-locator bug on diodes)
get_net_connections("+3V3") → 44 pins  ✓ same
get_net_connections("BATT_RAW") → 2 pins  ✓ exact match
get_net_connections("GATE_PROT") → 2 pins  ✓ exact match
```

(The few residual missing pins are diode pins from `PCM_Diode_Schottky_AKL:MBRS130`, a community symbol that defines pins 1 and 2 twice each. `PinLocator` returns the second definition which is an inner zero-length ghost. That is a separate bug.)

blinksoft:
# fix(wire_connectivity): pwr-flags must not bridge unrelated power rails

## What

`_parse_virtual_connections` no longer adds `#FLG` symbols to `label_to_points`. They remain in `point_to_label` (so `find_orphaned_wires` keeps accepting them, the original intent of 7f3a379), but they cannot serve as BFS-via-label-jump bridges keyed under the shared `"PWR_FLAG"` string.

Closes #176.

## Why

Every `power:PWR_FLAG` symbol carries `Value="PWR_FLAG"` — the ERC marker inherits its rail's name from the wire/label it sits on, it doesn't have one of its own. Since 7f3a379 added `#FLG` symbols to the same handling loop as `#PWR` power ports, every pwr-flag in the schematic was getting appended to `label_to_points["PWR_FLAG"]`. The BFS in `_find_connected_wires` uses `label_to_points` for virtual jumps when it encounters a labelled wire endpoint, so reaching any pwr-flag pin caused it to "teleport" to every other pwr-flag pin — and then walk across each one's stub wire into a different power rail.

Concretely, on a schematic with seven distinct rails each carrying a pwr-flag, `get_net_connections(rail)` returned the union of pins on all seven rails for *any* `rail`.

`find_orphaned_wires` is the only call site that needed pwr-flags as anchors and it only reads `point_to_label` — so we preserve its inputs and remove the harmful side effect.

## Tests

`tests/test_pwr_flag_no_bridge.py` adds three unit tests on `_parse_virtual_connections`:

1. `test_two_pwr_flags_do_not_share_a_label_to_points_entry` — proves the over-merge is gone. Two pwr-flags at distinct positions, both with `Value="PWR_FLAG"`, must not collide under a single `label_to_points` key.
2. `test_pwr_port_symbol_still_registers_in_both_maps` — regression check: `#PWR*` power ports continue to feed both maps, so BFS-via-label-jump still works for power rails with the same name across the sheet.
3. `test_pwr_flag_at_same_point_as_pwr_port_does_not_clobber_port_label` — edge case: if a pwr-flag and a power-port end up at the same point, `point_to_label[pt]` keeps the power-port's net name.

All three reliably fail on `main` (verified by stash + run), pass on this branch.

Full suite (modulo the pre-existing `test_get_pin_angle.py` collection error, which is unrelated to wire_connectivity): **667 passed, 11 skipped, 0 regressions**.

## End-to-end verification

Ran the patched `get_connections_for_net` directly against the real-world schematic that triggered the report (single-sheet board with 7 rails × pwr-flag each) and confirmed every nets pin-list now matches the official `kicad-cli sch export netlist` output, modulo a separate library-symbol bug on `PCM_Diode_Schottky_AKL:MBRS130` (duplicate pin definitions, ghost pin wins in PinLocator). That second bug is out of scope here — will be addressed in a follow-up PR.

## Diff

One file under `python/commands/wire_connectivity.py`, branch on the existing `#FLG` vs `#PWR` distinction:
- `#PWR*`: add to both maps (unchanged)
- `#FLG*`: add to `point_to_label` only, via `setdefault` so it can't clobber an upstream power-port label at the same point

Plus one new test file under `tests/test_pwr_flag_no_bridge.py`.

### DIFF
--- a/python/commands/wire_connectivity.py
+++ b/python/commands/wire_connectivity.py
@@ -236,7 +236,9 @@ def _parse_virtual_connections(
                 if not hasattr(symbol, "property") or not hasattr(symbol.property, "Reference"):
                     continue
                 ref = symbol.property.Reference.value
-                if not (ref.startswith("#PWR") or ref.startswith("#FLG")):
+                is_pwr_port = ref.startswith("#PWR")
+                is_pwr_flag = ref.startswith("#FLG")
+                if not (is_pwr_port or is_pwr_flag):
                     continue
                 if ref.startswith("_TEMPLATE"):
                     continue
@@ -248,8 +250,28 @@ def _parse_virtual_connections(
                     continue
                 pin_data = all_pins["1"]
                 pt = _to_iu(float(pin_data[0]), float(pin_data[1]))
-                point_to_label[pt] = name
-                label_to_points.setdefault(name, []).append(pt)
+
+                if is_pwr_port:
+                    # Power-port symbol: Value is the net name (+BATT, GND, ...).
+                    # Register in both maps so BFS-via-label-jump can bridge to
+                    # other instances of the same named power net.
+                    point_to_label[pt] = name
+                    label_to_points.setdefault(name, []).append(pt)
+                else:
+                    # Power-flag symbol (#FLG*): Value is always "PWR_FLAG" — an
+                    # ERC marker, not a net name. The pin position is registered
+                    # in point_to_label so find_orphaned_wires accepts wire ends
+                    # terminating on a PWR_FLAG as valid anchors. It is NOT added
+                    # to label_to_points: doing so would let BFS-via-label-jump
+                    # virtually bridge every distinct power rail that has a
+                    # pwr-flag into one mega-net (since every #FLG shares
+                    # Value="PWR_FLAG"). The pwr-flag remains electrically
+                    # connected to its rail via the wire-graph BFS through the
+                    # wire it sits on; the label-bridge mechanism is unneeded
+                    # and actively harmful here.
+                    # setdefault avoids clobbering an upstream power-port label
+                    # in the unlikely case that one sits at the same point.
+                    point_to_label.setdefault(pt, name)
             except Exception as e:
                 logger.warning(f"Error parsing power symbol: {e}")
 
--- a/tests/test_pwr_flag_no_bridge.py
+++ b/tests/test_pwr_flag_no_bridge.py
@@ -1,0 +1,210 @@
+"""
+Regression tests for the PWR_FLAG over-merge bug in _parse_virtual_connections.
+
+Background
+----------
+KiCad schematics use ``power:PWR_FLAG`` symbols (reference prefix ``#FLG``) as
+ERC markers that assert "this rail really is driven, do not warn." Every such
+symbol has ``Value="PWR_FLAG"`` regardless of which rail it sits on — it
+inherits the rail's name from the wire/label it is wired to.
+
+Commit 7f3a379 ("Treat PWR_FLAG anchors as connected in orphan-wire detection")
+added ``#FLG`` symbols to the same loop that handles ``#PWR`` power-port
+symbols inside ``_parse_virtual_connections``. The intent — registering the
+pwr-flag pin position so ``find_orphaned_wires`` would not flag wire ends
+terminating on a pwr-flag as dangling — was correct, but the implementation
+also added each pwr-flag pin to ``label_to_points`` keyed under the literal
+string ``"PWR_FLAG"``.
+
+This made ``label_to_points["PWR_FLAG"]`` a list of every pwr-flag pin in the
+sheet. In ``_find_connected_wires`` the BFS uses ``label_to_points`` for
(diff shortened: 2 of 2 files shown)


mixelpixx (approved):
Approving and merging.

The fix is surgical: splitting the `#PWR` and `#FLG` paths preserves the orphan-wire anchor benefit of commit 7f3a379 while removing the harmful `label_to_points["PWR_FLAG"]` bridge. The `setdefault` for the at-the-same-point edge case is exactly right — `test_pwr_flag_at_same_point_as_pwr_port_does_not_clobber_port_label` pins it.

The bug analysis in #176 and in this PR body is genuinely excellent — concrete repro, real-world net-count table, and a clean explanation of why `find_orphaned_wires` is unaffected (reads `point_to_label` only). Thanks for the careful work.

PR #179, opened 2026-05-14
fix(pin_locator): keep outer pin when a symbol defines a number twice

### LINKED ISSUE #178: PinLocator.parse_symbol_definition overwrites real pins with zero-length ghost duplicates
# `PinLocator.parse_symbol_definition` overwrites real pins with zero-length "ghost" duplicates

## Summary

When a KiCad symbol defines the same pin number more than once — as a few community-generated symbol libraries do — `PinLocator.parse_symbol_definition` keeps only the **last** definition encountered. If the last definition is an inner zero-length "ghost" pin used as a graphic-anchoring join, the real outer pin is overwritten and `get_pin_location` returns a coordinate that doesn't match any wire or label, breaking downstream connectivity tools (`get_schematic_pin_locations`, `get_net_connections`, `generate_netlist`).

## Affected symbol

The most visible case is `PCM_Diode_Schottky_AKL:MBRS130`, but the pattern applies to any symbol with duplicate pin-number definitions. The embedded `lib_symbols` block for MBRS130 looks like:

```
(pin passive line (at -3.81 0   0)  (length 2.54) (name "A") (number "2"))  ← outer pin 2, length 2.54
(pin passive line (at  3.81 0 180)  (length 2.54) (name "K") (number "1"))  ← outer pin 1, length 2.54
(pin passive line (at -2.54 -2.54 0) (length 0)   (name "A") (number "2"))  ← inner ghost
(pin passive line (at  2.54  2.54 180) (length 0) (name "K") (number "1"))  ← inner ghost
```

Both pin 1 and pin 2 are defined twice. The outer definitions have a visible stub (`length 2.54`) and their `at` coordinate is the wire-connection endpoint where users place labels and wires. The inner definitions are zero-length and live inside the symbol body.

## Root cause

`python/commands/pin_locator.py::parse_symbol_definition` (lines ~83–85 before this fix) walks `(pin …)` definitions recursively and stores each in a dict keyed by pin number:

```python
# Store by pin number
if pin_data["number"]:
    pins[pin_data["number"]] = pin_data
```

This is a plain assignment, so a duplicate-numbered pin encountered later in the walk overwrites the earlier one. For MBRS130, the inner zero-length ghosts are last in the sexp order, so the ghost wins and `pins["1"]` ends up pointing to the wrong coordinate.

`get_pin_location` then transforms the ghost's `at` through the symbol's translation/rotation/mirror and returns a coordinate that doesn't match anything in the user's schematic.

## Downstream impact

On a real-world schematic with the MBRS130 in a power chain (J1 → Q3 protection FET → D1 series Schottky → +3V3 rail loads), the resulting netlist gaps are:

| Net      | Expected (`kicad-cli`) | `get_connections_for_net` (broken) | Missing       |
|----------|------------------------|------------------------------------|---------------|
| `+BATT`  | 9 nodes                | 8 nodes                            | `D1/1`        |
| `+3V3`   | 46 nodes               | 44 nodes                           | `D1/2`, `D2/2`|

The diode pins simply don't appear on the rails the wires put them on. `get_schematic_pin_locations` for D1 returns the ghost coordinates (e.g. `(116.84, 48.26)` instead of the real `(118.11, 50.80)`).

## Fix

When storing a duplicate-numbered pin, keep the entry with the greater `length`. The outer "real" pin has `length > 0` (a visible stub out to the wire-attach point); the inner ghost has `length == 0`. Strict-greater comparison resolves ties to first-encountered, so legitimate same-length duplicates (e.g. per-unit repetitions in multi-unit symbols) keep stable existing behaviour.

See PR #179 for the patch and regression tests.

## Real-world before/after

End-to-end on the schematic that triggered the report:

```
Before: get_connections_for_net("+BATT") → 8 nodes  (D1/1 missing)
After:  get_connections_for_net("+BATT") → 9 nodes  ✓ exact match to kicad-cli
```

(Note: combine this fix with PR #177, which addresses an orthogonal PWR_FLAG-bridging over-merge in `wire_connectivity.py`, to get full parity with `kicad-cli sch export netlist` on schematics that use both pwr-flag markers and community diode symbols.)

blinksoft:
# fix(pin_locator): keep outer pin when a symbol defines a number twice

## What

`parse_symbol_definition` no longer overwrites a real outer pin definition with a duplicate inner zero-length ghost. When two `(pin …)` entries in a single symbol share a number, the one with the greater `length` wins; ties resolve to first-encountered.

Closes #178.

## Why

Some community-generated symbol libraries (notably `PCM_Diode_Schottky_AKL:MBRS130`, but the pattern is general) define each pin number twice — once as a visible outer pin with a stub (`length > 0`) and once as an inner zero-length "ghost" used as a graphic-anchoring join. Recursive walking through `lib_symbols` encounters both definitions, and `pins[number] = pin_data` is a plain overwrite. For MBRS130 the ghost appears last in sexp order, so the ghost wins, and `get_pin_location` returns a coordinate that doesn't match where the schematic author placed labels and wires.

Knock-on: `get_connections_for_net` silently drops diode pins from their actual rails. On the report-triggering schematic, `+BATT` and `+3V3` queries returned 8/9 and 44/46 pins respectively — exactly the diode pins missing.

## Heuristic choice

`length` is the discriminator because:
- The outer "real" pin has length > 0 (a visible stub running out to the wire-attach point that the schematic author actually wires to).
- The inner "ghost" pin has length == 0 (no visible stub, used only as a graphic-anchoring junction inside the symbol body).
- A pin's electrical connection point in KiCad is its `at` coordinate; `length` is purely the visible stub. For symbols with a single legitimate definition per number, length is unchanged either way and the behaviour is identical.

The comparison is strict-greater (not `>=`) so equal-length duplicates retain the first-encountered entry. That preserves behaviour for legitimate same-length cases — for example a multi-unit symbol whose per-unit blocks each redeclare the same pin number with the same stub length — and matches what pre-fix code did for the only case it handled correctly.

## Tests

`tests/test_pin_locator_duplicate_pin_defs.py` adds four unit tests on `PinLocator.parse_symbol_definition`:

1. `test_outer_pin_wins_when_ghost_appears_after` — mirrors the real MBRS130 ordering. The outer pin is defined first, ghost second. Before the fix the ghost clobbered the outer pin; after, the outer pin survives.
2. `test_outer_pin_wins_when_ghost_appears_first` — symmetry check. Ghost defined first, outer second. The length-based heuristic must win regardless of order; on pre-fix code the outer would have won "for the wrong reason" (last-wins) but on this branch it wins because length > 0 beats length == 0.
3. `test_no_duplicates_unaffected` — regression: a normal symbol (Device:R with unique pin numbers) stores the same data it always did.
4. `test_equal_length_duplicates_keep_first_encountered` — ties resolve to first-encountered, preserving stability for multi-unit-style same-length duplicates. On pre-fix code last-wins, so this test fails on `main` and passes on this branch.

Two of the four fail on `main` (verified by `git stash` + run), all four pass on this branch.

Full suite (modulo the pre-existing `tests/test_get_pin_angle.py` collection error, which is unrelated): **671 passed, 11 skipped, 0 regressions**.

## End-to-end verification

Ran the patched `get_connections_for_net` against the real schematic that triggered the report (single-sheet altimeter with seven power rails, two MBRS130 Schottky diodes in the power path):

```
+BATT  expected 9 got 9   ✓ exact match to kicad-cli sch export netlist
+3V3   expected 46 got 46  ✓ exact match
```

Net membership now matches `kicad-cli` output exactly on this schematic when combined with PR #177 (the orthogonal PWR_FLAG-bridging fix in `wire_connectivity.py`).

## Diff

One file under `python/commands/pin_locator.py`, ~5 lines changed: replace the plain `pins[number] = pin_data` assignment with a length-aware setdefault-like check.

Plus one new test file under `tests/test_pin_locator_duplicate_pin_defs.py`.

### DIFF
--- a/python/commands/pin_locator.py
+++ b/python/commands/pin_locator.py
@@ -80,9 +80,21 @@ class PinLocator:
                         elif item[0] == Symbol("number") and len(item) >= 2:
                             pin_data["number"] = str(item[1]).strip('"')
 
-                # Store by pin number
+                # Store by pin number. When the same pin number is defined
+                # more than once in a single symbol — which happens in some
+                # community-generated symbols (e.g.,
+                # ``PCM_Diode_Schottky_AKL:MBRS130``) where an inner
+                # zero-length "ghost" pin overlaps the real outer pin — keep
+                # the definition with the greater ``length``. That is the pin
+                # with a visible stub; its ``at`` coordinate is the wire-
+                # connection endpoint that matches where labels and wires
+                # are actually placed. Ties resolve to first-encountered, so
+                # legitimate same-length duplicates (e.g., per-unit
+                # repetitions in multi-unit symbols) retain stable ordering.
                 if pin_data["number"]:
-                    pins[pin_data["number"]] = pin_data
+                    existing = pins.get(pin_data["number"])
+                    if existing is None or pin_data["length"] > existing["length"]:
+                        pins[pin_data["number"]] = pin_data
 
             # Recurse into sublists
             for item in sexp:
--- a/tests/test_pin_locator_duplicate_pin_defs.py
+++ b/tests/test_pin_locator_duplicate_pin_defs.py
@@ -1,0 +1,153 @@
+"""
+Regression tests for ``PinLocator.parse_symbol_definition`` when a symbol
+defines the same pin number more than once.
+
+Background
+----------
+Some community-generated symbol libraries — for example
+``PCM_Diode_Schottky_AKL:MBRS130`` — include both an outer "real" pin
+with a visible stub (non-zero ``length``) and an inner zero-length "ghost"
+pin at a different ``at`` coordinate. Both definitions share the same pin
+number. Conceptually the ghost is an internal join used for symbol
+graphic anchoring; the real outer pin is where wires and labels are
+placed by the schematic author.
+
+Before this fix, ``parse_symbol_definition`` stored pins as
+``pins[pin_data["number"]] = pin_data`` — a plain assignment. Each
+duplicate-numbered definition encountered during recursion overwrote the
+previous one. The recursion order put the ghost pins last for the MBRS130
+symbol, so the ghost won and ``get_pin_location`` returned a coordinate
+that did not match any wire/label. As a knock-on effect,
+``get_connections_for_net`` failed to discover diode pins on the rails
+they were wired to (e.g. ``D1/1`` on ``+BATT``).
+
+The fix: when storing a duplicate-numbered pin, keep the entry with the
+greater ``length``. The outer "real" pin has length > 0 (a visible stub
+out to the wire-attach point); the inner ghost has length == 0. Ties
+resolve to first-encountered, so legitimate same-length duplicates
+(e.g., per-unit repetitions in multi-unit symbols) keep stable ordering
+and existing behaviour.
+"""
+
+import sys
+from pathlib import Path
+
+import pytest
+from sexpdata import Symbol
+
+sys.path.insert(0, str(Path(__file__).parent.parent / "python"))
+
+from commands.pin_locator import PinLocator  # noqa: E402
+
+# ---------------------------------------------------------------------------
+# Helper: build a (symbol …) sexp matching the structure KiCad writes
+# ---------------------------------------------------------------------------
+
+
+def _pin_sexp(number: str, name: str, x: float, y: float, angle: int, length: float):
+    """Return an s-expression list that mimics a KiCad ``(pin …)`` definition."""
+    return [
+        Symbol("pin"),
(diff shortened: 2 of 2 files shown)


mixelpixx (approved):
Approving and merging.

Five-line fix with the right heuristic: length>0 wins over length==0, strict-greater so legitimate same-length duplicates (multi-unit symbol per-unit blocks) keep first-encountered ordering and pre-fix behaviour. The four tests cover both ghost-orderings, the no-duplicate baseline, and the equal-length tie — exactly the cases that matter.

The MBRS130 root-cause write-up in #178 plus the end-to-end net-count parity here is a clean piece of work. Thanks!

Issue #181, opened 2026-05-17
[Bug] Huge amounts of logs being generated in the .kicad-mcp/logs directory


angelorodem:
I've been looking today, my C drive got full, when checking, after a single session, the logs stored on the directory `C:\Users\<user>\.kicad-mcp\logs\kicad_interface.log` were consuming **multiple gigabytes**. just issuing the mcp command "list_schematic_nets" was able to generate up to 200mb of logs in about 1 minute, there is something broken with logging.

my mcp.json is
```
{
  "servers": {
    "KiCAD-MCP-Server": {
      "type": "stdio",
      "command": "node",
      "args": ["E:\\git\\KiCAD-MCP-Server\\dist\\index.js"],
      "env": {
        "NODE_ENV": "production",
        "KICAD_PYTHON": "D:\\KiCad9\\bin\\python.exe",
        "PYTHONPATH": "D:\\KiCad9\\bin\\Lib\\site-packages",
        "KICAD_AUTO_LAUNCH": "false",
      }
    }
  }
}
```

i've also tried `"KICAD_MCP_DEV": "0"` the logs did not stop being recorded to this folder, it seems that the logs on this folder in windows is saved anyway.

on the schematic, it is not a big schematic, there is few components, but multiple hierarchical sheets.

i'm suspecting that the tool is going into an infinite loop when reading the nets on hierarchical sheets, but here is the analysis made about the `kicad_interface.log` generated:

```
I investigated the log growth and found a clear pattern.

A single `list_schematic_nets` request against `left.kicad_sch` generated approximately 923,841 log lines in the first 30 seconds, and 923,839 of them were `DEBUG`. The output is dominated by parser-internal messages from `skip.sexp.parser`, especially repeated lines like `Parse subentry`, `GOT NAME`, `Len of tree is`, and `Appended entry at`.

This does not look like a simple startup burst. The same parser-debug storm continues minutes later in the tail of the log.

There also seems to be repeated hierarchy traversal during that one request:
- `left.kicad_sch` was read 5 times in the first 30 seconds
- `switch.kicad_sch` was read 140 times in the first 30 seconds

That lines up with the schematic structure:
- `left.kicad_sch` contains 35 instances of `switch.kicad_sch`

From the server code, there appear to be two separate issues compounding each other:

1. Logging is globally too verbose
- The Python interface configures root logging at `DEBUG`
[...]

PR #182, opened 2026-05-17
feat: enhance wire connectivity analysis and logging configuration

angelorodem:
- Added SheetConnectivity dataclass for improved schematic data handling.
- Implemented caching for S-expression loading to optimize performance.
- Introduced environment-based logging level configuration in kicad_interface.py.
- Updated config.ts to support dynamic log level adjustments from environment variables.
- Added unit tests for net traversal efficiency and caching behavior in wire connectivity.

Fixes issue #181

### DIFF
--- a/python/commands/wire_connectivity.py
+++ b/python/commands/wire_connectivity.py
@@ -10,6 +10,7 @@ sub-sheet files and bridging nets via hierarchical labels / sheet pins.
 """
 
 import logging
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Tuple
 
@@ -20,6 +21,22 @@ from sexpdata import Symbol
 logger = logging.getLogger("kicad_interface")
 
 _IU_PER_MM = 10000  # KiCad schematic internal units per millimeter
+_SEXP_CACHE: Dict[str, Tuple[int, int, list]] = {}
+
+
+@dataclass
+class SheetConnectivity:
+    """Parsed schematic data reused while answering a single net-list request."""
+
+    schematic: Any
+    schematic_path: str
+    sexp: list
+    all_wires: List[List[Tuple[int, int]]]
+    adjacency: List[Set[int]]
+    iu_to_wires: Dict[Tuple[int, int], Set[int]]
+    point_to_label: Dict[Tuple[int, int], str]
+    label_to_points: Dict[str, List[Tuple[int, int]]]
+    symbol_instances: List[Dict]
 
 
 def _to_iu(x_mm: float, y_mm: float) -> Tuple[int, int]:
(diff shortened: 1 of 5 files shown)


mixelpixx:
Thanks for tackling this, angelorodem — the log spam in #181 has been hurting users and the analysis in the issue is spot-on.

Before merging I'd like to ask you to split this PR into two. The reason is that the diff currently bundles two changes with very different risk profiles, and we're trying hard to avoid regressions on the main branch.

**What I'd like as PR A (merge first):**
- `python/kicad_interface.py` — the `_parse_log_level` / `_parse_positive_int_env` / `_env_flag_enabled` helpers, the `RotatingFileHandler` setup, and the `skip.*` logger muting.
- `src/config.ts` — the `LogLevelSchema`, `getEnvLogLevel`, and `applyEnvironmentOverrides` changes.
- `python/commands/wire_connectivity.py` — only the `_load_sexp` LRU cache (the `_SEXP_CACHE` OrderedDict, mtime+size invalidation, size-128 cap).
- Tests covering the cache hit/miss and log-rotation behavior.

This piece directly fixes #181, is well-bounded, has no observable API change, and we can get it done quickly.

**What I'd like as PR B (held for design discussion):**
- The `SheetConnectivity`, `SheetReference`, `SheetPinReference`, `SheetTraversal` dataclasses.
- `_parse_sheet_references`, `collect_sheet_traversals`, `_build_child_net_name_map`, `pins_for_traversal_net`.
- The rewrite of `get_connections_for_net` and `_handle_list_schematic_nets`.
- The `formatConnection` change in `src/tools/schematic.ts` that appends `@ sheet_instance`.

Two reasons this piece needs to be separate:

1. **It overlaps with #169 and #170** (inktomi's hierarchical sub-sheet PRs). We have three open PRs touching hierarchical sheet semantics in the same code paths. We need to pick a single design rather than merge them in arrival order and discover conflicts later.
2. **It changes the user-visible output shape** of net connections. The old form is `"Component/Pin"`; the new form is `"Component/Pin @ sheet_instance"` when an instance is present. Any client parsing the connection strings (including the AI clients this server is built for) will see a behavior change. The fallback to the old shape on flat schematics is good, but we should still consider this a feature, not a bugfix, and document it explicitly.

Once PR A is in, we can review PR B alongside #169 and #170 together. Happy to help with the split if rebasing is awkward — let me know.

mixelpixx:
Checking in on this one. The log-spam fix in here (#181) is valuable and several users are waiting on it. Still hoping to take it as the two-part split we discussed: **PR A** = the log-rotation + `skip.*` muting + the `_load_sexp` LRU cache (lands fast, fixes #181), and **PR B** = the `SheetConnectivity` hierarchy rewrite held for the #169/#170 design discussion. If splitting the branch is a hassle, say the word and I can help carve out PR A. Otherwise no pressure — just keeping it visible.

angelorodem:
Sorry, i have no bandwidth to split in two, i could use a hand on this, i would appreciate.

mixelpixx:
Thanks for the offer — I carved the log fix out for you in #215 (you're credited as co-author). It takes the logging half of this PR verbatim: the `_parse_log_level` / env helpers + `RotatingFileHandler` + skip-logger muting in `kicad_interface.py`, and the `getEnvLogLevel` / `applyEnvironmentOverrides` in `config.ts`. I also added a matching size cap to the TypeScript per-day logs, a test, and a README note. It closes #181.

I deliberately left out the `SheetConnectivity` / hierarchical-sheet rewrite and the `_load_sexp` cache — those are the part that overlaps with #169/#170 and changes the net-connection output shape, so they should stay here (this PR) for the separate design discussion. So once #215 lands, this PR becomes just that hierarchy work (PR B). No pressure on timing for it.

mixelpixx:
Status: the logging half of this landed via #215 (co-credited). The wire-connectivity rewrite remains on hold pending the #169/#170 hierarchy design discussion — we don't want to merge a connectivity model that conflicts with where that lands.

mixelpixx:
Closing this one out, with thanks — and to be clear about why, since none of it is a reflection on the work.

Where things ended up:

- **The logging half landed.** The `_parse_log_l...` reduction and the logging-configuration changes went in via #215 back in June, with you credited as co-author. That was the part several users were actively waiting on (#181), and it is shipped.
- **The wire-connectivity half has been overtaken.** When you opened this in May it was ahead of where `main` was. Since then the connectivity and sheet-traversal code has been reworked on `main` and in the in-flight #314 hardening series, and several of the helpers this PR adds now already exist upstream under different names. Rebasing this branch would no longer be a merge so much as a rewrite against code that has moved underneath it — the branch currently conflicts, and resolving it would mean re-deciding each hunk against a different design than the one it was written for.

You told us in May you did not have bandwidth to split this, which was a completely fair answer, and carving out #215 was our way of making sure the valuable part did not get stuck behind that. It did not.

So this is being closed as superseded rather than rejected. The history stays here and the branch remains referenceable if any of the remaining connectivity ideas turn out to be worth lifting — if you (or anyone) later spot behavior in here that `main` still gets wrong, a small focused PR against current `main` is very welcome and will get a much faster path through review than a rebase of this would.

Thanks for the original report and the work on it.

PR #187, opened 2026-05-18
fix: resolve `${KICAD_3RD_PARTY}` (no version prefix) in lib URI resolvers

### LINKED ISSUE #153: lib-table URI resolver doesn't substitute `${KICAD_3RD_PARTY}` (no version prefix)
## Summary

Lib-table rows that use `${KICAD_3RD_PARTY}` (no KiCad-version prefix) in their `uri` field never resolve, so the libraries are silently dropped from `list_symbol_libraries` / `list_libraries` results — even when they are correctly registered in KiCad's lib-tables and visible in the KiCad GUI.

## Why this matters

The [Import-LIB-KiCad-Plugin (impart)](https://github.com/Steffen-W/Import-LIB-KiCad-Plugin) documentation explicitly registers its third-party libraries using `${KICAD_3RD_PARTY}` without a version prefix:

| Active | Visible | Nickname | Library Path |
| --- | --- | --- | --- |
| ✓ | ✓ | Samacsys | `${KICAD_3RD_PARTY}/Samacsys.kicad_sym` |
| ✓ | ✓ | Snapeda | `${KICAD_3RD_PARTY}/Snapeda.kicad_sym` |
| ✓ | ✓ | UltraLibrarian | `${KICAD_3RD_PARTY}/UltraLibrarian.kicad_sym` |
| ✓ | ✓ | EasyEDA | `${KICAD_3RD_PARTY}/EasyEDA.kicad_sym` |

KiCad itself accepts both unprefixed and version-prefixed forms in lib-tables. The MCP's `library.py` already handles the analogous `KICAD_SYMBOL_DIR` (no version prefix) — `KICAD_3RD_PARTY` is just missing from the same place by oversight.

## Root cause

`_resolve_uri` in both [`python/commands/library_symbol.py`](https://github.com/mixelpixx/KiCAD-MCP-Server/blob/main/python/commands/library_symbol.py) and [`python/commands/library.py`](https://github.com/mixelpixx/KiCAD-MCP-Server/blob/main/python/commands/library.py) builds an env-var substitution dictionary that contains:

```python
"KICAD10_3RD_PARTY": ...,
"KICAD9_3RD_PARTY":  ...,
"KICAD8_3RD_PARTY":  ...,
# but no "KICAD_3RD_PARTY" entry
```

So `${KICAD_3RD_PARTY}/Foo.kicad_sym` is never substituted, the resulting literal path doesn't exist, and the library is treated as missing. The same is true for the corresponding `_find_3rd_party_dir` / `_find_kicad_3rdparty_dir` helpers, which only check `KICAD8/9/10_3RD_PARTY` in `os.environ`.

## Reproduction

1. Set `KICAD_3RD_PARTY` (no version prefix) in the MCP server's environment, e.g. via `.mcp.json`:
   ```json
   "env": { "KICAD_3RD_PARTY": "/Users/me/KiCad/lib3rdparty" }
   ```
2. Place a real `EasyEDA.kicad_sym` at `/Users/me/KiCad/lib3rdparty/EasyEDA.kicad_sym` containing at least one symbol.
3. Add this row to the global `sym-lib-table`:
   ```
   (lib (name "EasyEDA") (type "KiCad") (uri "${KICAD_3RD_PARTY}/EasyEDA.kicad_sym") (options "") (descr ""))
   ```
4. Restart Claude Code (so the MCP server reloads its env).
5. Call `list_symbol_libraries` via the MCP. **Expected:** `EasyEDA` appears in the list. **Actual:** It's absent. The same row using `${KICAD10_3RD_PARTY}/EasyEDA.kicad_sym` (with `KICAD10_3RD_PARTY` env var set) **does** appear.

## Proposed fix

Add `KICAD_3RD_PARTY` to both `_resolve_uri` dicts and to the `_find_*_3rdparty_dir` helpers' `os.environ` checks. Tested locally; full diff (3 files including a regression test):

<details>
<summary>Diff</summary>

```diff
diff --git a/python/commands/library.py b/python/commands/library.py
@@ -145,6 +145,7 @@ class LibraryManager:
             "KICAD10_3RD_PARTY": self._find_kicad_3rdparty_dir(),
             "KICAD9_3RD_PARTY": self._find_kicad_3rdparty_dir(),
             "KICAD8_3RD_PARTY": self._find_kicad_3rdparty_dir(),
+            "KICAD_3RD_PARTY": self._find_kicad_3rdparty_dir(),
         }

@@ -205,10 +206,11 @@ class LibraryManager:
-        if "KICAD9_3RD_PARTY" in os.environ:
-            path = os.environ["KICAD9_3RD_PARTY"]
-            if os.path.isdir(path):
-                return path
+        for var in ("KICAD10_3RD_PARTY", "KICAD9_3RD_PARTY", "KICAD8_3RD_PARTY", "KICAD_3RD_PARTY"):
+            if var in os.environ:
+                path = os.environ[var]
+                if os.path.isdir(path):
+                    return path

diff --git a/python/commands/library_symbol.py b/python/commands/library_symbol.py
@@ -164,6 +164,7 @@ class SymbolLibraryManager:
             "KICAD10_3RD_PARTY": self._find_3rd_party_dir(),
             "KICAD9_3RD_PARTY": self._find_3rd_party_dir(),
             "KICAD8_3RD_PARTY": self._find_3rd_party_dir(),
+            "KICAD_3RD_PARTY": self._find_3rd_party_dir(),
             "KISYSSYM": self._find_kicad_symbol_dir(),
         }

@@ -227,6 +228,8 @@ class SymbolLibraryManager:
             possible_paths.insert(0, os.environ["KICAD9_3RD_PARTY"])
         if "KICAD8_3RD_PARTY" in os.environ:
             possible_paths.insert(0, os.environ["KICAD8_3RD_PARTY"])
+        if "KICAD_3RD_PARTY" in os.environ:
+            possible_paths.insert(0, os.environ["KICAD_3RD_PARTY"])

         for path in possible_paths:
             if os.path.isdir(path):
```

Plus a regression test at `tests/test_kicad_3rd_party_env_resolution.py` covering both managers (3 cases: symbol, footprint, both unprefixed and version-prefixed forms working when env vars are set).
</details>

I have the patch + tests committed locally on a branch and ready to send as a PR if you'd like — happy to fork and open one. Verified locally on macOS / KiCad 10 / Python 3.9 / MCP `main @ 216d3f1`. Tests pass:

```
tests/test_kicad_3rd_party_env_resolution.py ...   [100%]
3 passed in 0.04s
```

## Environment

- macOS 15.4 (arm64)
- KiCad 10.0.1
- KiCAD MCP Server `main` at commit `216d3f1`
- Python 3.9.13 (KiCad-bundled)
- Node 25.9.0

gsdali:
Fixes #153.

## Summary

- `${KICAD_3RD_PARTY}` (without a KiCad-version prefix) is the form documented by [Import-LIB-KiCad-Plugin (impart)](https://github.com/Steffen-W/Import-LIB-KiCad-Plugin) for registering its third-party libraries in `sym-lib-table` and `fp-lib-table`. KiCad's GUI accepts both unprefixed and version-prefixed forms.
- The MCP's `_resolve_uri` dictionaries in `python/commands/library.py` and `python/commands/library_symbol.py` only had `KICAD8/9/10_3RD_PARTY` — `KICAD_3RD_PARTY` was missing by oversight. The same gap existed in the corresponding `_find_*_3rdparty_dir` helpers.
- Result: lib-table rows using `${KICAD_3RD_PARTY}/Foo.kicad_sym` (or `.pretty`) never resolved, so the libraries silently disappeared from `list_symbol_libraries` / `list_libraries` even though they were registered correctly in KiCad and visible in the GUI.

## Changes

- `python/commands/library.py`:
  - Add `KICAD_3RD_PARTY` to the `_resolve_uri` env-var substitution dict.
  - Refactor `_find_kicad_3rdparty_dir` to check all four env-var forms (`KICAD10/9/8_3RD_PARTY` + `KICAD_3RD_PARTY`) in a single loop instead of repeated `if` blocks.
- `python/commands/library_symbol.py`:
  - Add `KICAD_3RD_PARTY` to the `_resolve_uri` env-var substitution dict.
  - Add `KICAD_3RD_PARTY` to the `_find_3rd_party_dir` `os.environ` checks.
- `tests/test_kicad_3rd_party_env_resolution.py` (new): regression tests covering both `LibraryManager` and `SymbolLibraryManager`, both unprefixed and version-prefixed env-var forms.

## Test plan

- [x] `pytest tests/test_kicad_3rd_party_env_resolution.py` — 3 passed (per the issue's local verification)
- [x] Reproduction from the issue: set `KICAD_3RD_PARTY` env var, register `(lib (name "EasyEDA") (type "KiCad") (uri "${KICAD_3RD_PARTY}/EasyEDA.kicad_sym") ...)` in `sym-lib-table`, restart Claude Code — `EasyEDA` now appears in `list_symbol_libraries`.
- [ ] Existing test suite stays green.

## Environment verified

macOS 15.4 (arm64), KiCad 10.0.1, MCP `main` @ 216d3f1 (the merge base for this branch), Python 3.9.13 (KiCad-bundled), Node 25.9.0.

### DIFF
--- a/python/commands/library.py
+++ b/python/commands/library.py
@@ -145,6 +145,7 @@ class LibraryManager:
             "KICAD10_3RD_PARTY": self._find_kicad_3rdparty_dir(),
             "KICAD9_3RD_PARTY": self._find_kicad_3rdparty_dir(),
             "KICAD8_3RD_PARTY": self._find_kicad_3rdparty_dir(),
+            "KICAD_3RD_PARTY": self._find_kicad_3rdparty_dir(),
         }
 
         # Project directory
@@ -205,10 +206,11 @@ class LibraryManager:
         import json
 
         # 1. Check shell environment variable first
-        if "KICAD9_3RD_PARTY" in os.environ:
-            path = os.environ["KICAD9_3RD_PARTY"]
-            if os.path.isdir(path):
-                return path
+        for var in ("KICAD10_3RD_PARTY", "KICAD9_3RD_PARTY", "KICAD8_3RD_PARTY", "KICAD_3RD_PARTY"):
+            if var in os.environ:
+                path = os.environ[var]
+                if os.path.isdir(path):
+                    return path
 
         # 2. Check kicad_common.json for user-defined variables
         kicad_common_paths = [
--- a/python/commands/library_symbol.py
+++ b/python/commands/library_symbol.py
@@ -164,6 +164,7 @@ class SymbolLibraryManager:
             "KICAD10_3RD_PARTY": self._find_3rd_party_dir(),
             "KICAD9_3RD_PARTY": self._find_3rd_party_dir(),
             "KICAD8_3RD_PARTY": self._find_3rd_party_dir(),
+            "KICAD_3RD_PARTY": self._find_3rd_party_dir(),
             "KISYSSYM": self._find_kicad_symbol_dir(),
         }
 
@@ -227,6 +228,8 @@ class SymbolLibraryManager:
             possible_paths.insert(0, os.environ["KICAD9_3RD_PARTY"])
         if "KICAD8_3RD_PARTY" in os.environ:
             possible_paths.insert(0, os.environ["KICAD8_3RD_PARTY"])
+        if "KICAD_3RD_PARTY" in os.environ:
+            possible_paths.insert(0, os.environ["KICAD_3RD_PARTY"])
 
         for path in possible_paths:
             if os.path.isdir(path):
--- a/tests/test_kicad_3rd_party_env_resolution.py
+++ b/tests/test_kicad_3rd_party_env_resolution.py
@@ -1,0 +1,75 @@
+"""
+Regression tests for ``${KICAD_3RD_PARTY}`` (no-version-prefix) URI resolution
+in both the symbol and footprint library managers.
+
+Background:
+    The Import-LIB-KiCad-Plugin (impart) documentation registers third-party
+    libraries using ``${KICAD_3RD_PARTY}`` without a KiCad-version prefix.
+    KiCad accepts both unprefixed and version-prefixed forms in lib-tables.
+
+    Prior to this fix the env-var dictionaries in ``_resolve_uri`` only handled
+    ``KICAD8/9/10_3RD_PARTY`` (analogous to ``KICAD_SYMBOL_DIR`` which *was*
+    in the dictionary without a prefix). Lib-table rows authored as
+    ``${KICAD_3RD_PARTY}/Foo.kicad_sym`` or ``${KICAD_3RD_PARTY}/Foo.pretty``
+    therefore failed to resolve and disappeared from MCP listings even though
+    KiCad's GUI showed them correctly.
+"""
+
+import sys
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).parent.parent / "python"))
+
+from commands.library import LibraryManager
+from commands.library_symbol import SymbolLibraryManager
+
+
+@pytest.mark.unit
+def test_symbol_manager_resolves_unprefixed_kicad_3rd_party(monkeypatch, tmp_path):
+    monkeypatch.setenv("KICAD_3RD_PARTY", str(tmp_path))
+    target = tmp_path / "EasyEDA.kicad_sym"
+    target.touch()
+
+    manager = SymbolLibraryManager.__new__(SymbolLibraryManager)
+    manager.project_path = None
+
+    resolved = manager._resolve_uri("${KICAD_3RD_PARTY}/EasyEDA.kicad_sym")
+    assert resolved == str(target)
+
+
+@pytest.mark.unit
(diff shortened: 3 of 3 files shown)


mixelpixx (approved):
Approving and merging. Thanks for opening the PR — the analysis in #153 was exactly right and the implementation matches it precisely.

Two things I like about the diff:

1. **`KICAD_3RD_PARTY` added to both managers** in a single, consistent shape — matches the existing handling for the analogous `KICAD_SYMBOL_DIR` (also unprefixed by KiCad convention).
2. **The `_find_kicad_3rdparty_dir` refactor** also fixes a pre-existing bug: the function had a `_resolve_uri` dict claiming to support `KICAD8/9/10_3RD_PARTY` but the env-var check only looked at `KICAD9_3RD_PARTY`. The new loop covers all four forms correctly. Good catch.

Three regression tests covering both managers, both unprefixed and version-prefixed forms, and the simultaneous case is the right coverage. Closes #153 cleanly. Thanks!

PR #188, opened 2026-05-18
test(pin_locator): add asymmetric multi-pin Y-flip regression (#135)

### LINKED ISSUE #135: get_schematic_pin_locations returns Y-flipped coordinates for placed symbols
# Bug Report: Y-flipped coordinates in `get_schematic_pin_locations`

**Target repo:** https://github.com/mixelpixx/KiCAD-MCP-Server
**File issue at:** https://github.com/mixelpixx/KiCAD-MCP-Server/issues/new

---

## Title

`get_schematic_pin_locations` returns Y-flipped coordinates for placed symbols

### Summary

`get_schematic_pin_locations` returns wrong Y coordinates for pins on placed
symbols. The result is mirrored vertically around the symbol's center, so
labels placed using these coordinates land on the wrong pin (or in empty
space) and netlists end up with shorted/wrong nets.

### Root cause

KiCad symbol library definitions use a **Y-up** coordinate convention, but
schematic file coordinates are **Y-down**. The conversion in
`python/commands/pin_locator.py::get_pin_location` adds `pin_rel_y` instead
of subtracting it (or negating before rotation):

```python
# Current (line ~329)
abs_x = symbol_x + pin_rel_x
abs_y = symbol_y + pin_rel_y   # ← bug
```

For symmetric symbols (R, C) this happens to produce a coordinate that's
still on a pin (just with pin numbers swapped), so it appears to work.
For asymmetric symbols (ESP32-WROOM-32, SN65HVD230, CP2102, USB-C
receptacle, microSD socket, etc.) every pin lands on a different pin or
on empty space, breaking `add_schematic_net_label` workflows.

### Reproduction

1. Add `RF_Module:ESP32-WROOM-32` to a schematic at (150, 110)
2. Call `get_schematic_pin_locations` for the ESP32
3. Tool reports Pin 3 (EN) at (134.76, 140.48)
4. Open the schematic — Pin 3 is actually at (134.76, 79.52)
5. KiCad CLI ERC confirms: pin is at 79.52, label at 140.48 is dangling

### Fix

Negate `pin_rel_y` before applying rotation so the rotation direction is
correct in screen coordinates as well:

```python
pin_rel_y = -pin_rel_y
if symbol_rotation != 0:
    pin_rel_x, pin_rel_y = self.rotate_point(pin_rel_x, pin_rel_y, symbol_rotation)
abs_x = symbol_x + pin_rel_x
abs_y = symbol_y + pin_rel_y
```

Verified against KiCad 10.0.0 with ESP32-WROOM-32, SN65HVD230, CP2102N,
USB_C_Receptacle, Micro_SD_Card — pin positions now match what KiCad's
ERC reports.

### Environment

- KiCad 10.0.0 (Windows)
- KiCAD-MCP-Server v2.2.3
- Python 3.11.5 (KiCad bundled)

---

## How to file

After running `gh auth login`:

```bash
gh issue create \
  --repo mixelpixx/KiCAD-MCP-Server \
  --title "get_schematic_pin_locations returns Y-flipped coordinates for placed symbols" \
  --body-file K:/AI_KiCad_ForgeTune/KiCAD-MCP-Server/PIN_LOCATOR_BUG_REPORT.md
```

Or paste the title and body manually at the issues URL above.

NiNjA-CodE:
Fixes #135.

## Context

The underlying Y-flip fix already landed in #179 — `WireDragger.pin_world_xy`
now does `ly = -py` (lib Y-up → screen Y-down) before mirror/rotation, which
is exactly what #135 identified. This PR fills the test gap the maintainer
asked for in [issue #135](https://github.com/mixelpixx/KiCAD-MCP-Server/issues/135#issuecomment-4480320123):

> Please open a PR with the fix... include a regression test against an
> asymmetric symbol (the ESP32-WROOM-32 case you used, or something similar)

The two existing `test_pin_locator_y_flip.py` tests use `Device:R` and
`Device:C` rotated — both **pin-symmetric**, so they pass even on the
pre-fix code path because swapping pin 1 ↔ pin 2 in the netlist is
electrically indistinguishable. The bug only manifests on asymmetric
multi-pin parts (`RF_Module:ESP32-WROOM-32` was the original report;
`Interface_CAN_LIN:SN65HVD230`, `Connector_USB:USB_C_Receptacle_USB2.0_16P`
all reproduce it).

## What this adds

A third unit test in `tests/test_pin_locator_y_flip.py`:
`test_asymmetric_multi_pin_symbol_y_flip`.

It builds a 6-pin asymmetric symbol **inline** (no system-library
dependency, so it runs in any CI environment without KiCad's `Symbols`
package installed). The pins are positioned at both positive and negative
library Y on both sides:

| Pin | Lib (x, y) | Expected world @ symbol (100,100) |
|---|---|---|
| 1  | (-10.16, +10.16) | (89.84, 89.84)  |
| 2  | (-10.16, +5.08)  | (89.84, 94.92)  |
| 3  | (-10.16, -5.08)  | (89.84, 105.08) |
| 4  | (-10.16, -10.16) | (89.84, 110.16) |
| 5  | (+10.16, +10.16) | (110.16, 89.84) |
| 6  | (+10.16, -10.16) | (110.16, 110.16) |

The test asserts every pin lands at its formula-predicted absolute
position, plus an explicit cross-check that lib +Y pins resolve to a
*smaller* world-Y (above the placement centre) — which is the specific
behaviour the original bug got wrong.

## Spot-check on symmetric passives

The two pre-existing symmetric-symbol tests (`Device:R` and rotated
`Device:C`) still pass alongside the new asymmetric one — meeting the
second ask in #135:

```
collected 3 items
tests/test_pin_locator_y_flip.py::test_stock_resistor_pin_y_matches_render_convention PASSED [ 33%]
tests/test_pin_locator_y_flip.py::test_rotated_capacitor_pin_x_matches_render_convention PASSED [ 66%]
tests/test_pin_locator_y_flip.py::test_asymmetric_multi_pin_symbol_y_flip PASSED [100%]
============================== 3 passed in 0.27s ==============================
```

Rebased on top of #179 (now in `main` at `ff6ae63`).

### DIFF
--- a/tests/test_pin_locator_y_flip.py
+++ b/tests/test_pin_locator_y_flip.py
@@ -111,3 +111,132 @@ def test_rotated_capacitor_pin_x_matches_render_convention():
         # World: (150-3.81, 100) = (146.19, 100). Verified vs kicad-cli netlist.
         assert p1[0] == pytest.approx(146.19), f"rotated pin 1 X wrong: {p1[0]}"
         assert p1[1] == pytest.approx(100.0), f"rotated pin 1 Y wrong: {p1[1]}"
+
+
+# ---------------------------------------------------------------------------
+# Asymmetric multi-pin regression (the case where the original bug hit hardest)
+# ---------------------------------------------------------------------------
+#
+# The symmetric-symbol tests above pass even with the pre-fix code because
+# Device:R and Device:C are pin-equivalent — swap pin 1 and pin 2 in the
+# netlist and the circuit still works. The bug only became visible on
+# asymmetric, multi-pin ICs like RF_Module:ESP32-WROOM-32, where pin 3 (EN)
+# and pin 35 (TXD0/IO1) are not interchangeable. With the original code,
+# pins at library y=+30.48 (top of the symbol) were reported at world
+# y=symbol_y+30.48 (bottom of the schematic), so labels meant for EN landed
+# on TXD0, and ERC reported either a dangling label or a wrong connection.
+#
+# Issue #135 specifically asked for an asymmetric-symbol regression. Rather
+# than depend on system libraries (RF_Module is not bundled with the repo),
+# this fixture builds a minimal 6-pin asymmetric symbol inline and asserts
+# every pin lands at the formula-predicted absolute position.
+
+_ASYMMETRIC_FIXTURE = """(kicad_sch
+    (version 20231120) (generator "test_pin_locator_y_flip")
+    (uuid "00000000-0000-0000-0000-000000000001")
+    (paper "A4")
+    (lib_symbols
+        (symbol "Test:Asym6"
+            (pin_numbers hide) (pin_names (offset 0.508))
+            (in_bom yes) (on_board yes)
+            (property "Reference" "U" (at 0 15.24 0)
+                (effects (font (size 1.27 1.27))))
+            (property "Value" "Asym6" (at 0 -15.24 0)
+                (effects (font (size 1.27 1.27))))
+            (property "Footprint" "" (at 0 0 0)
+                (effects (font (size 1.27 1.27)) hide))
+            (property "Datasheet" "" (at 0 0 0)
+                (effects (font (size 1.27 1.27)) hide))
+            (symbol "Asym6_0_1"
+                (rectangle (start -7.62 12.7) (end 7.62 -12.7)
+                    (stroke (width 0.254) (type default))
+                    (fill (type none))))
+            (symbol "Asym6_1_1"
+                (pin input line (at -10.16 10.16 0) (length 2.54)
+                    (name "TOP_LEFT" (effects (font (size 1.27 1.27))))
+                    (number "1" (effects (font (size 1.27 1.27)))))
+                (pin input line (at -10.16 5.08 0) (length 2.54)
+                    (name "MID_HIGH_L" (effects (font (size 1.27 1.27))))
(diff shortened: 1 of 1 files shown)


mixelpixx (approved):
Approving and merging.

Right diagnosis — the existing `test_pin_locator_y_flip.py` tests use `Device:R` and `Device:C`, which are pin-symmetric, so a pin 1 ↔ pin 2 swap is invisible to the netlist. The bug only surfaces on asymmetric multi-pin parts, which is exactly the gap your test fills.

Three things I like about the test design:

1. **Self-contained inline fixture** — building a 6-pin asymmetric symbol directly in the test rather than depending on `RF_Module:ESP32-WROOM-32` means it runs in any CI environment without KiCad's system libraries installed. That's the right choice.
2. **Pins on both positive and negative library Y on both sides** — a Y-flip bug or a left/right swap would visibly mis-place at least one pin, so the assertion is sensitive.
3. **Formula-predicted absolute positions** — the test pins down what the math should produce, not just "matches some other code path," so it'd catch a future regression even if both the implementation and the comparison helper drifted together.

Small note on the PR description: the underlying fix lives in `WireDragger.pin_world_xy` (`lx, ly = px, -py`), not specifically in #179 — the description says it landed there but #179 was the duplicate-pin-numbers fix. Doesn't matter for this PR's merge, just for the historical record.

Thanks for the prolific work today!

PR #189, opened 2026-05-19
feat: add check_courtyard_overlaps MCP tool

NiNjA-CodE:
## What

A new MCP tool, `check_courtyard_overlaps`, that detects courtyard
collisions between footprints and (optionally) flags courtyards that
extend past the board outline.

The key feature for AI-driven workflows is the **`positions`** parameter,
which accepts hypothetical placements `{ref: [x, y]}` or
`{ref: [x, y, rotation_degrees]}`. The tool evaluates the proposed
placement *without* writing to the board file — so an agent can validate
a `move_component` / `place_component` before committing it, instead of
the current loop of write → run DRC → parse violations → revert.

## Why

KiCad's DRC already reports `courtyards_overlap` violations, but only
*after* the move is committed to disk. AI agents (and humans) doing
iterative placement work end up writing, DRCing, parsing, reverting on
every attempt. With `check_courtyard_overlaps` they ask "is this
placement valid?" before committing — closing the feedback loop and
making automated placement tractable.

Concrete real-world numbers from validating against a 44-footprint
production board (TuneForge_TF001):

- Current placement: 1 known overlap correctly identified (SW1↔SW2,
  2.09×0.09mm — a tight squeeze from a recent relocation).
- Virtual placement test (`positions: {\"J2\": [15, 15]}`): the 6
  resulting overlaps with surrounding capacitors and U1 are all
  reported correctly. Without this tool the agent would have to commit
  the move and run DRC to learn the same thing.

## Implementation

- Uses real courtyard polygons (`fp.GetCourtyard(F_CrtYd)` /
  `B_CrtYd`) for accurate AABBs on custom and rotated footprints.
  Falls back to `fp.GetBoundingBox()` when no courtyard polygon is
  present.
- Virtual rotation rotates the four AABB corners and re-axis-aligns —
  conservatively (rotated-AABB ≥ rotated-polygon), so reports are never
  false-negatives.
- Optional `margin` parameter expands every courtyard by N mm.
- Optional `board_outline` parameter overrides Edge.Cuts for the
  boundary check.

## Attribution

The approach is ported from
[morningfire-pcb-automation](https://github.com/NiNjA-CodE/morningfire-pcb-automation),
specifically `scripts/placement/check_overlaps.py`. The upstream uses a
static per-footprint-type courtyard lookup table; this implementation
reads the real polygons from pcbnew so it works on any footprint
without maintaining a table. Attribution is in the docstring, the
TypeScript wrapper, the tool's MCP description (visible to clients),
and the CHANGELOG entry.

## Tests

`tests/test_check_courtyard_overlaps.py` — **12 cases, all passing**:

| Case | Verifies |
|---|---|
| `test_no_overlaps_when_components_are_spaced` | Happy path, no false positives |
| `test_overlap_detected_when_courtyards_intersect` | Detects 1mm-apart pair as overlap with correct extent |
| `test_margin_pushes_borderline_pairs_into_overlap` | `--margin` parameter |
| `test_refs_filter_restricts_to_subset` | `refs` parameter restricts pairs evaluated |
| `test_boundary_violation_flagged` | Past-edge flagged with correct `exceeds` distance |
| `test_include_boundary_false_suppresses_boundary_check` | Opt-out works |
| `test_virtual_position_does_not_mutate_footprint` | **Asserts `SetPosition` is never called** — virtual stays virtual |
| `test_virtual_rotation_swaps_aabb_extents` | 90° on a tall-narrow part exposes overlap, 0° doesn't |
| `test_no_board_loaded_returns_error_payload` | Clean error when board absent |
| `test_bad_position_spec_is_rejected_cleanly` | Wrong-arity position validated |
| `test_courtyard_fallback_to_bounding_box` | `GetCourtyard().OutlineCount()==0` → `GetBoundingBox()` |
| `test_board_outline_override_replaces_edge_cuts_bbox` | Override priority |

Tests use mocked pcbnew objects so they run under both the conftest
stub and real pcbnew installs.

```
collected 12 items
[all 12 PASS]
============================== 12 passed in 0.55s ==============================
```

## Files

| File | Change |
|---|---|
| `python/commands/component.py` | Implementation (`check_courtyard_overlaps` + 4 private helpers) |
| `python/kicad_interface.py` | Tool registration in the handler map |
| `python/schemas/tool_schemas.py` | MCP-protocol schema entry |
| `src/tools/component.ts` | TypeScript surface (builds clean with `npm run build`) |
| `tests/test_check_courtyard_overlaps.py` | 12 pytest cases |
| `CHANGELOG.md` | New entry under Unreleased → New MCP Tools |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,23 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### New MCP Tools
 
+- `check_courtyard_overlaps` — Detect courtyard overlaps between footprints
+  and (optionally) flag courtyards that extend past the board outline.
+  Returns overlap pairs with intersection extents (mm), per-component
+  boundary violations, and a placement summary. Accepts a `positions` dict
+  of hypothetical placements (with optional rotation) so an AI agent can
+  validate a proposed `move_component` / `place_component` before
+  committing it — closing the feedback loop that previously required
+  writing the move, running DRC, parsing violations, and reverting.
+
+  Approach ported from
+  [morningfire-pcb-automation](https://github.com/NiNjA-CodE/morningfire-pcb-automation)
+  (`scripts/placement/check_overlaps.py`). The original uses a static
+  per-footprint-type courtyard lookup table; this implementation reads
+  the real courtyard polygons (or pad bounding box fallback) from the
+  loaded board for accuracy on custom and rotated footprints, and adds
+  virtual placement + clearance margin support.
+
 - `query_zones` — Query copper zones (filled pours) on the board with optional
   filters by net, layer, or bounding box. Returns one entry per zone with its
   net, layers, priority, fill state, min thickness, bounding box, and filled
--- a/python/commands/component.py
+++ b/python/commands/component.py
@@ -1210,3 +1210,278 @@ class ComponentCommands:
                 module.SetPosition(pcbnew.VECTOR2I(pos.x, bottom - 2000000))  # 2mm offset from edge
         else:
             logger.warning(f"Unknown edge alignment: {edge}")
+
+    # -----------------------------------------------------------------------
+    # check_courtyard_overlaps
+    #
+    # Originally prototyped in morningfire-pcb-automation
+    #   https://github.com/NiNjA-CodE/morningfire-pcb-automation
+    #   (scripts/placement/check_overlaps.py — AABB lookup-table version)
+    #
+    # The version here uses the real courtyard polygons from the loaded
+    # board (more accurate than a static lookup), with virtual-placement
+    # support so an AI can validate a proposed move before committing it.
+    # -----------------------------------------------------------------------
+    def check_courtyard_overlaps(self, params: Dict[str, Any]) -> Dict[str, Any]:
+        """Detect courtyard overlaps between footprints (and board-edge violations).
+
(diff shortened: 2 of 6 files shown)


mixelpixx (approved):
Approving and merging.

This is exactly the right shape of tool for AI-driven placement workflows. The `positions` parameter that evaluates hypothetical placements without mutating the board is the key insight — it closes the write/DRC/parse/revert loop that makes automated placement painful today. `test_virtual_position_does_not_mutate_footprint` asserting `SetPosition` is never called pins that contract precisely.

Three implementation choices I like:

1. **Real courtyard polygons via `fp.GetCourtyard(F_CrtYd)/B_CrtYd`** with `GetBoundingBox()` fallback — more accurate than a per-footprint lookup table on custom or rotated footprints, and degrades gracefully.
2. **Conservative virtual-rotation AABB** (rotate corners, re-axis-align) — over-reports rather than miss, which is the right direction for a placement-validator.
3. **`include_boundary` and `board_outline` overrides** — gives callers control without forcing a rebuild for the common case.

Twelve focused tests pin the contract end-to-end. Attribution to your upstream `morningfire-pcb-automation` is clear in the docstring, schema, TS wrapper, and CHANGELOG — that's the right level of provenance disclosure.

Thanks!

PR #190, opened 2026-05-19
feat(autoroute): add best-of-N support (attempts, targetNets, passSchedule)

NiNjA-CodE:
## What

Three new optional parameters on `autoroute` that turn the existing
single-shot Freerouting call into a best-of-N attempt: run the router
several times with varied `--max-passes`, score each result by routing
completeness, and import the winning SES.

| Parameter | Default | Effect |
|---|---|---|
| `attempts` | `1` (unchanged) | Number of Freerouting runs. > 1 enables best-of-N. |
| `targetNets` | `[]` | Critical net names — routing all of them earns a 50,000-point scoring bonus. |
| `passSchedule` | `[50, 60, 65, 70, 75, 80, 85, 90, 55, 95]` | Per-attempt `--max-passes` values. Wraps if `attempts` exceeds length. |

Backward-compatible: omit `attempts` (or pass `1`) and the response and
behaviour are identical to today.

## Why

Single-shot Freerouting on dense boards regularly leaves 1–7 nets
unrouted. Re-running with varied `--max-passes` surfaces a better
solution most of the time; in my testing during the morningfire
TuneForge case study, 3 attempts reliably got to 0 unrouted on boards
where 1 attempt left 5–7.

Right now the workaround is bespoke per-project shell loops around the
existing `autoroute` tool. Folding best-of-N into the MCP closes that
gap so any AI agent can route a board cleanly without an external
driver.

## Scoring contract

Pinned by tests so future changes don't silently shift the ranking:

```
score = nets_routed * 1000 + segments
if targetNets and all routed: score += 50_000
```

- A single extra net (1000 pts) always beats any reasonable segment-count delta.
- Segments break ties when net counts are equal.
- The target-net bonus dominates marginal net-count gains from unrelated nets.

## Implementation notes

- Each scored attempt runs with `-mt 1` (single-thread optimisation).
  Freerouting 2.x's MT optimiser is documented to introduce clearance
  violations; forcing single-thread keeps the scoring comparison
  apples-to-apples.
- One failed attempt does **not** abort the whole best-of-N run — the
  failure is recorded in `response.attempts[]` and the remaining
  attempts compete. Only "every attempt failed" surfaces as an error.
- The winning SES is preserved as `<stem>_best.ses` alongside the
  canonical `<stem>.ses` so the caller can inspect it.
- Response shape changes minimally:
  - `attempts <= 1`: identical to today (no `attempts` / `best_attempt`).
  - `attempts > 1`: adds `attempts: [...]`, `best_attempt`, `best_score`, `best_ses_path`.

## Attribution

Scoring approach and the default pass schedule are ported from
[morningfire-pcb-automation](https://github.com/NiNjA-CodE/morningfire-pcb-automation)
(`scripts/routing/freeroute_runner.py::score_ses` + run loop). Credited
in the function docstring, the TypeScript wrapper, the tool's MCP
description (visible to clients), and the CHANGELOG entry.

The MCP version adds: cleaner per-attempt result reporting, the
automatic `-mt 1` switch during scored runs, graceful degradation on
partial failure, and validation of `attempts` (rejects `0` or non-numeric
values before any subprocess work).

## Tests

| File | Cases | What it locks in |
|---|---|---|
| `tests/test_autoroute_score.py` | 8 | The scoring formula (net dominance, seg tiebreak, target bonus, quoted-name normalisation, sorted output, vias don't count) |
| `tests/test_autoroute_best_of_n.py` | 6 | Orchestration: single-attempt response shape unchanged, best-of-3 picks correctly, one failed attempt continues, passSchedule wraps, targetNets bonus wins, invalid attempts rejected |

```
collected 14 items
[all 14 PASS]
============================== 14 passed in 0.25s ==============================
```

Tests are pure-Python — `subprocess` is mocked so the suite runs in any
environment (no Java / Freerouting / KiCad required).

TypeScript builds clean (`npm run build`).

## Files

| File | Change |
|---|---|
| `python/commands/freerouting.py` | `_score_ses` helper, refactored `autoroute`, `-mt 1` plumbing |
| `src/tools/freerouting.ts` | New `attempts` / `targetNets` / `passSchedule` parameters with descriptions |
| `tests/test_autoroute_score.py` | 8 scoring-contract tests |
| `tests/test_autoroute_best_of_n.py` | 6 orchestration tests |
| `CHANGELOG.md` | Unreleased → Tool Enhancements entry |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,28 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Tool Enhancements
 
+- `autoroute`: best-of-N support. New optional parameters `attempts`,
+  `targetNets`, and `passSchedule`. When `attempts > 1`, Freerouting is
+  invoked multiple times with varied `--max-passes` values, each result
+  is scored by `(nets_routed * 1000) + segments` plus a 50,000-point
+  bonus when every `targetNets` entry is routed, and the winning SES is
+  imported into the board. Single-attempt behaviour is unchanged when
+  `attempts` is omitted, so existing callers don't need updates.
+
+  Motivation: on dense boards a single Freerouting run routinely leaves
+  1–7 nets unrouted. Cycling through a few `-mp` values typically drives
+  the unrouted count to zero. Empirically, 3 attempts is usually enough
+  for 4-layer designs; 5–8 for stubborn cases.
+
+  The scoring approach and the default `passSchedule` are ported from
+  [morningfire-pcb-automation](https://github.com/NiNjA-CodE/morningfire-pcb-automation)
+  (`scripts/routing/freeroute_runner.py`). The MCP version adds:
+  cleaner per-attempt result reporting, automatic single-thread
+  optimisation (`-mt 1`) during scored attempts so the multi-threaded
+  optimiser's known clearance-violation bug doesn't distort the
+  comparison, and graceful degradation when one attempt errors out
+  (the run continues and the best of the remainder wins).
+
 - `edit_schematic_component`: extended with two new optional parameters that
   promote arbitrary custom properties to first-class citizens:
   - **`properties`** — map of property name to either a string value or a full
--- a/python/commands/freerouting.py
+++ b/python/commands/freerouting.py
@@ -11,11 +11,12 @@ Supports two execution modes:
 
 import logging
 import os
+import re
 import shutil
 import subprocess
 import time
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Iterable, List, Optional
 
 logger = logging.getLogger("kicad_interface")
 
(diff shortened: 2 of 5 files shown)


mixelpixx (approved):
Approving and merging.

Backward-compatible by construction (`attempts <= 1` keeps the existing response shape and behaviour) and the scoring formula is explicit and testable:

```
score = nets_routed * 1000 + segments
if targetNets and all routed: score += 50_000
```

The 1000× weighting on routed nets means a single extra net always beats any reasonable segment-count delta — that's the right priority order, and `test_autoroute_score.py` pins it in eight cases so future changes can't quietly shift the ranking.

Other implementation details that are the right calls:

1. **`-mt 1` forced on scored attempts** to neutralise Freerouting 2.x's MT optimiser's documented clearance-violation issue. Keeps the apples-to-apples comparison honest.
2. **One failed attempt doesn't abort the whole best-of-N run** — recorded in `attempts[]`, remaining attempts still compete. Only "every attempt failed" surfaces as an error.
3. **Winning SES preserved as `<stem>_best.ses`** so the caller can inspect.
4. **Tests mock subprocess** — the suite runs in any CI environment without Java / Freerouting / KiCad.

Real-world value statement (3 attempts reliably routing what 1 leaves with 5–7 unrouted nets) plus the attribution to `morningfire-pcb-automation` make this a clean contribution. Thanks!

PR #191, opened 2026-05-19
feat: add add_gnd_stitching_vias MCP tool with all-layer collision

NiNjA-CodE:
## What

A new MCP tool, `add_gnd_stitching_vias`, that drops GND stitching vias
across the board with collision checking against every non-GND segment,
via, and pad **on every copper layer**. PTH vias penetrate the full
stackup, so an F.Cu-only check (the most common shortcut taken by
ad-hoc tools) silently creates shorts on inner / B.Cu copper. This
implementation explicitly walks all layers.

## Why

GND stitching is one of the most common post-layout chores on every
real PCB — bridge fragmented zones, densify ground under noisy ICs,
satisfy stitching rules for shielded designs. Right now AI agents and
human users alike either route them by hand or ship bespoke scripts.
Both fail in the same way: the "are there obstacles here?" check
misses tracks on inner / B.Cu and produces silent shorts.

This bakes the right primitive into the MCP, with all-layer collision
as a first-class invariant and three composable placement strategies.

## Strategies (freely composable)

| Strategy | When to use |
|---|---|
| `grid` | First-pass stitching across the board interior |
| `around_refs` | Densify ground under specific footprints — MCUs, switching regulators, RF parts |
| `in_zones` | Only place vias where they actually land on a GND copper zone, so each one stitches real polygons rather than floating on silkscreen. Recommended on boards where the GND zone is fragmented. |

All three strategies share the same collision check + intra-call
clump prevention, so passing `["grid", "around_refs", "in_zones"]` is
a safe kitchen-sink configuration.

## Other parameters

- `gndNet` — explicit net, or auto-detect (`GND` → `GROUND` → `VSS` → `/GND`)
- `viaSize` / `viaDrill` / `clearance` / `edgeMargin`
- `spacing` — grid step
- `densifyRefs` / `densifyRadius` — for `around_refs`
- `maxVias` — cap on total placements; useful when iterating
- `dryRun` — return placements without modifying the board, for preview

Validates that `viaDrill < viaSize`, rejects unknown strategy names,
surfaces clear errors when the GND net can't be resolved or the
Edge.Cuts outline is missing.

## Attribution

Approach ported from
[morningfire-pcb-automation](https://github.com/NiNjA-CodE/morningfire-pcb-automation)
(`scripts/ground/add_gnd_vias.py`). The original parses the PCB text
with regex and writes new vias by string concatenation; this port:

- Reads obstacles via the pcbnew API — handles rotated footprints
  correctly, picks up net codes from the loaded board.
- Integrates with the live in-memory board — two sequential calls
  see each other's placements (e.g. "densify around U1" then
  "across the rest of the board" composes naturally).
- Adds the `in_zones` strategy (uses `Zone.HitTestFilledArea` to
  guarantee placements land on actual GND copper).
- Adds `maxVias` and `dryRun`.

Credit is in the function docstring, the TypeScript wrapper, the MCP
tool description (visible to clients), and the CHANGELOG entry.

## Tests

`tests/test_add_gnd_stitching_vias.py` — **18 cases, all passing**:

- `grid` strategy fills empty board with correct count
- Collision blocks via near a signal track (with extent assertion)
- GND-net obstacles are correctly ignored
- `around_refs` densifies near footprints with bounded extent
- `in_zones` rejects candidates outside `HitTestFilledArea`
- `dryRun` does NOT call `board.Add`
- Actual run calls `board.Add` per placement
- `maxVias` caps total placements
- Intra-call clump prevention (asserts pairwise distance)
- `viaDrill >= viaSize` is rejected
- Unknown strategy name is rejected
- Missing GND net returns clear error payload
- No board loaded returns clear error
- Named GND net (e.g. VSS) is honoured even when GND also exists
- 4 direct unit tests for the `_point_to_segment_distance_nm` helper

```
collected 18 items
[all 18 PASS]
============================== 18 passed in 0.19s ==============================
```

Tests use mocked pcbnew objects so they run in any environment.

## Real-board smoke test

Validated on TuneForge_TF001 (4-layer, 44 footprints, fully routed):

| Run | Result |
|---|---|
| dry-run `grid` (4mm spacing) | 141 vias placed, 129 candidates blocked by collision |
| dry-run `grid` + `in_zones` (4mm) | 140 placed, 15 rejected by zone membership, 115 by collision |

GND net auto-detected as expected; no false placements on signal traces.

## Files

| File | Change |
|---|---|
| `python/commands/routing.py` | Implementation (`add_gnd_stitching_vias` + helpers, ~370 LOC) |
| `python/kicad_interface.py` | Handler registration |
| `python/schemas/tool_schemas.py` | MCP-protocol schema |
| `src/tools/routing.ts` | TypeScript surface (builds clean) |
| `tests/test_add_gnd_stitching_vias.py` | 18 pytest cases |
| `CHANGELOG.md` | New entry under Unreleased → New MCP Tools |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,42 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### New MCP Tools
 
+- `add_gnd_stitching_vias` — Drop GND stitching vias across the board with
+  collision checking against every non-GND segment, via, and pad on every
+  copper layer. PTH vias penetrate the full stackup, so an F.Cu-only check
+  (the most common shortcut) silently creates shorts on inner / B.Cu
+  copper — this implementation explicitly walks all layers.
+
+  Combines three placement strategies, freely composable:
+
+  - `grid` — regular grid across the board interior.
+  - `around_refs` — densify around named footprints (good for tucking
+    extra ground under MCUs, switching regulators, or RF parts).
+  - `in_zones` — restrict candidates to points inside the filled
+    polygons of GND copper zones, so each new via actually stitches
+    real ground polygons together rather than floating on silkscreen.
+
+  Also supports per-via geometry control (`viaSize`, `viaDrill`,
+  `clearance`, `edgeMargin`), an `maxVias` cap for incremental work,
+  auto-detection of the GND net (tries `GND` / `GROUND` / `VSS` /
+  `/GND`), and a `dryRun` mode that returns the placements that
+  *would* be made without modifying the board — useful for previewing
+  before committing.
+
+  Returns `{ placed: [{x, y, unit}, ...], summary: {placed_count,
+  candidates_evaluated, skipped_by_zone_membership,
+  skipped_by_collision, ...} }`.
+
+  Approach ported from
+  [morningfire-pcb-automation](https://github.com/NiNjA-CodE/morningfire-pcb-automation)
+  (`scripts/ground/add_gnd_vias.py`). The original parses the PCB
+  text with regex and writes new vias by string concatenation; this
+  port reads obstacles via the pcbnew API so it handles rotated
+  footprints correctly, integrates with the in-memory board (two
+  sequential calls see each other's placements), picks up net codes
+  from the live board, and adds the `in_zones` strategy + the
+  `maxVias` cap + dry-run.
+
 - `check_courtyard_overlaps` — Detect courtyard overlaps between footprints
   and (optionally) flag courtyards that extend past the board outline.
   Returns overlap pairs with intersection extents (mm), per-component
--- a/python/commands/routing.py
+++ b/python/commands/routing.py
@@ -1649,3 +1649,391 @@ class RoutingCommands:
         dx = p1.x - p2.x
         dy = p1.y - p2.y
         return (dx * dx + dy * dy) ** 0.5
+
+    # -----------------------------------------------------------------------
+    # add_gnd_stitching_vias
+    #
+    # Originally prototyped in morningfire-pcb-automation:
+    #   https://github.com/NiNjA-CodE/morningfire-pcb-automation
+    #   (scripts/ground/add_gnd_vias.py — regex-on-PCB-text version)
+    #
+    # The version here uses the pcbnew API so it handles arbitrary
+    # rotations, gets net IDs / clearances from the loaded board, and
+    # works against the live in-memory board state (so two calls in
+    # sequence — e.g. "around U1" then "across the board" — both see
+    # the first call's placements). All copper layers are checked
+    # because a through-hole via penetrates the full stackup; missing a
+    # B.Cu collision check is the classic way GND-stitching tools
+    # create silent shorts.
+    # -----------------------------------------------------------------------
+    def add_gnd_stitching_vias(self, params: Dict[str, Any]) -> Dict[str, Any]:
+        """Drop GND stitching vias across the board, collision-checked on every copper layer.
+
+        Strategies (combine freely):
+          - ``grid``        Place candidates on a regular grid across the board
+                            interior. Each candidate is accepted only if its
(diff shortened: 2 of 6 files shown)


mixelpixx (approved):
Approving — really nice piece of work.

The implementation pattern is exactly right: all-layer obstacle gathering (`board.GetTracks()` + `board.GetFootprints()` walks every layer because PTH vias do), intra-call clump prevention via `placed_via_centres`, two-variant zone-membership test (`HitTestFilledArea` with a bbox fallback for API variants), and a dry-run mode for previewing before commit. The three composable strategies (`grid` / `around_refs` / `in_zones`) all share the same collision check + dedupe, so combining them in one call is safe and gives the LLM a flexible toolkit.

Test coverage (16 tests) is genuinely thorough — happy paths, collision-blocks-via-near-signal, GND-self-not-an-obstacle, around-refs densification, in-zones rejection, dry-run, actual-run, maxVias cap, intra-call clump prevention, geometry validation, error paths, named-GND override, and the segment-distance helper directly. The `pcbnew` mock builders (`_track`, `_via`, `_pad`, `_footprint`, `_board`) are a clean reusable pattern.

Two things I want to ask for before merging:

1. **Rebase needed.** GitHub reports CONFLICTING — my fault, not yours: yesterday's merges (#174 `query_zones`, this morning's #163 bbox/courtyard) both touched `python/commands/routing.py`, `python/kicad_interface.py`, `python/schemas/tool_schemas.py`, and `src/tools/routing.ts` in the same regions you're adding to. The semantic merge should be straightforward — your new tool slots in alongside `query_zones` in each file's tool list.

2. **Arc handling — flag for follow-up, not blocking.** We merged `route_arc_trace` yesterday (#165), so `PCB_ARC` items can now appear in `board.GetTracks()`. The current code does:

   ```python
   is_via = track.GetClass() == "PCB_VIA"
   if is_via: ...
   else: ... obstacle_tracks.append((s.x, s.y, e.x, e.y, ...))
   ```

   A `PCB_ARC` falls into the `else` branch and gets treated as a straight chord. The `_point_to_segment_distance_nm` collision check then under-estimates the arc's effective region — a via could land inside the arc's bulge and short it. Probably rare in practice (arc tracks are still uncommon), but worth either:
   - **Cheap option**: detect `GetClass() == "PCB_ARC"` and inflate the segment's effective half-width by the chord-sagitta to a conservative bound, or
   - **Proper option**: pcbnew exposes `PCB_ARC::GetRadius()` / `GetCenter()` / `GetMid()` — distance from a point to an arc is solvable.

   Either way: not blocking this PR. Could you file a follow-up issue once this lands so it doesn't get lost?

Once rebased, this is ready to go in. Thanks for the work — and thanks especially for the morningfire-pcb-automation context in the CHANGELOG, it makes the design decisions easy to trace.

NiNjA-CodE:
Thanks for the careful review — both items handled:

**Rebase done.** Rebased onto current `main` (now includes #163, #187,
#188, #189, and #190). The only conflict was in `CHANGELOG.md` where
my entry and #189's were both inserted under the same anchor; resolved
additively (both entries kept, `add_gnd_stitching_vias` immediately
above `check_courtyard_overlaps` since the latter merged first
chronologically). `kicad_interface.py`, `tool_schemas.py`, and
`routing.ts` all auto-merged cleanly — the new tool slots in next to
`query_zones` in each file's tool list exactly as you predicted.

Post-rebase sanity checks:

- All 18 tests in `tests/test_add_gnd_stitching_vias.py` still pass.
- `npm run build` clean.

**PCB_ARC follow-up filed as #192** — covers both the cheap
(chord-sagitta inflation) and proper (point-to-arc distance) fix
options with a reproduction sketch. Linked back to this PR so it
doesn't get lost. Happy to take a swing at the proper fix once this
merges.

Force-pushed the rebased branch to `feat/add-gnd-stitching-vias`;
should be a clean merge now.

PR #193, opened 2026-05-20
Fix/kicad10 symdir support

karu2003:

PR #194, opened 2026-05-20
 KiCad 10 compatibility & interactive schematic editing

karu2003:
KiCad 10 compatibility & interactive schematic editing

This branch extends the upstream MCP server with full KiCad 10 support and interactive schematic editing capabilities on Windows.

Key additions over main:

Interactive editing — schematic updates appear in KiCad instantly; background dialog watcher auto-dismisses reload confirmations without user intervention
Component mirroring — mirrorY parameter in add_schematic_component for horizontal symbol flipping (e.g. symmetric transistor layouts)
KiCad 10 library support — .kicad_symdir/ per-symbol directory format, sym-lib-table with quoted URIs and spaces, ${KICAD_3RD_PARTY} resolution
Wire routing fixes — hierarchical sub-sheet support, correct pin snapping, power-flag isolation
Component placement fixes — correct RefDes/Value orientation and justification from library properties
PCB tools — GND stitching vias, arc traces, courtyard overlap check, copper zone queries, best-of-N autorouting
IPC improvements — runtime reconnect, absolute angle for rotate, inch→mm unit conversion
Windows fixes — Cairo DLL preload for 2D board view, auto-reload after component add

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **KiCad 10 standard symbol libraries are now discoverable**: `search_symbols`,
+  `list_symbol_libraries`, and `add_schematic_component` previously returned no
+  results for stdlib symbols (`Device:R`, `RF_Module:ESP32-WROOM-DA`,
+  `Connector_Generic:*`, etc.) on KiCad 10 installations. Two underlying
+  causes:
+  1. `_find_kicad_symbol_dir` (in `library_symbol.py` and
+     `dynamic_symbol_loader.py`) hard-coded only KiCad 8.0 / 9.0 install
+     paths, so `${KICAD10_SYMBOL_DIR}` and the matching `KICAD10_SYMBOL_DIR`
+     env var never resolved. Added `C:/Program Files/KiCad/10.0/share/kicad/symbols`
+     and `KICAD10_SYMBOL_DIR` env-var support.
+  2. The `sym-lib-table` regex used `[^")\s]+` for the URI capture, which
+     rejects any URI containing a space. KiCad's default global table on
+     Windows references the stdlib via
+     `(lib (name "KiCad") (type "Table") (uri "C:/Program Files/KiCad/10.0/share/kicad/template/sym-lib-table"))`
+     — the space in `Program Files` caused the Table reference to be
+     skipped silently, hiding the entire stdlib. Regex now matches either
+     a quoted value (allowing spaces) or a bare token.
+
 - **`rotate_component` now treats `angle` as an absolute target rotation**,
   matching its schema description. Previously the IPC backend added the
   supplied angle to the current rotation, so two consecutive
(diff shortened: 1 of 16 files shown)


karu2003:
https://www.youtube.com/watch?v=o0wmptV-JBk  :)

mixelpixx:
Thanks for putting this up, @karu2003 — there's real, valuable work in here, especially the interactive-editing piece. I want to be upfront, though: I can't merge it as a single 16-file, +1568/−237 PR, for two reasons.

## 1. The repo moved a lot this week, and several pieces here already landed

Over the last few days a number of focused PRs from other contributors merged that overlap with this branch:

- `${KICAD_3RD_PARTY}` resolution → #187 (gsdali), merged
- Hierarchical sub-sheet wire support → #170 (inktomi), merged
- Power-flag isolation → #177 (blinksoft), merged
- `sym-lib-table` quoted-URI parsing → #164 (gcolonese), merged
- `.kicad_symdir` support → also in your own #193, and overlapping #148 (Kulitorum)

So a chunk of this diff is re-implementing things that are already on `main`. A rebase will shrink it substantially and make the genuinely-new parts visible.

Also a direct collision: this PR adds `tests/test_backend_metadata.py`, but #140 (scorp508, already approved and awaiting its own rebase) creates that same file. Two different versions of the same test file can't both land.

## 2. The novel pieces deserve to be reviewed — and merged — separately

There are at least three distinct features bundled here, with very different review needs:

**a) Interactive schematic editing + the KiCad dialog watcher.** This is the headline and it's genuinely useful, but it's also the piece that needs the most careful discussion, because `_start_kicad_dialog_watcher` is a different *kind* of thing than the rest of the codebase:

- It runs a daemon thread for the whole server lifetime that polls `win32gui.EnumWindows` and `PostMessage(child, BM_CLICK)` to auto-click buttons in KiCad's GUI.
- It's Windows-only via `pywin32` (soft-gated — good, it disables cleanly when pywin32 is absent).
- The dialog match is title-keyword + button-text based, and the button text is hardcoded to `Yes`/`Ja` — that breaks on any non-English/German KiCad locale.
- Auto-dismissing dialogs the user didn't initiate carries a real data-loss risk: if the watcher ever clicks "Yes" on a destructive confirmation (revert, overwrite, discard), there's no undo. I'd want the dialog-matching to be very tightly scoped, and ideally opt-in via an env var that defaults off.

This one I'd like as its own PR with its own thread so we can work through the locale handling, the safety scoping, and whether pywin32 should be a documented optional dependency.

**b) `mirrorY` component mirroring.** Small and clean-sounding, and it also happens to address #72 (the "missing FLIP function" request). Worth its own small PR — it'll merge fast.

**c) KiCad 10 symdir support.** Same content as your #193 — let's consolidate it there (see my note on #193 about coordinating with #148) rather than carrying it in two PRs.

## What I'd ask

Rebase onto current `main`, then split into focused PRs:

1. Interactive editing + dialog watcher (its own PR + design thread)
2. `mirrorY` mirroring (small, references #72)
3. symdir → fold into #193

Drop `tests/test_backend_metadata.py` from this branch (it belongs to #140).

I know splitting an already-working fork branch is annoying, and I appreciate that this all works together on your setup. But focused PRs are how everything else this week got reviewed and merged quickly, and they'll get *your* work in faster too. Happy to help triage which hunks go where if the rebase is gnarly.

karu2003:
Closing in favor of focused PRs per review feedback: #193 (symdir), #207 (mirrorY, closes #72), #208 (opt-in interactive reload, KICAD_INTERACTIVE_SCHEMATIC default off).

PR #201, opened 2026-05-23
Fix MCP error result for DRC failure payloads

jbjardine:
## Summary
- Add a shared formatter for KiCad command results that marks explicit `success: false` payloads as MCP `isError` results.
- Route the design-rule/DRC TypeScript tools through that formatter so failures such as `No board is loaded` are not reported as OK tool calls.
- Add a static regression test covering the design-rule wrappers and strict `success === false` behavior.

## Test Plan
- `npm run build`
- `npm run lint:ts` (passes with existing warnings outside this change)
- `pytest -o addopts= tests/test_ts_tool_registry.py tests/test_mcp_error_wrapping_static.py -v`
- `node -e "import('./dist/tools/tool-response.js').then(m => { const failed = m.formatKicadResult({ success: false, message: 'No board is loaded' }); const ok = m.formatKicadResult({ success: true }); if (failed.isError !== true) throw new Error('failed payload missing isError'); if ('isError' in ok) throw new Error('success payload marked isError'); console.log(JSON.stringify({ failed, ok })); })"`

Note: `pytest -o addopts= tests -q` does not collect fully in my local Windows environment because optional Python dependencies such as `PIL` and `sexpdata` are not installed there.

### DIFF
--- a/src/tools/design-rules.ts
+++ b/src/tools/design-rules.ts
@@ -7,6 +7,7 @@
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { z } from "zod";
 import { logger } from "../logger.js";
+import { formatKicadResult } from "./tool-response.js";
 
 // Command function type for KiCAD script calls
 type CommandFunction = (command: string, params: Record<string, unknown>) => Promise<any>;
@@ -52,14 +53,7 @@ export function registerDesignRuleTools(server: McpServer, callKicadScript: Comm
       logger.debug("Setting design rules");
       const result = await callKicadScript("set_design_rules", params);
 
-      return {
-        content: [
-          {
-            type: "text",
-            text: JSON.stringify(result),
-          },
-        ],
-      };
+      return formatKicadResult(result);
     },
   );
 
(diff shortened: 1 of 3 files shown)


jbjardine:
Live before/after repro evidence from Windows + KiCad bundled Python 9.0.7:

Before this fix (`upstream/main` / `ece116d`), `run_drc` with no board loaded returns a failed KiCad payload but the MCP result is not marked as an error:

```json
{
  "result": {
    "content": [
      {
        "type": "text",
        "text": "{\"success\":false,\"message\":\"No board is loaded\",\"errorDetails\":\"Load or create a board first\",\"_backend\":\"swig\",\"_realtime\":false}"
      }
    ]
  }
}
```

After this fix, the same `run_drc` call preserves the same payload and adds `isError: true`:

```json
{
  "result": {
    "content": [
      {
        "type": "text",
        "text": "{\"success\":false,\"message\":\"No board is loaded\",\"errorDetails\":\"Load or create a board first\",\"_backend\":\"swig\",\"_realtime\":false}"
      }
    ],
    "isError": true
  }
}
```

The server logs show the actual MCP server booting through stdio and using pcbnew 9.0.7 in both runs.

PR #202, opened 2026-05-23
Add backend state MCP tool

jbjardine:
## Summary
- Add a `get_backend_state` MCP tool exposing backend/realtime status plus loaded project/board paths.
- Track the last auto-save status and compare board disk signatures so callers can see clean, dirty, and unknown board state.
- Add Python/TypeScript regression coverage for backend state visibility and tool registration.

## Live proof
Before this PR (`upstream/main` / `ece116d`), `get_backend_state` is not listed and calling it returns `Tool get_backend_state not found`; `get_project_info` without a board only reports `No board is loaded` plus `_backend/_realtime`, with no `loadedProject`, `loadedBoard`, or `dirty`.

After this PR, the real MCP stdio server reports:
- no board: `loadedProject:false`, `loadedBoard:false`, `dirty:false`
- after `create_project`: `loadedProject:true`, `loadedBoard:true`, `dirty:false`
- after external board-file edit in the same session: `dirty:true`, `diskChangedExternally:true`

## Test Plan
- `npm run build`
- `npm run lint:ts` (passes with 9 existing warnings outside this change)
- `python -m pytest -o addopts= tests/test_backend_metadata.py tests/test_ts_tool_registry.py -v`
- `python -m black --check --workers 1 python/kicad_interface.py tests/test_backend_metadata.py`
- `git diff --check`
- Live before/after MCP stdio proof using KiCad bundled Python 9.0.7

Note: running the same pytest target with repository coverage addopts on this local Windows environment executes all 20 tests successfully, then crashes in pytest-cov coverage report generation.

PR #203, opened 2026-05-24
fix: parse quoted KiCad library table URIs with spaces

jbjardine:
## Summary
- Fix fp-lib-table and sym-lib-table parsing so quoted `name`, `type`, and `uri` values are handled correctly.
- Preserve quoted URIs containing spaces, such as OneDrive/company library paths, instead of truncating at the first space.
- Add regression coverage for both footprint and symbol library managers.

## Test Plan
- `python -m pytest -o addopts= tests/test_kicad_3rd_party_env_resolution.py -q`
- `python -m pytest -o addopts= tests/test_kicad_3rd_party_env_resolution.py tests/test_loader_global_sym_lib_table.py -q`
- `python -m pytest -o addopts= tests/test_symbol_library.py tests/test_kicad_3rd_party_env_resolution.py tests/test_loader_global_sym_lib_table.py -q`
- `python -m black --check --workers 1 python/commands/library.py python/commands/library_symbol.py tests/test_kicad_3rd_party_env_resolution.py`
- Docker clean proof from `git archive HEAD`: `python -m pytest -o addopts= tests/test_kicad_3rd_party_env_resolution.py tests/test_loader_global_sym_lib_table.py -q`

## Notes
- Independent of PR #201 and PR #202: this touches only library-table parsing and related tests.
- Does not add new MCP tools or change public tool schemas.

### DIFF
--- a/python/commands/library.py
+++ b/python/commands/library.py
@@ -91,14 +91,19 @@ class LibraryManager:
             with open(table_path, "r") as f:
                 content = f.read()
 
-            # Simple regex-based parser for lib entries
-            # Pattern: (lib (name "NAME")(type TYPE)(uri "URI")...)
-            lib_pattern = r'\(lib\s+\(name\s+"?([^")\s]+)"?\)\s*\(type\s+"?([^")\s]+)"?\)\s*\(uri\s+"?([^")\s]+)"?'
+            # Simple regex-based parser for lib entries. Name, type, and URI
+            # may be quoted or bare; quoted URIs can contain spaces.
+            lib_pattern = (
+                r"\(lib\s+"
+                r'\(name\s+(?:"([^"]+)"|([^"\)\s]+))\)\s*'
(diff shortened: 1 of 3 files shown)


mixelpixx:
Reviewed and verified — merging. The old `(name "?...")` pattern used `[^")\s]` which truncates a quoted URI at the first space; the new alternation `(?:"([^"]+)"|([^"\)\s]+))` correctly captures quoted-with-spaces (OneDrive/company paths) or bare tokens for name/type/uri. Confirmed it doesn't regress `${KICAD_3RD_PARTY}` resolution (that happens after parsing; the env-resolution tests pass), and it's applied symmetrically to both the footprint (`library.py`) and symbol (`library_symbol.py`) table parsers.

Verification (Windows): `tsc` builds clean, the 5 env-resolution tests pass, full suite at the known baseline. Thanks for the tests.

PR #204, opened 2026-05-24
fix(jlcpcb): download prebuilt catalog instead of broken JLCSearch offset loop (#199)

### LINKED ISSUE #199: download_jlcpcb_database broken: JLCSearch API ignores offset parameter (returns same 100 parts for every page)
## Environment
- KiCAD-MCP-Server v2.1.0-alpha (commit `9c672d4` on main)
- KiCad 10.0.1 (macOS, Apple Silicon)
- JLCSearch public API at `jlcsearch.tscircuit.com`

## Problem
`download_jlcpcb_database` downloads the same 100 parts 8,000+ times. Final DB contains only 100 unique records instead of the full catalog. The tool reports "Downloaded 833500 parts" but all are duplicates.

## Root Cause
The JLCSearch API does **not** support pagination via the `offset` parameter. Every call to `/components/list.json?offset=N` returns the identical 100 results.

Verified:
```python
offset=0    → [14663, 1525, 1591, 25804, 60474, ...]
offset=100  → [14663, 1525, 1591, 25804, 60474, ...]  # identical
offset=200  → [14663, 1525, 1591, 25804, 60474, ...]  # identical
```

`download_all_components()` in `python/commands/jlcsearch.py` iterates with `offset += len(batch)` but every page returns the same results. `all_parts` grows to 833,500 entries (all duplicates), and `INSERT OR REPLACE` collapses them to 100 unique rows.

## Impact
- `download_jlcpcb_database` effectively useless beyond 100 parts
- User sees "833500 parts downloaded" progress but only 100 are unique
- `search_jlcpcb_parts` returns very limited results

## Suggested Fix
1. Remove or disable `download_jlcpcb_database` until a paginating source is available (document as "popular parts only, 100 max")
2. Switch download to the official JLCPCB API (`python/commands/jlcpcb.py`) which does support pagination (requires API credentials)
3. Find an alternative public API that supports real offset-based pagination

## Note
The `import_jlcsearch_parts` field mapping is correct — the insert logic works fine. The issue is entirely on the download side: the API source does not paginate.

mixelpixx:
Fixes #199.

## Problem
`download_jlcpcb_database` paged the community JLCSearch endpoint (`jlcsearch.tscircuit.com`) with an `offset` param. That endpoint is a search front-end (built on yaqwsx/jlcparts), not a bulk-export API — it ignores `offset` and returns the same first 100 parts on every page, so a full catalog download never actually paginated.

## Approach — layered prebuilt download, no scraping
The whole ecosystem downstreams from yaqwsx/jlcparts, so reuse a prebuilt catalog rather than reinvent a scraper:

1. **CDFER single-file SQLite** (primary) — uncompressed `.sqlite3` on GitHub Pages, no 7z/zip needed (reliable on Windows, the main pain point).
2. **yaqwsx/jlcparts split 7z** (fallback) — only if CDFER fails and a 7z CLI is present.
3. **Official JLCPCB API** (optional) — cursor pagination, used only if `JLCPCB_APP_ID/API_KEY/API_SECRET` are set.

New `python/commands/jlcpcb_downloader.py` owns this. Conversion reads CDFER's `v_components` view (or sniffs the largest table for yaqwsx), C-prefixes integer LCSC, derives `library_type` from basic/preferred, maps `mfr`→`mfr_part`, and normalizes price JSON to the manager's `[{qty,price}]` shape. The tool reports the `source`, the catalog's `Last-Modified`, and — per review — a **stale warning** when the catalog is older than 14 days (CDFER's upstream pipeline has stalled), pointing users to the fresh sources.

`_handle_download_jlcpcb_database` now calls the module (closing/reopening the manager connection so the DB file can be rewritten on Windows). The broken offset loop is removed from `jlcsearch.py` (the client stays for interactive lookups). `download_jlcpcb.py` is now a thin CLI wrapper, and the TS tool schema gains an optional `source` param.

## Testing
- End-to-end against live CDFER on Windows: ~616k parts downloaded + converted in ~40s; `search_jlcpcb_parts` / `get_jlcpcb_part` work unchanged.
- 5 new unit tests (no network): conversion, price normalization, source fall-through, stale-vs-fresh flagging.
- Full suite: no new failures vs baseline; black + flake8 clean; `tsc` builds.

## Notes
- CDFER's published DB is currently ~stale (its scraper pipeline has been failing since ~Apr 2); content is valid but frozen, hence the freshness warning + yaqwsx/official fallbacks. Thanks to @bhoot1234567890 for independently confirming the landscape and the CDFER pipeline breakage.

## Test plan
- [ ] macOS: CDFER primary path
- [ ] macOS: yaqwsx fallback (7z present)
- [ ] Windows: CDFER primary (done)

### DIFF
--- a/download_jlcpcb.py
+++ b/download_jlcpcb.py
@@ -1,314 +1,69 @@
 #!/usr/bin/env python3
 """
-Download JLCPCB parts database from yaqwsx/jlcparts pre-built cache.
+CLI to download the JLCPCB parts database into the MCP server's local DB.
 
-This downloads the full JLCPCB catalog (~421MB compressed, ~1.5GB SQLite)
-from GitHub Pages in ~5 minutes instead of the broken JLCSearch API approach.
+This is a thin wrapper around ``commands.jlcpcb_downloader`` (the same code the
+``download_jlcpcb_database`` MCP tool uses). It downloads a prebuilt catalog and
+converts it into ``jlcpcb_parts.db``:
 
-The cache.sqlite3 file contains all JLCPCB parts with stock, pricing,
-and category data. We then convert it into the format expected by the
-KiCad MCP server's JLCPCBPartsManager.
+    CDFER single-file SQLite (primary, no 7z) ->
+    yaqwsx split-7z (fallback, needs 7z CLI) ->
+    official JLCPCB API (optional, if JLCPCB_APP_ID/API_KEY/API_SECRET set)
+
+Usage:
+    python download_jlcpcb.py [--source cdfer|yaqwsx|official] [--force]
 """
 
-import json
-import os
-import shutil
-import sqlite3
-import subprocess
+import argparse
 import sys
 import time
 from pathlib import Path
 
 sys.path.insert(0, str(Path(__file__).parent / "python"))
-from utils.platform_helper import PlatformHelper  # noqa: E402
 
-DATA_DIR = PlatformHelper.get_data_dir()
-DATA_DIR.mkdir(parents=True, exist_ok=True)
-
-CACHE_DIR = DATA_DIR / "jlcparts_cache"
-CACHE_DIR.mkdir(parents=True, exist_ok=True)
-
-BASE_URL = "https://yaqwsx.github.io/jlcparts/data"
-MAX_PARTS = 30  # probe up to this many split volumes
-
-TARGET_DB = DATA_DIR / "jlcpcb_parts.db"
-
-
-def curl_download(url: str, dest: Path) -> bool:
-    """Download url to dest with curl. Returns False on 4xx/5xx or network error."""
-    result = subprocess.run(
-        ["curl", "-L", "-f", "-o", str(dest), "--progress-bar", url], capture_output=False
-    )
-    return result.returncode == 0
-
-
-def download_files() -> bool:
-    """Download split archive volumes (z01..zNN) then cache.zip, stopping volumes at 404 or MAX_PARTS."""
-    print("Downloading jlcparts database (~421MB)...")
-
-    for i in range(1, MAX_PARTS + 1):
-        part = f"cache.z{i:02d}"
-        dest = CACHE_DIR / part
-        if dest.exists() and dest.stat().st_size > 1000:
-            print(f"  {part} already exists, skipping")
-            continue
-        print(f"  Downloading {part}...")
-        if not curl_download(f"{BASE_URL}/{part}", dest):
-            # -f causes non-zero exit on 4xx/5xx; treat as end of volumes
-            if dest.exists():
-                dest.unlink()
-            print(f"  {part} not found — {i - 1} volumes total")
-            break
-
-    dest = CACHE_DIR / "cache.zip"
-    if dest.exists() and dest.stat().st_size > 1000:
-        print("  cache.zip already exists, skipping")
-    else:
-        print("  Downloading cache.zip...")
-        if not curl_download(f"{BASE_URL}/cache.zip", dest):
-            print("  ERROR downloading cache.zip")
-            return False
-
-    return True
-
-
-def extract_database() -> bool:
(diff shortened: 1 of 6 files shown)


bhoot1234567890:
Quick note: while testing this on macOS cold-start, I found that the MCP connection still times out at 30s. The root cause is in **PR #198** (merged — my earlier warm-up fix): the `_warm_cache()` call in `SymbolLibraryManager.__init__()` parses 232 symbol libraries (~70s) **before** `{"type":"ready"}` is emitted, so the MCP client kills the node process during Phase 1 before the transport ever connects.

Fixed in **PR #206** with three changes:
- Defer `_warm_cache()` from constructor → lazy + runs during `_warmup` phase
- Swap MCP transport connect **before** warm-up in server.ts
- Symbol library init drops from 70s → 0.3s

PR #206 is against main and a blocker for this (PR #204) to work on macOS. Happy to rebase this on top of #206 or merge independently — whichever you prefer.

bhoot1234567890 (approved):
## Tested on macOS (Apple Silicon, KiCad 10.0.1)

### CDFER primary path ✅
| Metric | Result |
|---|---|
| Download | 1.5 GB in ~3s |
| Parts imported | 616,593 |
| Basic / Extended | 351 / 615,026 |
| DB size | 465 MB |
| Conversion time | ~49s (incl. FTS index) |
| Schema detection | `v_components` view used |

### Search tests ✅
```
nRF52832 → 3 results (C77540 NRF52832-QFAA-R, QFN-48, stock=407)
TPS63900 → 1 result  (C1518762 TPS63900DSKR, WSON-10, stock=6233)
0603 100nF → 3 results (C105907 0603X104K101T, stock=17414)
```

### Staleness warning ✅
```
Catalog from 'cdfer' is ~52 days old (dated Thu, 02 Apr 2026).
For fresher data, re-run with source='yaqwsx'.
```
Clear, actionable. The warning is loud, not silent.

### yaqwsx fallback
Not tested — requires 171 MB 7z download + 10 GB extraction (~30 min).
7z CLI available at `/opt/homebrew/bin/7z`.

### Bug found 🔴
`download_database(force=False)` does not check if target DB already exists — re-downloads unconditionally. On a second call, the resume logic 416-loops because the file is already complete (Range request beyond file size). The MCP handler (`_handle_download_jlcpcb_database`) should gate this with a DB-exists check, but the library function itself should also be safe to call twice.

### CDFER pipeline status (explains 52-day staleness)
Checked GitHub Actions — every run for 10+ days fails:
- **Primary:** JLCPCB Cart API returns HTTP 500 (blocking scrapers)
- **Secondary:** `NameError: name 'page_components' is not defined` in `scrape-jlcpcb.py` line 180
The yaqwsx upstream cache IS fresh (last-modified: May 24, 2026).

### macOS startup blocker 🔴
PR #198 introduced eager `_warm_cache()` which blocks `{"type":"ready"}` for ~70s. MCP client connection timeout is 30s → kicad server never connects on cold start. **PR #206** fixes this — needed before this PR works on macOS.

mixelpixx:
Fixed the force/416 re-download bug you caught (thanks again). `download_cdfer` now HEADs for `Content-Length` up front and short-circuits when the local cache file is already complete, and treats an in-loop 416 as "complete" (or drops a stale partial and restarts) rather than retrying it to failure.

Re-verified on Windows:
- Happy path unchanged: downloads + converts 616,593 parts in ~50s, staleness warning fires correctly.
- A pre-existing complete cache file now short-circuits in ~0.4s ("already fully downloaded; skipping") instead of 416-looping.
- New no-network regression test for the skip path; full suite at the known baseline (27 pre-existing failures only).

The macOS cold-start blocker you flagged is also resolved on `main` now (we took #210's root-cause fix: background-thread the warm-cache + fix the underlying paren-in-quoted-string parser). So this works on macOS too. Merging.

PR #207, opened 2026-05-26
feat(schematic): add angle and mirrorY to add_schematic_component

karu2003:
## Summary

- Adds \ngle\ (CCW rotation) and \mirrorY\ (horizontal flip) parameters to \dd_schematic_component\
- Rotates Reference/Value/Footprint property positions from library definitions
- Fixes property-position regex regression that broke RefDes/Value placement

Split from #194 per maintainer review. Closes #72.

## Test plan

- [ ] Add a BC547 with \mirrorY: true\ — symbol flips horizontally
- [ ] Add a component with \ngle: 90\ — pins and RefDes rotate correctly
- [ ] Verify Reference/Value use library positions with \(justify left)\, not resistor fallbacks

Made with [Cursor](https://cursor.com)

### DIFF
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -477,7 +477,7 @@ class DynamicSymbolLoader:
             # Iterate over every top-level (property ...) block in the symbol
             search_pos = 0
             while True:
-                m = re.search(r'\(property\s+"([^"]+)"\s+"[^"]*"\s+\(at\s+([-\d.]+)\s+([-\d.]+)\s+([-\d.]+))',
+                m = re.search(r'\(property\s+"([^"]+)"\s+"[^"]*"\s+\(at\s+([-\d.]+)\s+([-\d.]+)\s+([-\d.]+)\)',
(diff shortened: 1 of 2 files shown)


mixelpixx:
Reviewed and verified — merging. Thanks for splitting this out of #194 cleanly.

What I checked:
- Adds `angle` (CCW) + `mirror_y` to `create_component_instance`; both default to no-op, so existing callers are unaffected.
- RefDes/Value/Footprint/Datasheet positions + text angle + effects are now derived from the injected lib_symbols definition and rotated by the placement angle, instead of the old hardcoded ±2.54 offsets — the right approach, and it's what makes rotated/mirrored placements label correctly.
- `_rotate_offset` math checks out (e.g. (2.032,0)@90° → (0,-2.032); identity at 0°), and `_extract_paren_block` matches balanced blocks correctly.

Verification (Windows): `tsc` builds clean; `test_add_schematic_component.py` (15) and the broader placement-path tests (79) all pass — so the property-position change doesn't regress existing placement behavior. Closes #72 (the flip request).

PR #208, opened 2026-05-26
feat(schematic): opt-in Windows interactive schematic reload

karu2003:
## Summary

Opt-in Windows interactive schematic reload after MCP file edits — addresses PR #208 review (option 1: hardened auto-reload).

- **KICAD_INTERACTIVE_SCHEMATIC=true** (default off) enables reload after schematic add / edit / wire tools
- Sends Schematic Editor Revert (WM_COMMAND 20236), then auto-confirms only the reload/revert dialog
- No background dialog watcher — reload runs only on explicit post-edit hook

## Safety (per review)

| Requirement | Implementation |
|-------------|----------------|
| PID-scoped windows | KiCADProcessManager.get_running_pids() + GetWindowThreadProcessId on every hwnd |
| No generic dialogs | Rejects titles that are only Warning/Information/Confirmation; requires reload-specific keywords |
| No destructive auto-confirm | Skips dialogs with Discard/Don't Save/No/Cancel buttons or destructive title keywords |
| test_backend_metadata.py | Untouched — not deleted |

## Test plan

- [x] npm run build passes
- [x] tests/test_interactive_schematic.py (3 tests)
- [x] tests/test_backend_metadata.py still present and passing
- [ ] With KICAD_INTERACTIVE_SCHEMATIC=true on Windows: MCP add/edit component updates schematic without manual Revert
- [ ] Dialog offering discard of unsaved edits is not auto-confirmed

### DIFF
--- a/config/windows-config.example.json
+++ b/config/windows-config.example.json
@@ -8,6 +8,7 @@
         "PYTHONPATH": "C:\\Program Files\\KiCad\\9.0\\bin\\Lib\\site-packages",
         "LOG_LEVEL": "info",
         "KICAD_AUTO_LAUNCH": "false",
+        "KICAD_INTERACTIVE_SCHEMATIC": "false",
         "KICAD_MCP_DEV": "0"
       },
       "description": "KiCAD PCB Design Assistant - Note: PYTHONPATH auto-detected if venv exists"
--- a/docs/UI_AUTO_LAUNCH.md
+++ b/docs/UI_AUTO_LAUNCH.md
@@ -185,6 +185,7 @@ Claude:
 | Variable            | Default     | Description                    |
 | ------------------- | ----------- | ------------------------------ |
 | `KICAD_AUTO_LAUNCH` | `false`     | Auto-launch KiCAD when needed  |
+| `KICAD_INTERACTIVE_SCHEMATIC` | `false` | Windows only: after MCP schematic edits, send Revert to reload from disk (PID-scoped; never auto-confirms discard dialogs) |
 | `KICAD_EXECUTABLE`  | auto-detect | Override KiCAD executable path |
 
 ### Custom Executable Path
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,10 @@ module = [
     "schematic",
     "PIL",
     "PIL.*",
+    "win32api",
+    "win32con",
+    "win32gui",
+    "win32process",
 ]
 ignore_missing_imports = true
 
--- a/python/commands/schematic_handlers.py
+++ b/python/commands/schematic_handlers.py
@@ -26,6 +26,7 @@ from commands.library_schematic import LibraryManager as SchematicLibraryManager
 from commands.schematic import SchematicManager
 from commands.wire_manager import WireManager
 from utils.kicad_cli import kicad_cli_not_found_message, resolve_kicad_cli
+from utils.interactive_schematic import reload_kicad_schematic
 from utils.sexpr_format import dumps as kicad_dumps
 
 logger = logging.getLogger("kicad_interface")
@@ -96,6 +97,10 @@ class SchematicHandlersMixin:
     # mypy can resolve ``self.board`` when type-checking this module in isolation (the
     # module docstring notes the mixin relies on the host's ``self.board``).
     board: Any
+
+    def _reload_kicad_schematic(self) -> None:
+        """Opt-in: reload open Schematic Editor after external file edits (Windows)."""
+        reload_kicad_schematic()
 
     def _handle_create_schematic(self, params: Dict[str, Any]) -> Dict[str, Any]:
         """Create a new schematic"""
@@ -222,6 +227,8 @@ class SchematicHandlersMixin:
                 unit=unit,
                 project_path=derived_project_path,
             )
+
+            self._reload_kicad_schematic()
 
             return {
                 "success": True,
@@ -640,6 +647,7 @@ class SchematicHandlersMixin:
                 changes["propertiesRemoved"] = properties_removed
 
             logger.info(f"Edited schematic component {reference}: {changes}")
+            self._reload_kicad_schematic()
             return {"success": True, "reference": reference, "updated": changes}
 
         except Exception as e:
@@ -931,6 +939,7 @@ class SchematicHandlersMixin:
                 message = "Wire added successfully"
                 if snapped_info:
                     message += "; " + "; ".join(snapped_info)
+                self._reload_kicad_schematic()
                 return {"success": True, "message": message}
             else:
                 return {"success": False, "message": "Failed to add wire"}
--- a/python/utils/interactive_schematic.py
+++ b/python/utils/interactive_schematic.py
@@ -1,0 +1,240 @@
+"""Opt-in Windows interactive schematic reload after MCP file edits.
+
+Sends KiCad Schematic Editor Revert (reload from disk) and auto-confirms only
+the specific reload/revert dialog. Every window handle is PID-scoped to KiCad;
+destructive discard/unsaved dialogs are never auto-confirmed.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import re
+import time
+from typing import Optional
+
+from utils.kicad_process import KiCADProcessManager
+
+logger = logging.getLogger("kicad_interface")
+
+INTERACTIVE_SCHEMATIC = os.environ.get("KICAD_INTERACTIVE_SCHEMATIC", "false").lower() == "true"
+
+if INTERACTIVE_SCHEMATIC:
+    logger.info("KiCAD interactive schematic reload enabled (KICAD_INTERACTIVE_SCHEMATIC=true)")
+
+REVERT_CMD_ID = 20236
+
+# Affirmative labels for reload-from-disk confirmation only.
+_AFFIRMATIVE_LABELS = frozenset(
+    {
+        "Yes",
+        "&Yes",
+        "OK",
+        "&OK",
+        "Ja",
+        "&Ja",
+        "Oui",
+        "&Oui",
+        "Reload",
+        "&Reload",
+        "Neu laden",
+        "&Neu laden",
+    }
+)
+
+# Never auto-click these — data-loss risk (#194 / #244).
+_DESTRUCTIVE_LABELS = frozenset(
+    {
+        "Discard",
+        "&Discard",
+        "Discard Changes",
+        "Don't Save",
+        "Do Not Save",
+        "Do not save",
+        "No",
+        "&No",
+        "Cancel",
+        "&Cancel",
+        "Abort",
+        "&Abort",
+        "Verwerfen",
+        "&Verwerfen",
+        "Nicht speichern",
+    }
+)
+
+_GENERIC_ONLY_TITLES = frozenset(
+    {
+        "Warning",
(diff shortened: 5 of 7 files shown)


mixelpixx:
Thanks for splitting this out of #194 and making it opt-in — the gating is exactly right (`KICAD_INTERACTIVE_SCHEMATIC` defaults off, Windows-only `os.name != "nt"` guard, pywin32 soft-gated). Two things I need addressed before this can merge, though.

## 1. It deletes `tests/test_backend_metadata.py` (605 lines) — please restore it

This branch forked *after* #202 merged (merge-base is the #202 merge), and it removes `tests/test_backend_metadata.py`, which is the backend-state test coverage from #202/#140. Merging would delete that merged coverage. My earlier #194 note ("drop test_backend_metadata.py from this branch") meant *don't carry your conflicting copy* — but since the file now lives on `main`, the branch should leave it untouched rather than delete it. Could you `git checkout main -- tests/test_backend_metadata.py` (restore it) so the PR no longer touches it?

## 2. The dialog watcher is too broadly scoped for the data-loss risk

This is the concern from the original #194 review and it's still live. As written, `_start_kicad_dialog_watcher`:

- Enumerates **all** top-level windows via `EnumWindows` and matches on **title strings** — it never verifies the window belongs to KiCad's process (PID). A dialog from any app titled "Warning"/"Information"/"Confirmation" with an OK/Yes button is in scope.
- `RELOAD_KEYWORDS` includes very generic words — `"Warning"`, `"Information"`, `"Confirmation"` — so it matches far more than reload dialogs.
- It then clicks `YES_LABELS` (`OK`/`Yes`/`Reload`/`Discard`/…), and the keyword list even includes `"Discard"` and `"Unsaved"`. Auto-clicking Yes/Discard on a "discard unsaved changes?" dialog is exactly the unrecoverable data-loss case we were worried about.

Even behind the opt-in flag, a user who enables it for auto-reload shouldn't risk losing unsaved work. Could you tighten it to:
- **Scope to KiCad's process** — resolve each hwnd's PID (`GetWindowThreadProcessId`) and only act on windows owned by the KiCad process you launched/track.
- **Match the specific reload/revert dialog only**, not generic "Warning"/"Information"/"Confirmation".
- **Never auto-confirm a destructive action** — drop `Discard`/`Unsaved` from the match, and skip any dialog whose buttons indicate a discard/overwrite choice rather than a plain reload.

Once the test file is restored and the watcher is PID-scoped + narrowed away from destructive dialogs, I'm happy to take another look. The interactive-reload feature itself is genuinely useful — just want it to be safe-by-construction for the people who opt in.

mixelpixx:
Saw the rebase — thanks. Just flagging that the two blockers from my review above are still open in the current diff:

1. It still deletes `tests/test_backend_metadata.py` (+0/-605) — that's merged coverage from #202, please restore it (`git checkout main -- tests/test_backend_metadata.py`) so this PR doesn't remove it.
2. The dialog watcher still matches generic titles ("Warning"/"Information"/"Confirmation") across all windows and can click affirmative on destructive ("Discard"/"Unsaved") dialogs — needs to be PID-scoped to KiCad and never auto-confirm a discard.

Happy to merge once those two are addressed — the opt-in gating itself is already right.

mixelpixx:
Status: still holding on the two review points — the PR deletes `tests/test_backend_metadata.py` and the dialog watcher auto-confirms arbitrary KiCad dialogs, which we can't take as-is. If you trim it to the opt-in reload without the watcher (and restore the test), we're happy to merge that subset.

karu2003:
Thanks for the clear review — I want to land the mergeable subset you described (opt-in reload without the watcher, and restore the test). Before I push the rebase, one clarification on the reload path, because the trim is a bit more involved than it looks.

**On the test:** agreed — I'll `git checkout main -- tests/test_backend_metadata.py` so this PR no longer touches it.

**On the dialog-dismiss / data-loss concern:** the unsafe pattern isn't only in `_start_kicad_dialog_watcher` — the same logic is also baked into `_reload_kicad_schematic` itself (the method the subset would keep). It currently:

- enumerates **all** top-level windows via `EnumWindows` with no PID check (doesn't verify the window belongs to KiCad),
- matches generic titles (`Warning` / `Information` / `Confirmation`) and includes `Discard` / `Unsaved` in the keyword list,
- sends the `Revert` `WM_COMMAND` and then auto-clicks `Yes`/`OK` on the resulting confirmation.

So just deleting the watcher wouldn't fully address blocker #2 — the reload method would still auto-confirm a destructive Revert. Which direction do you prefer for the kept reload path?

1. **Hardened auto-reload** — PID-scope every `hwnd` to the KiCad process, match only the specific schematic reload/revert dialog, and drop `Discard`/`Unsaved` so it never auto-confirms a destructive choice; or
2. **No auto-dismiss at all** — just send `Revert` to the KiCad schematic window (resolved by PID) and let the user confirm the dialog manually.

I'm happy to implement either — option 1 keeps it fully hands-free, option 2 is maximally safe-by-construction.

**One heads-up on the rebase:** since this branch was opened, `main` extracted the schematic handlers into `commands/schematic_handlers.py` (`SchematicHandlersMixin`, #214), and `_reload_kicad_schematic` doesn't exist on `main` at all. So the rebased PR will effectively re-integrate the opt-in reload into the new mixin rather than re-add the old inline handlers — the final diff should be small and focused on the reload hook + `create_project` KiCad-cli upgrade. Just flagging so the diff shape makes sense when it lands.

Let me know which reload option you'd like and I'll push the rebased, trimmed version.

mixelpixx:
Thanks for digging into where the auto-confirm actually lives — you're right that deleting the watcher alone would leave the same pattern inside `_reload_kicad_schematic`, and that's exactly the kind of detail I wanted surfaced before a rebase.

**Go with option 1 — the hardened auto-reload.** Reasoning: the feature exists so an unattended MCP session can push schematic changes and have the open KiCad window follow along. Option 2's manual confirmation stalls precisely in the scenario the feature targets (nobody at the keyboard), which would make the opt-in flag mostly decorative. Option 1 keeps the feature useful and removes the data-loss vector by construction rather than by policy.

Concretely, the acceptance bar I'll review against:

1. Every window handle from `EnumWindows` is PID-scoped to the KiCad process before anything is inspected or clicked. No window that doesn't belong to KiCad is ever touched.
2. The dialog matcher targets only the specific schematic reload/revert confirmation, not generic `Warning`/`Information`/`Confirmation` titles.
3. `Discard` and `Unsaved` are removed from the keyword list entirely. If KiCad raises a dialog offering to discard unsaved user edits, the reload must stop and report, never auto-confirm. Losing a user's manual edits silently is the one outcome this feature is not allowed to produce (see #244 for how much that class of bug hurts).
4. `tests/test_backend_metadata.py` restored to main's version, as you already agreed.
5. The whole path stays behind the existing opt-in (`KICAD_INTERACTIVE_SCHEMATIC` default-off, Windows-only guard, pywin32 soft-gated) — that gating was right in the original PR.

On the rebase shape: understood and no concern — re-integrating via the `SchematicHandlersMixin` is the correct landing spot now that #214 moved the handlers, and a smaller focused diff is a feature. One heads-up: main has moved again since your last look (#259, #260, #261, #193 merged this week), so rebase onto current main to save yourself a second pass.

karu2003:
@mixelpixx — rebased onto current main and implemented option 1 (hardened auto-reload) from your Jul 2 comment. tests/test_backend_metadata.py not touched; background watcher removed; PID-scoped reload with no destructive auto-confirm. Ready for re-review.

karu2003:
Follow-up: exclude Schematic Editor main-window titles from the reload dialog matcher and skip the schematic hwnd during post-Revert dialog scan (avoids false positive on the broad \schematic\ keyword). Tests updated.

mixelpixx:
Thanks for the trimmed resubmission. I reviewed the current diff locally against main (merges cleanly) and the "option 1" hardening is implemented correctly — I verified each point in `python/utils/interactive_schematic.py`:

- Every `EnumWindows`/`EnumChildWindows` pass is PID-scoped: `_owned_by_kicad(hwnd)` checks `GetWindowThreadProcessId(hwnd)` against `KiCADProcessManager.get_running_pids()`, so windows from other processes are never touched. Both the Schematic Editor lookup and the dialog scan apply this filter.
- Only the specific reload/revert prompt is auto-confirmed: `is_reload_confirmation_title()` requires a reload/revert keyword and explicitly excludes generic dialog titles, main editor frames, and destructive titles.
- Discard/unsaved dialogs are never auto-confirmed — refused at both the title level (`_DESTRUCTIVE_TITLE_KEYWORDS`) and the button level (`has_destructive_button`). Confirmation only proceeds when a safe affirmative button is found.
- Confirmation clicks the specific matched button via `PostMessage(child, BM_CLICK)` rather than a blind Enter/Return, so a focus race can't dismiss the wrong dialog.
- `tests/test_backend_metadata.py` is restored (not in the diff — matches main).

Gates:
- `tests/test_interactive_schematic.py` is present; `black --check` is clean on the changed files.
- Full suite shows no new failures attributable to this PR — the only failures on this branch also reproduce on clean main (the Java-not-found autoroute/freerouting tests plus one kicad-cli-resolver test that depends on the install location), so nothing here is a regression.

One item before this is mergeable: mypy is no longer clean. `python -m mypy python/` reports 4 new errors, all from the new module:

```
python/utils/interactive_schematic.py:130: error: Library stubs not installed for "win32api"  [import-untyped]
python/utils/interactive_schematic.py:131: error: Library stubs not installed for "win32con"  [import-untyped]
python/utils/interactive_schematic.py:132: error: Library stubs not installed for "win32gui"  [import-untyped]
python/utils/interactive_schematic.py:133: error: Library stubs not installed for "win32process"  [import-untyped]
```

The imports themselves are fine — they're correctly guarded in a `try/except ImportError`. This is only missing third-party stubs. The project already handles stub-less dependencies through a mypy override list, so the consistent fix is one line: add the four modules to the existing `[[tool.mypy.overrides]]` block in `pyproject.toml` (which already lists `pcbnew`, `cairosvg`, `skip`, `fitz`, `kipy`, etc.) so they inherit `ignore_missing_imports = true`. Please avoid per-line `# type: ignore` here since the override list is the established pattern in this repo.

With that one change the mypy gate goes green and this is ready to merge. The reload logic itself is exactly what was asked for.

mixelpixx (approved):
The option-1 hardening is implemented exactly to the bar we agreed: every EnumWindows hit is PID-scoped to actual KiCad processes, the dialog matcher requires a reload-specific title and rejects generic titles, destructive title keywords, and main editor frames (your follow-up), and the destructive-label set is a hard never-click list — with multi-language coverage as defensive depth. tests/test_backend_metadata.py is untouched, the feature is opt-in (KICAD_INTERACTIVE_SCHEMATIC=true) and Windows-gated, and the pywin32 mypy overrides keep the type gate green on non-Windows checkers.

Verified locally: all 20 tests in the interactive-schematic and backend-metadata files pass, black/mypy clean, full suite at the known-environmental baseline. One small push to your branch (maintainer edit): test_interactive_schematic.py imported utils.* without the sys.path bootstrap, so it only collected when another test file had already inserted python/ — fine in the full suite, a collection error standalone (the order-dependence class we track in #287). Added the standard two-line bootstrap.

Thanks for the patience through the long arc of this one — the trimmed, hardened version is something we can ship without the data-loss worry the original watcher carried. Merging.

PR #210, opened 2026-05-29
Fix MCP tool launch timeout on macos / unable to connect

MoChahadeh:
The kicad MCP showed ✗ Failed to connect with a very long wait. Tracing the
  startup:

  1. The Node server (src/server.ts) deliberately does not connect to the MCP
  transport until the Python subprocess emits {"type":"ready"} and a warm-up
  completes. So if Python is slow to signal ready, Claude Code sees a long hang
  followed by a connection failure.
  2. Python never emitted "ready". KiCADInterface.__init__ constructs
  SymbolLibraryCommands() → SymbolLibraryManager(), whose __init__ called
  _warm_cache() synchronously — parsing every .kicad_sym library before the
  ready handshake.
  3. The parser itself was the killer. In
  library_symbol.py::_parse_kicad_sym_file, the routine that finds a symbol
  block's end walked the text counting (/) without skipping quoted strings.
  KiCAD descriptions routinely contain parens (e.g. "MCU (LQFP-64)"), which
  unbalanced the depth counter so the walk ran to end-of-file for every
  subsequent symbol — turning each multi-MB library (the STM32H7 file is 15.5
  MB) into an O(n²) scan.

  I measured it: 60 symbols of one file took 65 seconds, with 42 falsely flagged
  "malformed". Warming all ~200 libraries would take many minutes to over an
  hour, so "ready" was never sent within the timeout → failed connect.

  Fix

  Two changes in python/commands/library_symbol.py:

  1. String-aware paren walker — the block-end scan now skips over double-quoted
  strings (with escape handling), so parens inside property values no longer
  
  1. String-aware paren walker — the block-end scan now skips over double-quoted
  strings (with escape handling), so parens inside property values no longer
  break depth tracking. Same file: all 225 symbols in 2.2s, 0 malformed (was ~4
  min for the full file).
  2. Non-blocking warm-up — _warm_cache() now runs on a daemon thread instead of
  blocking __init__, so the READY handshake fires immediately and the cache
  fills in the background. Added a threading.Lock around the cache write so the
  background warm and any on-demand list_symbols parse can't race.
  
  Verification
  
  - Python now emits {"type":"ready"} immediately.
  - Full Node→Python startup completes in under 600 ms (previously never).
  - initialize + tools/list round-trip succeeds, returning 142 tools.
  - claude mcp list → kicad: ... ✓ Connected.


mixelpixx:
Reviewed and verified — merging. Thanks for tracking this to the root cause.

**What this fixes:** the macOS cold-start regression from #198 (eager `_warm_cache()` blocking `{"type":"ready"}` past the 30s client timeout). Two real fixes here: warm-cache now runs on a daemon thread (non-blocking), and `_parse_kicad_sym_file` is made string-aware so parens inside quoted descriptions (e.g. `"MCU (LQFP-64)"`) no longer send the depth counter to EOF — that was an O(n²) minutes-per-file scan, the actual killer.

**Verification (Windows):**
- `tsc` builds clean.
- Full suite back to the known baseline (27 pre-existing failures: Java-not-installed + the mil-enum gap + sync test-ordering pollution that passes in isolation) — **no new regressions**.
- Confirmed the string-aware parser still parses the SPICE symbols correctly (OPAMP/PJFET, `${SIM.PARAMS}`, sim_pins).

**One fix I pushed to your branch** (maintainer edit): the 8 `test_symbol_library.py::TestSymbolLibraryManagerParsing` tests were failing with `AttributeError: _cache_lock` because `list_symbols()` now guards cache writes with `self._cache_lock`, but `_manager_for_fixture()` builds the manager via `__new__` (to skip disk I/O) and so bypasses `__init__` where the lock is created. Added one line to set the lock in the fixture; all 8 pass. Production construction always runs `__init__`, so this was test-only.

**Minor note for future PRs:** the `package-lock.json` dependency bumps are unrelated churn — harmless here (benign version updates), but cleaner to keep out of a focused fix. No action needed now.

PR #211, opened 2026-05-29
Fix: reference-based schematic tools fail when lib_symbols contains a paren inside a quoted string

MoChahadeh:
get_schematic_component, edit_schematic_component, and delete_schematic_component would return Component 'X' not found in schematic for
  every reference in certain schematics — even components that demonstrably exist in the file. The root cause is a hand-rolled S-expression
  paren matcher that counted parentheses inside quoted strings.
  
  Root cause

  These three handlers resolve a reference by text-walking the .kicad_sch file. They locate the (lib_symbols …) block (to skip it) using a
  depth-counting paren matcher, then scan the placed-symbol blocks that follow it.

  The matcher (_find_matching_paren plus two duplicated inline copies) counted ( and ) without skipping quoted strings. KiCad does not
  backslash-escape bare parens inside quoted strings — and they're common:

  - MCU pin names such as "PA13(JTMS", "PA14(JTCK", "PB3(JTDO" (STM32 alternate-function names split at commas — 630+ such strings exist in
  the stock KiCad symbol libraries)
  - descriptions such as "Vin(fwd) 40V"
  
  A single unbalanced ( inside a quoted string within lib_symbols makes the depth counter never rebalance, so lib_sym_end runs to near-EOF.
  Because every placed symbol sits after lib_symbols, they all then satisfy lib_sym_pos <= pos <= lib_sym_end and get skipped → every
  reference resolves to "not found".
  
  The S-expression–aware tools (list_schematic_components, get_schematic_pin_locations) parse via sexpdata/kicad-skip, which handle quoted
  strings correctly — which is why the same references resolved in one code path but not the other.

  This is the same bug class fixed earlier in library_symbol.py's _parse_kicad_sym_file (commit 36820d6); three copies in the schematic
  handlers were missed.

  Fix

  - Made the static _find_matching_paren string-aware: it now tracks an in_string flag and honors \ escapes, so parens inside quoted
  strings are ignored. Corrected its docstring, which had asserted the false "parens are backslash-escaped" assumption.
  - Routed the two inline closures (in the delete/get handlers) through the fixed static method, removing the duplicated naive logic.

  The write-path handlers intentionally use text manipulation rather than kicad-skip (to avoid reformatting the file on save), so the fix
  hardens the matcher in place instead of swapping in a full parser.

  Testing

  - Added tests/test_schematic_lookup_paren_in_string.py (5 tests) reproducing the failure with an STM32-style "PA13(JTMS" pin name and
  verifying get/edit/delete now succeed — including that edit actually rewrites the value, delete removes only the placed instance (library
  definition preserved), and a genuinely-absent reference still correctly returns "not found" (guards against over-matching).
  - Existing suites pass with no regressions: test_delete_schematic_component (13), plus edit / fields / properties / pin_locator (66).

  Notes

  - A previously reported board-state hypothesis was a red herring — these handlers read schematicPath directly and never touch board
  state.
  - The minor pin-orientation reporting discrepancy between get_schematic_pin_locations and list_schematic_components is unrelated to this
  bug and left for a separate change.

### DIFF
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -1408,19 +1408,11 @@ class KiCADInterface:
             with open(sch_file, "r", encoding="utf-8") as f:
                 content = f.read()
 
-            def find_matching_paren(s: str, start: int) -> int:
-                """Find the closing paren matching the opening paren at start."""
-                depth = 0
-                i = start
-                while i < len(s):
-                    if s[i] == "(":
-                        depth += 1
-                    elif s[i] == ")":
-                        depth -= 1
-                        if depth == 0:
-                            return i
-                    i += 1
-                return -1
+            # String-aware paren matcher (see _find_matching_paren): a naive
+            # counter over-runs on unescaped parens inside quoted strings (e.g.
+            # MCU pin names like "PA13(JTMS"), which would extend lib_symbols to
+            # EOF and make every placed-symbol lookup fail.
+            find_matching_paren = self._find_matching_paren
 
             # Skip lib_symbols section
             lib_sym_pos = content.find("(lib_symbols")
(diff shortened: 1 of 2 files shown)


mixelpixx:
Reviewed and verified — merging. Nice companion to #210: that fixed the same paren-in-quoted-string bug in the symbol-library parser; this fixes it in the schematic-lookup path and de-dups the three hand-rolled matchers in get/edit/delete_schematic_component onto the now string-aware `_find_matching_paren`.

Verification (Windows):
- `tsc` builds clean.
- New `test_schematic_lookup_paren_in_string.py` passes (5).
- Full schematic test set + full suite back to the known baseline (27 pre-existing failures only — Java/mil/sync-pollution), 830 passed. No new regressions; normal (no-paren-in-string) schematics unchanged.

Thanks for catching that the reference-based tools and the kicad-skip path disagreed — good find.

Issue #84, opened 2026-04-04
Is KiCad 10 supported?


raffaeler:
I read the notes about SWIG, but still can't figure out whether KiCad10 is supported or not

Error: `MCP server 'KiCad-MCP-Server': Python process for KiCAD scripting is not running`
In Preferences - Plugins I activated the `Enable KiCad API` and installed the `Kicad-Python` package as specified [here](https://github.com/mixelpixx/KiCAD-MCP-Server/issues/31).

Is it supported?

TIA

scorp508:
Mine seems to be working, but it took some work. I had KiCAD 9 installed for all users, but KiCAD 10 installed only for myself (I don't know why I chose that). This meant multi-version path differences the MCP config didn't seem setup to account for. 

I had Claude Code look through all of the MCP's config and updated paths to be v10. I then had to install the python dependencies under v10's python. There may have been a couple other things I'm forgetting, but at the moment Schematic and PCB layout seems to be working the same way it was in v9.

66-firebat:
Could you elaborate on how you set up your kicad-mcp server and the repo files to be compatible with the IPC-API instead of swig? I am still running on the SWIG bindings:

<img width="1920" height="1041" alt="Image" src="https://github.com/user-attachments/assets/139f7a59-3f6e-42af-9f00-7f117bc3515b" />

scorp508:
I had to install these packages by running python FROM the KiCAD 10 python location in /KiCAD/10.0/bin/

`python.exe -m pip install kicad-skip Pillow cairosvg colorlog pydantic requests python-dotenv typing-extensions
`

My mcp config for Claude is...

```
{
  "mcpServers": {
    "kicad": {
      "command": "node",
      "args": ["C:/KiCAD-MCP-Server/dist/index.js"]
    }
  }
}
```

Then Claude made four edits in two files (kicad-server.ts and server.ts) in /src/ directory of the MCP bits to work with KiCAD 10. I've included each area below. I'm on Windows, so if it were a different OS they changes may be different. I think I told it to remove versions before 10.0, so you may see other lines in your .ts files.


**kicad-server.ts**
**=====EDIT 1=====**
```
async start() {
    try {
      console.error("Starting KiCAD MCP server...");

      // Start the Python process for KiCAD scripting
      console.error(`Starting Python process with script: ${this.kicadScriptPath}`);
[...]

PR #213, opened 2026-05-30
fix(windows): derive PYTHONPATH from the detected KiCAD python (#84)

mixelpixx:
Addresses the recurring manual-fixup in #84.

## Problem
`findPythonExecutable` already auto-detects KiCAD's bundled python across versions (10.0/9.0/…) and across both `C:\Program Files\KiCad` and per-user `%LOCALAPPDATA%\Programs\KiCad`. But the Python subprocess spawn still defaulted `PYTHONPATH` to a hardcoded `C:/Program Files/KiCad/9.0/lib/python3/dist-packages`. So a KiCad 10 (or per-user) user whose `python.exe` was found correctly still got the wrong `PYTHONPATH` and a `pcbnew` import failure unless they set `PYTHONPATH` by hand — exactly the edits people keep posting in this thread.

## Fix
`deriveKiCadSitePackages(pythonExe)`: for a detected KiCAD python at `<root>/<version>/bin/python.exe`, return the matching site-packages (`<...>/bin/Lib/site-packages`, or the older `lib/python3/dist-packages` layout), verified to exist. Use it as the `PYTHONPATH` default.

Precedence (top wins): explicit `PYTHONPATH` env override → derived KiCAD site-packages → legacy 9.0 string as last resort. So it **cannot regress below current behavior** — it only improves the default when no env override is set.

Path math verified for both install locations:
- `…\Local\Programs\KiCad\10.0\bin\python.exe` → `…\Local\Programs\KiCad\10.0\bin\Lib\site-packages`
- `C:\Program Files\KiCad\10.0\bin\python.exe` → `C:\Program Files\KiCad\10.0\bin\Lib\site-packages`

`tsc` builds clean. Gated to Windows + a path containing `kicad`, so venv/system pythons are unaffected.

## Verification ask
I can't exercise a real KiCAD spawn in CI. @scorp508 / @66-firebat / @NiNjA-CodE — if any of you can run this branch on KiCad 10 (Program Files or per-user) **without** manually setting `PYTHONPATH` and confirm `pcbnew` imports, that's the confirmation I'd want before merging.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -72,6 +72,29 @@ function getWindowsKiCadPythonCandidates(): string[] {
 }
 
 /**
+ * Derive the KiCAD bundled-Python site-packages path for a detected python.exe,
+ * so PYTHONPATH follows the *same* install we picked (any version, Program Files
+ * or per-user %LOCALAPPDATA%) instead of a hardcoded KiCad 9.0 path.
+ *
+ * KiCAD on Windows installs python at `<root>/<version>/bin/python.exe`, with
+ * pcbnew under `<...>/bin/Lib/site-packages` (older/alt layouts use
+ * `<version>/lib/python3/dist-packages`). Returns the first existing candidate,
+ * or undefined if pythonExe isn't a KiCAD bundled python.
+ */
+function deriveKiCadSitePackages(pythonExe: string): string | undefined {
+  if (process.platform !== "win32") return undefined;
+  const lower = pythonExe.toLowerCase();
+  if (!lower.endsWith("python.exe") || !lower.includes("kicad")) return undefined;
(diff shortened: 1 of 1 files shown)


mixelpixx:
Merging on test strength — this only improves the PYTHONPATH default (an explicit `PYTHONPATH` env still wins, and the legacy 9.0 string remains as a last-resort fallback), so it can't regress below current behavior. Rebased clean onto current main, `tsc` builds clean, path-derivation logic verified for both Program Files and per-user KiCad 10 installs. A Windows + KiCad 10 confirmation is still welcome (cc @scorp508 / @66-firebat / @NiNjA-CodE) — but this unblocks the common case now.

PR #214, opened 2026-05-30
refactor: extract schematic handlers into SchematicHandlersMixin (kicad_interface.py 6.7k→3.7k)

mixelpixx:
## Why
`python/kicad_interface.py` had grown to **6,728 lines** — the largest file in the repo by far. Most domains already delegate to dedicated classes in `commands/` (`board_commands`, `routing_commands`, etc.), but the **schematic** domain was never extracted: 43 `_handle_*` methods plus the schematic-only `_svg_to_png` helper (~3,000 lines) lived inline in the dispatcher.

## What
Move that domain into `python/commands/schematic_handlers.py` as `SchematicHandlersMixin`, and make `KiCADInterface(SchematicHandlersMixin)`.

- `kicad_interface.py`: **6,728 → 3,665 lines**
- new `commands/schematic_handlers.py`: 3,099 lines

## Why it's safe — pure mechanical relocation, no logic change
The mixin methods stay bound to the same instance, so every `self.board`, `self._auto_save_board()`, cross-handler call, and the `command_routes` table resolve exactly as before (via MRO). To prove there's no behavior change:

- **AST-identity verified:** every one of the 132 methods (89 kept + 43 moved) and the relocated `_svg_to_png` is byte-for-byte AST-identical between the original and the split. Nothing was edited, only relocated.
- **Full test suite at the exact prior baseline:** 839 passed; the 27 failures are the same pre-existing env/pollution set (Java-not-installed, the mil-enum gap, the sync test-ordering pollution). All ~100 schematic tests pass, including the paren-matcher regression tests from #211.
- `tsc` builds clean; `flake8 --select=F` clean (no undefined names / unused imports); black-clean.

## Reviewer note
`git diff` shows some *kept* methods (footprint/netlist/net handlers) as moved — that's a diff-alignment artifact of deleting 43 interleaved methods from a 6.7k-line file, not real changes. The AST-identity check above is the ground truth: those bodies are unchanged.

## Scope
First step only — schematic domain. A natural follow-up could extract the 7 `_handle_ipc_*` handlers the same way, but I kept this PR to one domain so it's reviewable as "moved, not modified." Posting for input before merge given it touches the central file.


mixelpixx:
Im looking for feedback before pulling the trigger.  The file size was just getting out of hand, and editing it is a pain.  I just want someone else to look at this before i click '**Merge**'

mixelpixx:
Regenerated against current main (post merge wave: #226-#229, #231-#234, #238, #252, #253).

- 43 schematic handlers + `_svg_to_png` (~3,100 lines) moved verbatim into `commands/schematic_handlers.py` as `SchematicHandlersMixin`; `kicad_interface.py` goes 8,596 -> 5,528 lines.
- All 159 original functions/methods verified AST-identical between original and split.
- Scope deliberately excludes the 19 new `_handle_export_*` kicad-cli handlers from #253 (export domain, not schematic).
- One test file updated: `test_sync_schematic_to_board_footprints.py` patch targets now point at the moved module (`commands.schematic_handlers.pcbnew`) - previously masked by test-ordering pollution that #253's collection-order change cured.
- Full suite: 8 failed / 1044 passed - exactly the current main baseline (the 8 are the Java-not-installed freerouting/autoroute set). `npm run build` clean.

Holding for maintainer review before merge.

PR #218, opened 2026-06-01
fix(library): rebuild SymbolLibraryManager when cache is empty

dubanoze:
## Problem

When a project is opened for the first time (e.g. right after `create_project`),
the `sym-lib-table` may not exist on disk yet. `SymbolLibraryManager.__init__`
succeeds but leaves `self.libraries = {}`.

The original guard in `use_project()` fired immediately because the path already
matched the newly-created (empty) manager, so the manager was never rebuilt once
the `sym-lib-table` appeared on disk. All subsequent `list_symbols` / `search_symbols`
calls returned nothing.

## Fix

Also require that at least one library was loaded before treating the cache as valid:

    if (self.library_manager.project_path == project_path
            and len(self.library_manager.libraries) > 0):
        return

## Tests

Added `tests/test_symbol_library_empty_cache_rebuild.py` — 3 unit tests, all pass:
- `test_rebuild_triggered_when_libraries_empty` — reproduces the bug
- `test_no_rebuild_when_libraries_already_loaded` — happy path unchanged
- `test_rebuild_on_different_project_path` — project-switch behaviour

### DIFF
--- a/python/commands/library_symbol.py
+++ b/python/commands/library_symbol.py
@@ -607,7 +607,11 @@ class SymbolLibraryCommands:
         """
         if project_path is None:
             return
-        if self.library_manager.project_path == project_path:
+        # Patch (SER2RJ45): keep cache when project_path matches AND libraries were
+        # actually loaded. Original early-return skipped rebuild even when the cache
+        # was empty (e.g. first call after create_project, before sym-lib-table existed).
+        if (self.library_manager.project_path == project_path
+                and len(self.library_manager.libraries) > 0):
             return
         logger.info(f"Rebuilding SymbolLibraryManager for project: {project_path}")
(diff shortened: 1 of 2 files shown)


mixelpixx:
Reviewed — clean, well-targeted fix. The original early-return in `use_project()` skipped the rebuild whenever `project_path` matched, even if `libraries` was empty — which is exactly the first-open-after-create_project case where `sym-lib-table` doesn't exist yet, leaving symbol lookups permanently empty. Adding `len(libraries) > 0` to the guard fixes it. Safe against the #210 background warm-up (the `libraries` dict is populated on the main thread in `__init__` before the warm thread starts; the `_cache_lock` guards `symbol_cache`, a different dict — no race). Tests cover all three paths. Full suite at baseline. Merging — thanks!

PR #196, opened 2026-05-22
feat(schematic): expose justify directive on field labels via MCP

gcolonese:
## Summary

This PR adds `justify` support to schematic field label positioning in the KiCAD MCP Server, exposing KiCad's native `(justify ...)` S-expression directive through both `edit_schematic_component` and `set_schematic_component_property`.

### Feature Ask

When placing components in a schematic, KiCad stores text justification for field labels (Reference, Value, Tolerance, Rating, etc.) as a `(justify ...)` directive inside the field's `(effects ...)` block:

```
(property "Reference" "R1" (at 51.27 47.46 0)
  (effects (font (size 1.27 1.27)) (justify left))
)
```

This is essential for workflows where Reference designator labels are hand-positioned with per-instance `(justify left)` / `(justify right)` directives, and the same justification needs to be mirrored onto Value / Tolerance / Rating fields.

### Changes

**Python backend (`python/kicad_interface.py`)**
- New `_set_justify_on_property()` helper that adds, replaces, or removes the `(justify ...)` token inside a property's `(effects ...)` block
- Passing `"center"` removes the directive entirely (KiCad's default — omitting `(justify ...)` means centered)
- Integrated into `_set_property_in_block()` (the `properties` dict path) and into the `field_positions` loop in `_handle_edit_schematic_component()`
- `_handle_set_schematic_component_property()` now forwards `justify` from params to the spec dict

**TypeScript MCP layer (`src/tools/schematic.ts`)**
- `fieldPositions` schema extended: each field entry now accepts an optional `justify?: string | string[]`
- Array form (e.g. `["left", "top"]`) is normalised to `"left top"` before the Python call
- `set_schematic_component_property` gains a `justify?: string` parameter

### KiCad Justify Syntax Reference

The `(justify ...)` token accepts 1 or 2 space-separated alignment keywords:
- Horizontal: `"left"` | `"right"` | `"center"`
- Vertical: `"top"` | `"bottom"` | `"center"`
- Combined: `"left top"` | `"right bottom"` | `"left"` (vertical defaults to center)

Passing `"center"` removes the directive (KiCad default = centered).

### Example Call

```json
{
  "tool": "edit_schematic_component",
  "args": {
    "schematicPath": "/path/to/board.kicad_sch",
    "reference": "R1",
    "fieldPositions": {
      "Reference": { "x": 51.27, "y": 47.46, "justify": "left" },
      "Value":     { "x": 51.27, "y": 52.54, "justify": "left" }
    }
  }
}
```

[...]

PR #226, opened 2026-06-03
feat(schematic): add list_symbol_pins to read pins from symbol libraries

ravishivt:
## Summary
Adds `list_symbol_pins` (and a batch variant `batch_list_symbol_pins`) to read the pin list —
number, name, electrical type, and local position/orientation — of a symbol straight from the
`.kicad_sym` libraries, **without placing it on a schematic**.

## Why
`library_symbol.py` can already search libraries (`search_symbols`), list a library's symbols, and
read symbol properties — but there is no way to see a part's **pins** before you place and wire it.
An agent building a schematic otherwise has to guess pin numbers or place-then-inspect. This closes
that gap and pairs naturally with the existing `get_schematic_pin_locations` (which only works
*after* a symbol is placed).

## Changes
- **`python/commands/symbol_pins.py`** — `SymbolPinCommands.list_symbol_pins`, building on the
  existing `DynamicSymbolLoader.extract_symbol_from_library`; adds the pin parser the loader lacks.
  `batch_list_symbol_pins` resolves many symbols in one call and adds a body bounding box (with an
  optional `compact` mode for symmetric 2-pin passives).
- **`src/tools/library-symbol.ts`** — tool wrappers; **`src/tools/registry.ts`** — new `symbol_pins`
  category.
- **`python/kicad_interface.py`** — register the two handlers.
- **`tests/test_symbol_pins.py`** — unit tests (MagicMock loader; no system KiCad install needed).

## Test plan
- `python -m pytest tests/test_symbol_pins.py -q -o addopts=""` → **12 passed**
- Manual (real libraries): `list_symbol_pins {"symbol":"Device:R"}` → 2 passive pins (±3.81 mm);
  `Connector:Conn_01x04_Pin` → 4 pins; `batch_list_symbol_pins` over `Device:R`/`Device:LED`/
  `Connector:Conn_01x02_Pin` resolves each with a body bounding box.

Python 3.9-compatible; params are camelCase; type hints + docstrings on new signatures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #227, opened 2026-06-03
feat(schematic): reposition and auto-place Ref/Value field labels

ravishivt:
## Summary
Adds field-placement tools: `set_schematic_property_position` and
`batch_set_schematic_property_positions` to move a symbol's Reference/Value labels, and
`autoplace_schematic_fields` to automatically place every symbol's fields clear of its body and of
nearby net labels.

## Why
#196 added control over a field's *justify*, but there is still no way to **move** a field or
auto-place it. Auto-generated schematics routinely stack Ref/Value text on top of the symbol body
or on top of wires/labels — the single biggest readability problem. This makes the output legible
without hand-nudging.

## Changes
- **`python/commands/schematic_field_layout.py`** — field-move + autoplace handlers and the
  geometry helpers they need (upstream `list_schematic_components` doesn't emit field bounding
  boxes).
- **`python/commands/schematic_text_utils.py`** — small standalone module of `.kicad_sch`
  text/S-expression helpers, so this and sibling tools can share them without cross-importing
  handler modules.
- **`src/tools/schematic-layout.ts`** + **`src/tools/registry.ts`** (`schematic_layout` category) +
  **`src/server.ts`** (registration, 2 lines) + **`python/kicad_interface.py`** (dispatch).
- **`tests/test_schematic_field_layout.py`** — incl. end-to-end set/batch on a real `.kicad_sch`.

## Test plan
- `python -m pytest tests/test_schematic_field_layout.py -q -o addopts=""` → **19 passed**
- Manual: place 3 parts, then `set_/batch_set_schematic_property_position(s)` and
  `autoplace_schematic_fields`; re-render (`get_schematic_view`) — no field overlaps the body and
  the file re-parses at every step.

Python 3.9-compatible; params are camelCase; type hints + docstrings; `pathlib`.

### DIFF
--- a/python/commands/schematic_field_layout.py
+++ b/python/commands/schematic_field_layout.py
@@ -1,0 +1,545 @@
+"""
+Schematic field-placement commands.
+
+Tools:
+  - set_schematic_property_position:          move one Reference/Value field
+  - batch_set_schematic_property_positions:   move many fields in one file read/write
+  - autoplace_schematic_fields:               auto-position Ref/Value fields outside the
+                                              component body and any attached net labels
+
+This module is self-contained: it gathers the enriched component/label data it needs
+(body bounding boxes, field positions, pin tips) directly from the schematic + PinLocator,
+rather than depending on the output shape of other handlers.
+"""
+
+import logging
+from pathlib import Path
+from typing import Any, Dict, List
+
+from commands.pin_locator import PinLocator
+from commands.schematic import SchematicManager
+from commands.schematic_text_utils import (
+    _extract_property_position,
+    _extract_property_visible,
+    _find_placed_symbol_block,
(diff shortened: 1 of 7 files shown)


mixelpixx:
Reviewed + rebased onto current main (conflicted with #226/#229's adjacent registrations — kept all entries; resolved the registry.ts category list into three separate objects). Field placement (`set/batch_set_schematic_property_position(s)`, `autoplace_schematic_fields`) is a clean additive feature; new `schematic_field_layout.py` + the shared `schematic_text_utils.py`. `tsc` builds, 19 field-layout tests pass (36 across the schematic features, no regression), full suite at baseline. Merging — #228 will then rebase to drop the duplicate `schematic_text_utils.py`.

PR #254, opened 2026-06-12
refactor: migrate file paths to pathlib in kicad_interface.py and export.py

gcolonese:
## Summary

Brings `python/kicad_interface.py` and `python/commands/export.py` in line with the CONTRIBUTING.md mandate to use `pathlib.Path` for file paths instead of `os.path` (the rule was added for cross-platform/Linux support, and the rest of the newer codebase already follows it). This is the isolated legacy sweep referenced by the export-tools PR.

## Changes

- **`kicad_interface.py`**: 51 `os.path`/`os.makedirs`/`os.listdir` uses -> 0.
- **`export.py`**: 24 -> 0.
- Converted `exists/isfile/isdir`, `dirname/basename/splitext`, `join`, `abspath(expanduser())`, `makedirs`, `listdir`, and `unlink` to their `pathlib.Path` equivalents.
- Behavior preserved: values passed to subprocess arg lists, `pcbnew.SaveBoard`/`PLOT_CONTROLLER`/`SetOutputDirectory`, `open()`, `shutil.copy2`, `sys.path.insert`, and JSON response dicts are kept as `str(...)`, so external API contracts are unchanged. Public return types (e.g. `_current_board_path() -> Optional[str]`) are unchanged.
- Intentionally left as-is (not `os.path`): `os.environ`, `os.getcwd`, `os.sep`, `os.pathsep`, `os.stat`, `os.dup`/`dup2`/`write`.

## Two housekeeping commits riding along

- `chore: add fitz to mypy ignore_missing_imports overrides` -- PyMuPDF has no stubs and fails mypy on any edit to files importing it; same treatment as the existing pcbnew/kipy/PIL entries. (This same one-line override also appears on the export-tools branch; the duplicate is a trivial identical-line merge.)
- `chore: strip pre-existing UTF-8 BOM from export.py` -- `export.py` shipped a UTF-8 BOM on `main`; removed it.

## Testing

- `python -m pytest`: the only failures (13, in `test_swig_dehydration.py` + `test_auto_save_guard.py`) are **pre-existing on `main`** -- verified identical `13 failed, 18 passed` on a clean `main` checkout. They stem from `patch("kicad_interface.pcbnew")` when the SWIG backend isn't loaded, and are unrelated to this change. All other tests pass, including the path-handling-sensitive cairo-preload, logging, ERC, netlist, and create-schematic suites.
- `npx tsc` clean (no TS touched).
- Pre-commit (black/isort/flake8/mypy) passes on every commit; files verified UTF-8/no-BOM.

## Checklist
- [x] Code follows style guidelines (pathlib per CONTRIBUTING)
- [x] No new test failures (pre-existing failures documented)
- [x] No merge conflicts

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,16 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Tooling
 
+- **pathlib migration, first slice**: `kicad_interface.py` and
+  `schematic_handlers.py` now use `pathlib.Path` for file-path handling
+  (`os.path.normcase` remains in `_normalize_board_path` — it has no pathlib
+  equivalent). Values crossing into JSON responses and subprocess argv stay
+  `str`. Also strips a stray UTF-8 BOM from `commands/export.py` and bumps
+  mypy's `python_version` to 3.10 — required by current mypy, which dropped
+  the 3.9 target (note: the project's declared `requires-python = ">=3.9"`
+  floor is therefore no longer verified by the type checker). `export.py`
+  and the remaining `os.path` call sites are follow-up slices.
+
 - **Interface construction smoke test**: a new test constructs
   `KiCADInterface` with the stubbed pcbnew and asserts every
   `command_routes` entry is callable, every schema-listed tool has a route,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ profile = "black"
 line_length = 100
 
 [tool.mypy]
-python_version = "3.9"
+python_version = "3.10"
 warn_return_any = false
 warn_unused_configs = true
 check_untyped_defs = false
--- a/python/commands/export.py
+++ b/python/commands/export.py
@@ -1,4 +1,4 @@
-﻿"""
+"""
 Export command implementations for KiCAD interface
 """
 
--- a/python/commands/schematic_handlers.py
+++ b/python/commands/schematic_handlers.py
@@ -8,10 +8,8 @@ No behavior change — pure relocation.
 """
 
 import base64
-import glob
 import json
 import logging
-import os
 import re
 import shutil
 import subprocess
@@ -25,8 +23,8 @@ import sexpdata
 from commands.library_schematic import LibraryManager as SchematicLibraryManager
 from commands.schematic import SchematicManager
 from commands.wire_manager import WireManager
-from utils.kicad_cli import kicad_cli_not_found_message, resolve_kicad_cli
 from utils.interactive_schematic import reload_kicad_schematic
+from utils.kicad_cli import kicad_cli_not_found_message, resolve_kicad_cli
 from utils.sexpr_format import dumps as kicad_dumps
 
 logger = logging.getLogger("kicad_interface")
@@ -54,7 +52,7 @@ def _svg_to_png(svg_path: str, width: int, height: int) -> Optional[bytes]:
     except Exception:
         pass
 
-    out_path = os.path.join(tempfile.mkdtemp(), "out.png")
+    out_path = Path(tempfile.mkdtemp()) / "out.png"
 
     try:
         r = subprocess.run(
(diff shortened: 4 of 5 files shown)


mixelpixx:
Thanks @gcolonese — this follow-up is welcome. Heads up: #214 just merged, which moved the 43 inline schematic handlers (and `_svg_to_png`) out of kicad_interface.py into `python/commands/schematic_handlers.py` as `SchematicHandlersMixin` (the file went 8,596 -> 5,528 lines). Your pathlib migration will need a rebase onto current main — and any path changes targeting the moved schematic handlers now belong in `commands/schematic_handlers.py`. Happy to review as soon as it's rebased.

mixelpixx:
Status check, since this has been conflicting for three weeks and main has moved a long way since (the June merge wave, and ten more PRs merged today — including #267 and #268, which rewrote substantial parts of both files this PR migrates, kicad_interface.py and export.py).

At this point the branch needs a near-total redo rather than a rebase, so let's decide explicitly rather than let it sit:

1. If you still want to carry it, no rush — but please rebase onto current main and re-scope to whatever os.path usage actually remains. Note the newest utility modules (utils/kicad_cli.py, utils/seven_zip.py) are pathlib-native already, so the migration surface is shrinking on its own.
2. If you don't have the bandwidth (completely understandable for a refactor-only change), say the word or simply leave it — I'll close this in about a week and file a tracking issue for the remaining os.path-to-pathlib migration so the intent isn't lost, crediting this PR as the starting point.

Thanks again for the contribution either way — the direction is right, the codebase just moved underneath it.

gcolonese:
Sorry, didn't see this as I use my work GH acc for this and wasn't in the tail end of last week. I'll rebase this to current main, fix conflicts, then re-merge.

mixelpixx:
No problem, and thanks for checking in -- that resets the clock. A heads-up for the rebase: main has moved substantially since this branch was cut (the v2.3.0 and v2.3.1 releases landed in between), and kicad_interface.py specifically has grown several new handlers (_handle_save_project, close_project routing, freerouting net-class wiring), so expect real conflicts there rather than trivial ones. If it turns out to be more archaeology than it is worth, an equally good outcome is re-cutting the pathlib migration fresh on current main file-by-file -- the mechanical nature of the change means the old branch has little unique value to preserve.

mixelpixx:
Verified and merging — and an apology is owed: you re-cut this cleanly on 07-06, an hour after saying you would, and it sat for six days because our queue survey misclassified it as "awaiting author." The miss was ours, not yours.

What was verified on this branch merged with today's main (which absorbed five merges this session):

- The migration is disciplined about the one thing that matters in a str-to-Path conversion: every value that crosses a boundary — JSON responses, subprocess argv, shutil calls — comes back through `str(...)`, so no `Path` leaks into a serialization or string-consumer context. Spot-checked the subtler equivalences too: `os.path.dirname("foo") == ""` vs `Path("foo").parent == Path(".")` both end at `"."`; `resolve()` vs `abspath()` in `_normalize_board_path` is consistent because both comparands go through the same function.
- Full suite: only the 8 known environmental failures (1381 passing). black and mypy clean.
- The mypy `python_version` bump to 3.10 turns out to be forced, not optional: current mypy rejects a 3.9 target outright ("must be 3.10 or higher"), so main's setting was already being discarded with a warning. Your commit message said "match runtime" — the stronger justification is that the old value no longer functions. One consequence worth recording (added to the CHANGELOG entry on the branch): the project's declared `requires-python = ">=3.9"` floor is no longer verifiable by the type checker. Whether to raise the floor to 3.10 is a separate decision — macOS KiCad builds still bundle Python 3.9, so it is not automatic.
- Scope accounting for the follow-up slices: `schematic_handlers.py` is now os.path-free, `kicad_interface.py` keeps exactly one intentional call (`os.path.normcase`, no pathlib equivalent), and `export.py` — despite the PR title carried over from the original branch — still has ~22 os.path call sites; this slice only strips its stray UTF-8 BOM. That is the right call for a re-cut (small, reviewable, lands), just noting it so the remaining migration surface is written down.

One commit was added on the branch before merge: the CHANGELOG entry (house rule: every merge gets one) documenting the slice boundaries and the mypy note above. Thanks for coming back to this after the codebase moved underneath it twice — the file-by-file re-cut was exactly the right response.

Issue #220, opened 2026-06-03
`create_project` writes a non-conformant `.kicad_pro` stub (missing entire project structure)


ptr727:
**Environment**

- KiCad **10.0** (Windows 11, x64), running with IPC API server enabled.
- KiCAD-MCP-Server: docs/CHANGELOG report **v2.2.3** (note: `package.json` shows
  `2.1.0-alpha` — see KNOWN_ISSUES.md #5).
- Backend: `get_backend_state` → `{backend:"ipc", realtime_sync:true, ipcConnected:true}`.
- Claude instructed to document any issues it encountered while creating the project.

## 1. `create_project` writes a non-conformant `.kicad_pro` stub (missing entire project structure)

**Severity:** High — the created project carries none of the KiCad 10 project
settings, so net classes, design settings, BOM/field templates, text variables,
and ERC/DRC config are all absent.

**Steps:**

1. `create_project({path, name:"EspDinIoT"})` → `{success:true, _backend:"swig"}`.
2. Inspect the resulting `EspDinIoT.kicad_pro`.

**Actual:** 122 bytes, only:

```json
{
  "board": { "filename": "EspDinIoT.kicad_pcb" },
  "sheets": [ ["root", "EspDinIoT.kicad_sch"] ]
}
```

**Expected:** a complete KiCad 10 `.kicad_pro` (`meta.version 3`, `net_settings`,
`board.design_settings`, `schematic`, `text_variables`, `erc`, etc.) — i.e. what
KiCad writes for a new project. As-is, opening in KiCad 10 silently regenerates
defaults and discards any intended configuration. The `sheets` entry also uses
the literal id `"root"` rather than the schematic root-sheet UUID.

**Suggested fix:** emit a full KiCad 10 project template (current schema), or
drive `kicad-cli`/IPC to create the project so KiCad writes the file.

mixelpixx:
Triaged — confirmed. `create_project` (project.py ~111-119) writes a hand-rolled ~122-byte `.kicad_pro` stub with only `board`+`sheets` and a literal "root" sheet id, missing the KiCad 10 project schema (meta/version, net_settings, board.design_settings, schematic, text_variables, erc/drc), so KiCad regenerates defaults and discards intended config. Plan: write a conformant KiCad 10 `.kicad_pro` (proper schema + UUID root id), ideally from a template. Scheduled with the #221 template work.

PR #255, opened 2026-06-12
fix(backend): pin SWIG/IPC backend per loaded project (#223)

### LINKED ISSUE #223: Backend selected inconsistently per call (SWIG vs IPC) on a single loaded project
**Environment**

## 4. Backend selected inconsistently per call (SWIG vs IPC) on a single loaded project

**Severity:** Medium — root cause of #1/#3 symptoms.

Within one session: `get_backend_state` reports `ipc` connected, but
`create_project`, `open_project`, and `add_layer` all run on `swig`
(`_realtime:false`), while `save_project` runs on `ipc`. The SWIG-created/opened
project is never loaded into the live IPC GUI instance, and the IPC save path
cannot see SWIG-side edits, so changes are lost.

**Suggested:** route a project's lifecycle (create/open/edit/save) through one
backend, or document which operations are SWIG-only and require an explicit
re-open in the GUI.

mixelpixx:
Fixes #223.

## Problem
On a single loaded project, the backend was chosen per call. `create_project`/`open_project` always ran on SWIG (they're not IPC-capable), but any later IPC-capable command — including `save_project` — silently upgraded the session to IPC when KiCad was reachable. The IPC save then wrote the live GUI's board, which never saw the SWIG-side edits: lost work, and the driver behind the #220/#222 symptoms ptr727 reported.

## Fix: session-pinned backend
- **Pin at load**: `open_project` pins the session to `ipc` only when the live GUI provably has the *same* `.kicad_pcb` open (new `IPCBackend.get_open_board_path()` + normalized path comparison). Otherwise — including every `create_project`, since the GUI can't have a brand-new board open — the session pins to `swig`.
- **No silent upgrades**: a swig-pinned session never routes board commands over IPC, even with IPC connected. The GUI's in-memory board may be stale relative to our edits, so an IPC save could clobber them. Responses carry a `_backend_note` explaining the pin and how to adopt the GUI (reopen the project while it's open in KiCad).
- **Graceful downgrade**: an ipc-pinned session falls back to swig if the connection drops (GUI closed), reloading the board from its last on-disk state.
- **Observability**: `get_backend_state` now returns `sessionBackend`/`sessionBoardPath`, and its `backend` field reflects the pin while a project is loaded — previously it reported `ipc` connectivity while every project command actually ran on SWIG, which is exactly the confusion in the issue report.
- Runtime reconnect with **no project loaded** is unchanged.

## Testing
- 14 new tests in `tests/test_backend_session_pinning.py`, including the literal issue repro: open via SWIG, then `save_project` with IPC connected must stay SWIG and never touch `_ipc_save_project`.
- Existing `test_backend_metadata.py` invariants all still pass (17/17).
- Full suite at baseline (8 failed / 1062 passed — the failures are the pre-existing Java-not-installed freerouting set), `npm run build` clean.
- Docs: session-pinning section added to `docs/IPC_BACKEND_STATUS.md`, CHANGELOG entry under Unreleased.

## Verification ask
The live-GUI path can't be fully exercised headless (CI covers SWIG only). @ptr727 — when you get a chance, could you re-run your repro from #223 on this branch (Win11 / KiCad 10 with IPC enabled)? Expected: `get_backend_state` shows `sessionBackend` after open, every project command reports the same `_backend`, and edits survive save.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,18 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **Backend is now pinned per loaded project (SWIG vs IPC)** (#223): commands
+  on a single loaded project previously ran on whichever backend happened to
+  be reachable per call — `create_project`/`open_project`/`add_layer` on SWIG
+  while `save_project` silently upgraded to IPC, saving the live GUI's (stale)
+  board and losing the SWIG-side edits. Now `open_project` pins the session to
+  IPC only when the GUI provably has the same `.kicad_pcb` open; otherwise the
+  whole lifecycle (including `save_project`) stays on SWIG, with a
+  `_backend_note` on responses explaining why IPC wasn't used. IPC-pinned
+  sessions fall back to SWIG (reloading from disk) if the GUI connection
+  drops. `get_backend_state` gains `sessionBackend`/`sessionBoardPath`, and
+  its `backend` field reflects the session pin while a project is loaded.
+
 - **`rotate_component` now treats `angle` as an absolute target rotation**,
   matching its schema description. Previously the IPC backend added the
   supplied angle to the current rotation, so two consecutive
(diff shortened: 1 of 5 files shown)


mixelpixx:
Self-review pass (independent reviewer agent) caught three issues, now fixed in the second commit: stale pin survival after an unrecoverable SWIG-dehydration reopen, silent board-reload failure in the IPC->SWIG downgrade path, and the explicit `ipc_save_board` bypassing the pin without warning. 16 pinning tests now; full suite still at baseline (8 failed / 1063 passed).

PR #259, opened 2026-06-25
fix: correct pin positions and outward angles for rotated+mirrored symbols

uberjay:
## Summary

Two related fixes for symbol→world pin geometry on symbols that are **both rotated and mirrored**, where the schematic tools reported pins — and the wire-stub direction used to attach net labels — on the wrong side. Both bugs were masked by the same gap: the tests exercised only single-axis placements and a vertical-pin `Device:R`, so neither the rotated+mirrored case nor horizontal pins were ever covered.

The two commits are ordered — the angle fix builds directly on the position fix.

---

## 1. `pin_world_xy` — mirror applied before rotation

`WireDragger.pin_world_xy` — the symbol→world pin transform behind `get_schematic_pin_locations`, `get_net_at_point`, and net-label pin snapping — applied the `(mirror x|y)` reflection in **library space, before the rotation**. eeschema reflects the *placed* (already-rotated) symbol, i.e. **after** the rotation.

The two orders agree for rotations of 0° and 180°, but **diverge for 90°/270° combined with any mirror**: the pins come out swapped. Polarized parts are where it bites — a diode at `rotation 90 + mirror x` reports its cathode (pin 1) and anode (pin 2) at each other's positions. Downstream that:

- snaps polarity-bearing net labels onto the wrong pins, and
- makes `get_schematic_pin_locations` / `get_net_at_point` report the wrong pin→net for any symbol that is **both rotated and mirrored**.

### Fix
- **`WireDragger.pin_world_xy`** — reorder to `Y-flip → rotate → mirror` (mirror in screen space, after the rotation). Single-axis behavior is byte-identical; only the rotated-*and*-mirrored case changes.
- **`WireManager._collect_pin_positions`** — was a second, divergent copy of the transform (mirror-before-rotate *and* an inverted rotation sign). Now routes through `pin_world_xy`, so there's one source of truth.
- **New test** `test_diode_polarity_rotation_and_mirror_through_eeschema`, parametrized over rotation × mirror, using `kicad-cli sch export netlist` as the oracle (a snapped `_K` label must bind to pin 1). The existing tests covered rotation-only and mirror-only but never the combination — which is exactly why this slipped through.

### Verification
Against `kicad-cli sch export netlist` for all 12 (rotation × mirror) combinations, the snapped K/A labels bind to the correct pins. Before the fix the four `{90,270} × {x,y}` cases failed (pins swapped); after, all pass. Existing single-axis pure-math assertions and the resistor set-membership junction tests are unaffected.

---

## 2. `get_pin_angle` — outward bearing wrong for horizontal pins

`pin_locator.get_pin_angle` returns the *outward* bearing of a pin endpoint — the direction a wire stub leaves the pin, away from the body — consumed by `connection_schematic.connect_to_net` to place a stub + net label (`stub_end = pin + L·(cos θ, −sin θ)`, screen Y down).

It derived that bearing by **negating the library pin angle** (a Y-flip). That only yields the outward direction for **vertical** pins (270↔90); **horizontal** pins (0→0, 180→180) came out pointing *into* the body — 180° wrong. So `connect_to_net` placed the stub/label on the wrong side of any horizontal-pin part, and the `angle` field of `get_schematic_pin_locations` was likewise inverted.

The old `test_get_pin_angle.py` used only a vertical-pin `Device:R` **and** an oracle that re-derived the same flawed math, so it was blind to this — the same fixture gap that hid the position bug.

### Fix
- **`get_pin_angle`** now derives the bearing from the (commit-1-fixed) `pin_world_xy`: transform the pin endpoint and a bodyward point, take `atan2` of the difference. Correct by construction for both pin orientations and every rotation/mirror — no hand-rolled angle transform.
- **Rewrote `test_get_pin_angle.py`** to add horizontal pins (`Device:D`) and three independent oracles: a behavioural check (the placed stub must extend *away* from the body), hand-computed bearings for canonical placements incl. the real-world D10 freewheel diode (`rot 90 + mirror x` → up), and exhaustive agreement with `pin_world_xy`.

### Verification
101/101 in the rewritten matrix with the fix; reverting only `get_pin_angle` fails 51, all of them `Device:D` (horizontal) cases. The `connect_to_net` stub+label+netlist integration tests (`test_label_at_pin_net_connections`, `test_net_connectivity`) pass unchanged — the corrected angles don't alter connectivity.

### DIFF
--- a/python/commands/pin_locator.py
+++ b/python/commands/pin_locator.py
@@ -274,7 +274,7 @@ class PinLocator:
             if transform is None:
                 return None
 
-            _, _, symbol_rotation, mirror_x, mirror_y, lib_id = transform
+            sym_x, sym_y, symbol_rotation, mirror_x, mirror_y, lib_id = transform
             if not lib_id:
                 return None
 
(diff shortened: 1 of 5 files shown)


uberjay:
As a side note -- this bug _really_ messed with me and resulted in a multi-hour back and forth with Claude. Because the AI's tools rendered something different from what KiCad showed me, I thought I was losing my mind. I was fairly sure I knew which pin of the diode was pointing up, but the LLM kept telling me that I was just mis-reading the symbol and it was pointing down. 🤪

Mehanik:
@uberjay Logic's correct and well-tested, netlist-oracle tests are spot on. Nice work!

One thing before merge: please run `black` — the two new test files need to be formatted according to the project code style.

Also worth doing `pre-commit install` so this gets caught automatically. Thanks!

uberjay:
Done, reformatted!

mixelpixx:
Merged. For the record, verification here went beyond the pure-math suite: I staged the Device library so the eeschema-truth tests would run on Windows against a real KiCad 10.0.0 kicad-cli, and all 17 oracle tests passed, including the full 12-combination rotation-by-mirror matrix. So the fix is now netlist-verified on both your platform and a stock Windows KiCad 10 install. The rewritten get_pin_angle matrix (101 cases) and the full regression suite were green as well. Thanks for the careful before/after failure documentation — it made this easy to trust.

One note for anyone reading later: the eeschema-truth tests gate on the Linux path /usr/share/kicad/symbols/Device.kicad_sym, and KiCad 10 on Windows ships sharded .kicad_symdir directories instead, so these tests skip on a stock Windows checkout. Making that gate platform-aware would let them run everywhere KiCad is installed.

PR #261, opened 2026-06-26
feat: add suggest_placement (PCB) and suggest_schematic_declutter (schematic) layout tools

jflaflamme:
## Summary

Adds two new AI-driven layout tools to the MCP server. Both are **additive** (no existing behavior changed), require **no new dependencies**, and are **dry-run by default** (never mutate a file unless `apply: true`).

- **`suggest_placement`** (PCB) — connectivity-driven footprint placement optimizer.
- **`suggest_schematic_declutter`** (schematic) — re-orients overlapping net/global labels for readability.

## `suggest_placement` (PCB)

Turns the netlist into an optimized placement — the heuristic an engineer does against the ratsnest, automated:

- **Connectivity clustering** — force-directed solve (rest-length springs) pulls connected parts together, so an IC's feedback divider and decoupling caps end up hugging it.
- **Power-aware weighting** — high-current nets (VBAT/VBUS/…) and IC↔decoupling links pull tightest, keeping those paths short and direct.
- **Pin-facing rotation** — rotates each part 0/90/180/270 to minimize wire length to its neighbours, removing airwire crossings.
- **Density spreading** — diffuses parts across free board area so a whole-board run stays legal instead of over-packing into a blob.
- **Row/column alignment** — snaps near-collinear part centers onto shared row/column lines, so the result reads as tidy rows/columns with items centered (like Align Centers + Distribute).
- **Scoped mode** — `refs` + `bounds` regroup one IC's passives at a time within a box; unlisted parts stay as fixed anchors.
- Returns `{ref:[x,y,rot]}` proposals + an HPWL/overlap score. Deterministic (dry-run matches the applied run). Reuses `ComponentCommands._footprint_courtyard_bbox`, so collision matches `check_courtyard_overlaps`.

## `suggest_schematic_declutter` (schematic)

Re-orients overlapping net/global labels so their text lands in free space:

- **Connectivity-safe** — a label's `(at x,y)` is its electrical connection point, so the anchor is held fixed; only orientation (0/90/180/270) and justification change. Nets are never altered.
- Returns proposals `[{name, at, from_angle→to_angle}]` + an overlap score. `apply: true` rewrites the orientations.
- Phase 1 = labels; symbol spreading + wire reroute is a deliberately separate future phase.

## Integration

- `ComponentCommands` gains the optimizer via a new `PlacementOptimizerCommands` mixin; declutter reuses `schematic_field_layout`'s bbox model. Only additions to the dispatch map + TS tool registrations.
- New tools advertised in `tool_schemas.py` (placement) and `src/tools/schematic-layout.ts` (declutter).

## Testing

- Unit tests cover all pure geometry/solver logic on mocked pcbnew + synthetic schematic: clustering, power weighting, pin-facing rotation, density spreading, row/column alignment, legalization, label re-orientation, and the regex apply path. `tsc` builds clean.

## Caveats / not-yet-done

- **Validated via unit tests; not yet validated against a live board/schematic.** Reviewers should smoke-test on a real design.
- Placement HPWL is a wire-length proxy (pad/center based), not a route guarantee; legalizer is O(n²) (fine to a few hundred parts).
- Declutter Phase 1 doesn't yet move free text or slide labels along their wire.

### DIFF
--- a/python/commands/component.py
+++ b/python/commands/component.py
@@ -10,11 +10,12 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import pcbnew
 from commands.library import LibraryManager
+from commands.placement_optimizer import PlacementOptimizerCommands
 
 logger = logging.getLogger("kicad_interface")
 
 
-class ComponentCommands:
+class ComponentCommands(PlacementOptimizerCommands):
     """Handles component-related KiCAD operations"""
 
     def __init__(
--- a/python/commands/placement_optimizer.py
+++ b/python/commands/placement_optimizer.py
@@ -1,0 +1,833 @@
+"""Connectivity-driven component placement optimizer.
+
+DRAFT — proposed new MCP capability.
+
+The existing tools (place_component, move_component, align_components,
+check_courtyard_overlaps) give you placement primitives plus *validation*, but
+nothing turns the netlist into a suggested layout. This module adds a single
+command, ``suggest_placement``, that does the heuristic an engineer does by
+eye against the ratsnest:
+
+    1. Build a component-level connection graph from pad nets.
+    2. Force-directed relaxation (Fruchterman-Reingold flavoured, adapted for
+       PCB): nets pull connected parts together, overlapping courtyards push
+       apart, the board outline contains everything. Power/high-current nets
+       and IC<->decoupling links are weighted so those parts hug tightest.
+    3. Pin-facing rotation: each movable part is rotated (0/90/180/270) to the
+       orientation that minimises the pad-level wire length to its neighbours —
+       this is what removes airwire crossings.
+    4. Snap to grid and return *proposed* {ref:[x,y,rot]} positions. It does
+       NOT modify the board unless ``apply=true``.
+
+It reuses ComponentCommands._footprint_courtyard_bbox so the collision model is
+identical to check_courtyard_overlaps (AABB on the real courtyard polygon).
+Intended loop:
+
+    suggest_placement(dryRun) -> check_courtyard_overlaps(positions=...) ->
+    apply -> autoroute(best-of-N) -> run_drc
+
+Quality metric reported is pad-level HPWL (half-perimeter wire length), the
+standard placement proxy for routed length — lower is better.
+"""
+
+from __future__ import annotations
+
+import logging
+import math
+from typing import Any, Dict, List, Optional, Tuple
+
+import pcbnew
+
+logger = logging.getLogger("kicad_interface")
+
+# Net-name fragments treated as high-current / power and pulled short & direct
+# (step 4 of the placement priority order). Case-insensitive substring match.
+DEFAULT_POWER_NETS = [
+    "VBAT",
+    "VBUS",
+    "VCC",
+    "VDD",
+    "VIN",
+    "VOUT",
+    "+5V",
+    "+3V3",
+    "+3.3V",
+    "3V3",
+    "5V",
+    "12V",
+    "PWR",
+    "BAT",
+    "VSYS",
+    "PVDD",
+]
+
+
+class PlacementOptimizerCommands:
+    """Mixin providing suggest_placement.
+
+    Expects the host (ComponentCommands) to provide:
+      - self.board                      : the loaded BOARD
+      - self._footprint_courtyard_bbox  : bbox in mm for a footprint
+    so it composes with the existing component handlers.
+    """
+
+    # ------------------------------------------------------------------ #
+    #  Public command                                                    #
+    # ------------------------------------------------------------------ #
+    def suggest_placement(self, params: Dict[str, Any]) -> Dict[str, Any]:
+        """Propose an optimized placement that shortens net length, orients
+        parts toward their partners, and removes courtyard overlaps — without
+        committing unless asked.
+
+        Args:
+            refs: References to move. Default: every non-locked footprint.
+            locked: References to hold fixed (connectors, mounting-constrained,
+                RF, edge parts). They still pull movable parts. KiCad's own
+                IsLocked() footprints are auto-added here.
+            iterations: Relaxation passes (default 200).
+            grid_mm: Snap proposed positions to this grid (default 0.5).
+            margin_mm: Extra keepout between courtyards (default 0.3).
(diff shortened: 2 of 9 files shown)


mixelpixx (changes requested):
Thanks for this — it is a genuinely useful pair of tools, and the design is careful in the ways that matter most. I reviewed both new modules in full and ran the suite locally. The headline claims all hold up: both tools are dry-run by default (every disk/board write is gated behind `apply: true`), there are no new dependencies, the changes are additive, and `PlacementOptimizerCommands` mixes into `ComponentCommands` without shadowing any existing method. `suggest_placement` is also deterministic, which is exactly right for a tool whose dry-run preview must match its apply result.

There are two things I need fixed before this can merge, plus a couple of follow-ups.

## Must fix

### 1. New `unit` enums omit `mil`, which fails an existing test
`python/schemas/tool_schemas.py:957` and `:976` (the two `unit` parameters in the `suggest_placement` schema) declare:

```json
"enum": ["mm", "inch"]
```

Every other unit-bearing tool in the repo uses `["mm", "mil", "inch"]`, and `tests/test_mil_unit_support.py::test_python_tool_schema_unit_enums_include_mil` walks `TOOL_SCHEMAS` and asserts that every `unit` enum includes `mil`. With this PR the test fails:

```
AssertionError: Unit enum ['mm', 'inch'] is missing 'mil'
```

Please change both enums to `["mm", "mil", "inch"]` and make sure the handler's unit conversion accepts `mil` (1 mil = 0.0254 mm) so the schema and behavior agree. CI did not run on the branch, so this would otherwise land as a red test on `main`.

### 2. `suggest_schematic_declutter` has no Python schema entry
The TypeScript definition in `src/tools/schematic-layout.ts` is complete, but there is no matching entry in `python/schemas/tool_schemas.py` — `suggest_schematic_declutter` appears in none of the `*_TOOLS` lists, so it never makes it into `TOOL_SCHEMAS`. The effect is that the Python server's own `tools/list` falls through to the no-schema branch and logs `No schema or annotation for tool` at runtime. `suggest_placement` is wired up correctly (it is in `COMPONENT_TOOLS`); please add the declutter schema the same way (e.g. to `SCHEMATIC_TOOLS`) so both surfaces agree. You can port it directly from the TS definition.

## Should fix

### 3. Apply mode does not add a `(justify)` token when one is absent
In `schematic_declutter.py`, `_set_justify_near` only updates justification when the label block already contains a `(justify ...)` token before `(uuid ...)`. KiCad omits `(justify)` for default left-justified labels, so for those labels a 180/270 reorientation updates the angle but leaves justification unchanged — the text then renders pointing the wrong way. The file is not corrupted and connectivity is preserved (the anchor x,y is never moved), so this is a correctness gap in apply mode rather than a safety issue. Either inject a `(justify ...)` token when none exists, or document the limitation in the tool description.

## Follow-ups (not blocking)
- No tests accompany roughly 1100 new lines. The pure functions here are very testable — a smoke test of `plan_label_declutter` / `_label_bbox` and the no-board / fewer-than-two-movable paths of `suggest_placement` would lock in the behavior. Worth adding before this grows.
- `black` would reformat `placement_optimizer.py` and `schematic_declutter.py` (a few lines exceed the configured length). Running `black python/` will clear it.
- Minor: `placement_optimizer.py` uses `logging.getLogger("KiCADInterface")` while the rest of the package uses `"kicad_interface"`. Worth aligning so the log streams stay together.

Happy to re-review as soon as the two must-fix items are in. Thanks again — the dry-run-by-default discipline and the determinism work made this much easier to trust.

jflaflamme:
Thanks for the thorough review — all must-fix and should-fix items are addressed, plus the test follow-up. Force-pushed as a single squashed commit.

**Must fix**
1. **`mil` unit enum** — both `suggest_placement` unit params (`bounds`, `board_outline`) now use `["mm", "mil", "inch"]` in the Python schema *and* the TS registration, and the handler's unit conversion accepts `mil` (1 mil = 0.0254 mm). `test_python_tool_schema_unit_enums_include_mil` passes.
2. **Missing Python schema** — `suggest_schematic_declutter` is now in `SCHEMATIC_TOOLS`, so it lands in `TOOL_SCHEMAS` (no more no-schema fallthrough). Ported from the TS definition.

**Should fix**
3. **Justify injection** — `_set_justify_near` now injects a `(justify …)` token as the last child of the `(effects …)` group (paren-matched, quote-aware) when KiCad omitted it for a default-left label, so a 180/270 reorientation actually re-points the text. Covered by a test.

**Follow-ups**
- **Tests** — added `tests/test_suggest_placement.py` (no-board / fewer-than-two-movable paths, optimize, scoped `bounds`, determinism, alignment cluster-snap, mil/inch conversion) and `tests/test_schematic_declutter.py` (bbox, planner, regex apply incl. justify injection). 22 new assertions, all green.
- **black** — ran `black python/`; the two modules are clean.
- **Logger name** — aligned to `kicad_interface`.

Still flagging honestly: validated via unit tests, **not yet on a live board/schematic** — happy to add a confirmation once I've smoke-tested on a real design. Thanks again!

mixelpixx (approved):
Re-reviewed the force-pushed branch; every item from the previous review is verified in the code, not just claimed:

- Both unit enums are now [mm, mil, inch] in the Python schema and the TS registration, the handler converts mil at 0.0254 mm, and test_mil_unit_support passes against the branch.
- suggest_schematic_declutter is registered: TOOL_SCHEMAS now contains both new tools (verified by importing the module and checking the built dict), so the Python tools/list no longer falls through to the no-schema branch.
- _set_justify_near now injects a (justify ...) token into the (effects ...) group when KiCad omitted it, with test coverage, so 180/270 reorientations re-point the text.
- 22 new tests pass locally; the full suite shows only the 8 known environmental failures (Freerouting needs Java, absent on this runner); mypy and black are clean on the new modules; the TS build and vitest suite pass.

On the honest flag about not yet running this on a live board: dry-run-by-default means the worst case for a bad proposal is a rejected suggestion, not a mutated file, and apply mode never writes the board to disk (save_project stays a separate explicit step). That risk profile is acceptable to merge now — if you do the live smoke test later and anything looks off, please open an issue and we will treat it as a bug.

Thanks for the fast, complete turnaround.

PR #263, opened 2026-06-28
feat: MCP tools for footprint/component 3D models

karu2003:
## Summary
Adds three MCP tools so the full 3D-model workflow lives in the server:

- **`import_3d_model`** — copy a `.step/.stp/.wrl/.x3d/.iges` file into the project's `*.3dshapes` library and return a portable `${KIPRJMOD}/...` path.
- **`add_footprint_3d_model`** — attach or replace a `(model ...)` block in a library `.kicad_mod` file (idempotent, supports `${KIPRJMOD}` and offset/scale/rotate).
- **`add_component_3d_model`** — attach a model to one or more *placed* footprints on the open board via the **IPC backend** (live, instant UI update). Forces `scale = 1,1,1` so the body actually renders (a fresh `Footprint3DModel` proto defaults to `scale = 0`).

Wiring:
- Python: `FootprintCreator.add_3d_model` / `import_3d_model`, `IPCBoardAPI.add_3d_model`, command handlers + routing in `kicad_interface.py`, and `add_component_3d_model` registered in `IPC_CAPABLE_COMMANDS`.
- TypeScript: tool schemas in `src/tools/footprint.ts`.

Single commit, `+503 / -0` across 4 files, based directly on `main`.

## Test plan
- [x] `npm run build` (tsc) passes.
- [x] `add_footprint_3d_model` adds exactly one `(model ...)` block and is idempotent with `replace=true`.
- [x] `import_3d_model` copies into `*.3dshapes` and returns the `${KIPRJMOD}` path.
- [x] `add_component_3d_model` over IPC sets `scale=1` and bodies render (verified via `kicad-cli pcb render`).
- [x] Reviewer: confirm tool descriptions/schemas read well in your MCP client.

### DIFF
--- a/python/commands/footprint.py
+++ b/python/commands/footprint.py
@@ -11,6 +11,7 @@ KiCAD 9 .kicad_mod format reference:
 import logging
 import os
 import re
+import shutil
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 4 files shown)


mixelpixx (changes requested):
This is a well-scoped feature and the workflow split (import into `3dshapes` / edit a library `.kicad_mod` / attach to placed footprints live) maps cleanly onto how people actually work with 3D models. Forcing `scale = 1,1,1` on the fresh proto is exactly the kind of sharp edge worth capturing in the server. Verified locally on top of current `main`: merges cleanly, `tsc` passes, `mypy` clean, backend suites pass. One must-fix before merge, plus two smaller asks.

## Must fix: `_handle_add_component_3d_model` bypasses the session pin

The handler routes to IPC on its own check:

```python
if self.use_ipc and getattr(self, "ipc_board_api", None):
    return self._ipc_add_component_3d_model(params)
```

This skips `_session_allows_ipc()`. Since #255, a loaded project is pinned to one backend for its whole lifecycle: when `open_project` ran on SWIG (because the GUI didn't have this same board open), IPC-capable commands must not run over IPC even though it is connected — the GUI may hold a different or stale board, and an IPC edit would mutate *that* board, not the one the session loaded. That divergent-state failure is exactly what issue #223 was, and this handler reintroduces it for `add_component_3d_model`: the dispatcher correctly refuses the IPC route on a SWIG-pinned session and falls through to this "SWIG" handler, which then calls the IPC path directly anyway.

Fix is one condition:

```python
if self.use_ipc and getattr(self, "ipc_board_api", None) and self._session_allows_ipc():
    return self._ipc_add_component_3d_model(params)
```

With a SWIG-pinned session the command then returns your existing clear "requires the live IPC backend" message, which is the correct behavior — the user reopens the project while the GUI has it open, gets an IPC pin, and the live path works.

## Should fix

1. **`black` formatting** — `black --check python/kicad_api/ipc_backend.py` fails on this branch. `black python/` will clear it (the other two Python files are already clean).
2. **Tests** — the PR is +503 with none. Most of it needs a live KiCad and can't run in CI, but `FootprintCreator._find_model_blocks` / `add_3d_model` are pure text manipulation on `.kicad_mod` content and are very testable: add-to-file-without-model, replace-same-filename with `replace=true`, the `replace=false` no-op, and the malformed-no-closing-paren error. That locks in the idempotency contract your test plan verified manually.

## Nit
`_handle_add_footprint_3d_model` and `_handle_import_3d_model` return `{"success": False, "error": ...}` on exception; the surrounding handlers use `message` (plus optional `errorDetails`). Aligning keeps client-side error handling uniform.

Happy to merge as soon as the pin check and black are in — the test ask can land here or as a fast follow-up, your call.

mixelpixx (approved):
Re-verified against the review: the _session_allows_ipc() check is in place in _handle_add_component_3d_model, so a SWIG-pinned session now gets your clear requires-live-IPC message instead of silently editing whatever board the GUI holds — the #223 hazard class is closed. black is clean on all three Python files, and the error shape was aligned to message as well. Verified locally on top of current main: mypy clean, tsc and vitest green, backend-pinning suite passes, full run at the known-environmental baseline. The .kicad_mod text-manipulation tests remain welcome as a follow-up whenever convenient, as discussed. Merging.

PR #264, opened 2026-06-29
fix: ModuleNotFoundError: No module named 'skip' (first crash); sqlit…

fage2022:
…e3.OperationalError: unable to open database file (second crash)

There were two issues that caused the server to crash on startup:

### 1. ModuleNotFoundError: No module named 'skip' (first crash) The Python files in the project ( schematic.py , library_schematic.py , etc.) import from a module called skip :

```
from skip import Schematic
```
This skip module is provided by the kicad-skip package, listed in python/requirements.txt . But the package was never installed into the Python interpreter the server actually uses — the KiCAD 10.0 bundled Python at C:\Program Files\KiCad\10.0\bin\python.exe . Installing it resolved this:

```
pip install kicad-skip
```
### 2. sqlite3.OperationalError: unable to open database file (second crash) After fixing the skip import, the server then crashed because JLCPCBPartsManager.__init__() tried to create a SQLite database at C:\Users\admin\.kicad-mcp\data\jlcpcb_parts.db , but the sandboxed environment ( trae-sandbox ) blocks writes to that directory path.

The fix was making JLCPCBPartsManager fail gracefully — instead of throwing an unhandled exception that kills the Python process, it now catches PermissionError / OSError (directory creation) and sqlite3.OperationalError (database open), logs a warning, and continues with the database disabled . All query methods check self.conn is None and return empty results.

The server now starts successfully — the JLCPCB parts database is simply unavailable in this environment, but all other functionality works.

### DIFF
--- a/python/_fix_perms.py
+++ b/python/_fix_perms.py
@@ -1,0 +1,31 @@
+import os, sqlite3, sys
+
+data_dir = os.path.join(os.path.expanduser("~"), ".kicad-mcp", "data")
+print(f"Data dir: {data_dir}")
+
+# Create the directory
+os.makedirs(data_dir, exist_ok=True)
+print(f"Directory exists: {os.path.isdir(data_dir)}")
+
+# Try creating the db
+db_path = os.path.join(data_dir, "jlcpcb_parts.db")
+print(f"DB path: {db_path}")
+
+try:
+    conn = sqlite3.connect(db_path)
+    print("SQLite connect SUCCESS")
+    conn.close()
+except Exception as e:
+    print(f"SQLite connect FAILED: {e}")
+    # Try with a simpler path
+    try:
+        alt_path = os.path.join(os.environ.get("TEMP", "C:\\temp"), "jlcpcb_parts.db")
+        print(f"Trying alternate path: {alt_path}")
+        conn = sqlite3.connect(alt_path)
+        print("Alt path SUCCESS")
+        conn.close()
+    except Exception as e2:
+        print(f"Alt path FAILED: {e2}")
+    sys.exit(1)
+
+print("All good!")
--- a/python/commands/jlcpcb_parts.py
+++ b/python/commands/jlcpcb_parts.py
@@ -34,14 +34,34 @@ class JLCPCBPartsManager:
                 user data directory, e.g. ~/.local/share/kicad-mcp/jlcpcb_parts.db
                 on Linux). See PlatformHelper.get_data_dir() for platform paths.
         """
+        self.conn: Optional[sqlite3.Connection] = None
+        self.db_path = db_path or ""
+
         if db_path is None:
             data_dir = PlatformHelper.get_data_dir()
-            data_dir.mkdir(parents=True, exist_ok=True)
-            db_path = str(data_dir / "jlcpcb_parts.db")
+            try:
+                data_dir.mkdir(parents=True, exist_ok=True)
+            except (PermissionError, OSError) as exc:
+                logger.warning(
+                    "Cannot create JLCPCB data directory %s: %s. "
+                    "JLCPCB parts database will be disabled.",
+                    data_dir,
+                    exc,
+                )
+                return  # self.conn stays None — disabled
 
-        self.db_path = db_path
-        self.conn: Optional[sqlite3.Connection] = None
-        self._init_database()
+            db_path = str(data_dir / "jlcpcb_parts.db")
+            self.db_path = db_path
+
+        try:
+            self._init_database()
+        except sqlite3.OperationalError as exc:
+            logger.warning(
+                "Cannot open JLCPCB parts database at %s: %s. "
+                "JLCPCB parts database will be disabled.",
+                self.db_path,
+                exc,
+            )
 
     def _init_database(self) -> None:
         """Initialize SQLite database with schema"""
@@ -102,6 +122,9 @@ class JLCPCBPartsManager:
             parts: List of part dicts from JLCPCB API
             progress_callback: Optional callback(current, total, message)
         """
+        if self.conn is None:
+            logger.warning("JLCPCB database not available — skipping import_parts")
+            return
         cursor = self.conn.cursor()
         imported = 0
         skipped = 0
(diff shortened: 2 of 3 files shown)


mixelpixx (changes requested):
Thanks for the detailed failure report — the second half of this PR fixes a real and nasty class of bug: a sandboxed or locked-down environment turning an optional feature (the JLCPCB parts cache) into a fatal startup crash. The graceful-degradation work in `jlcpcb_parts.py` is exactly right in shape: catch the directory-creation and database-open failures, log a warning, run with the database disabled, and guard every query method on `self.conn is None`. I want that part merged.

The PR can't go in as-is, though, for three reasons:

## 1. `python/_fix_perms.py` must be removed
This is a local debugging scratch script (prints, `sys.exit(1)`, "All good!") that was committed by accident. It isn't imported by anything and doesn't belong in the package. Please drop it from the branch.

## 2. The `sys.path` insertion in `kicad_interface.py` needs to be dropped or justified
```python
# 获取当前python文件夹绝对路径
BASE_DIR = os.path.dirname(os.path.abspath(__file__))
# 把根目录放入sys.path最前面
sys.path.insert(0, BASE_DIR)
```
Two problems. First, the mechanics: prepending the package directory to `sys.path[0]` makes every top-level module in `python/` (`commands`, `utils`, `resources`, `schemas`, `annotations`) shadow any identically-named module anywhere else in the interpreter, for every downstream import. That is a global side effect with real collision potential, applied to fix what appears to be a launch-environment issue specific to your sandbox — the TypeScript server already launches the Python backend with the correct working directory, which is why nobody else hits this. If your environment genuinely needs a path fix, `PYTHONPATH` in the launch configuration is the non-invasive place for it. Second, if some variant of this does turn out to be necessary, the comments need to be in English to match the rest of the codebase.

Note the first crash in your report (`ModuleNotFoundError: No module named 'skip'`) was resolved by `pip install kicad-skip` into KiCad's bundled Python — an environment fix, not a code fix — so nothing in the diff is actually needed for that half. A docs note in the Windows troubleshooting guide about installing requirements into the *bundled* interpreter would be a welcome addition if you want to capture it.

## 3. Keep and finish the good part
For the `jlcpcb_parts.py` changes:
- Run `black python/` — at least the `get_database_stats` disabled-path line exceeds the line-length limit.
- Add a small test: construct `JLCPCBPartsManager` with `db_path` pointing into a location where opening fails (or mock `mkdir`/`sqlite3.connect` to raise), then assert the manager reports disabled, `search_parts` returns `[]`, and `get_part_info` returns `None`. That locks in the never-crash contract this PR exists to establish.
- Have the disabled `get_database_stats` return the same key set as the live one so callers don't need two shapes (it looks close already — just worth an explicit check).

Trimmed to the `jlcpcb_parts.py` hardening plus a test, this merges quickly.

mixelpixx:
Checking in on this one — it has been quiet since the 07-03 review and the useful half is still sitting unmerged, which is a shame because it fixes a real bug.

To restate the split, because the verdict was mixed and that is easy to lose track of:

**The sqlite hardening is good and we want it.** Your changes in `python/commands/jlcpcb_parts.py` are the valuable part: `JLCPCBPartsManager.__init__` currently lets a `PermissionError` or `sqlite3.OperationalError` escape and take down server startup, so a sandboxed or read-only data directory bricks the whole MCP server over an optional parts database. Degrading to `self.conn = None` with a warning, plus the `if self.conn is None` guards in `import_parts` / `import_jlcsearch_parts`, is exactly the right shape. That half stands on its own.

**Three things still block it, all unchanged since the review:**

1. **`python/_fix_perms.py` and the `sys.path.insert` hack need to come out.** In `python/kicad_interface.py` the branch still adds:

   ```python
   BASE_DIR = os.path.dirname(os.path.abspath(__file__))
   sys.path.insert(0, BASE_DIR)
   ```

   The mechanism to be aware of: mutating `sys.path` at import time makes module resolution depend on which file happened to be imported first, which is how you get bugs that appear only under pytest, or only when the server is launched from a different working directory. The `ModuleNotFoundError: No module named skip` you originally hit is real, but it is an environment/packaging problem, and this fix trades a visible startup error for an invisible import-order one. If you can reproduce the original failure, please open an issue describing how you launch the server and we will fix it at the packaging layer instead.

2. **`black` does not pass.** From the repo root:

   ```
   would reformat python\commands\jlcpcb_parts.py
   would reformat python\kicad_interface.py
   2 files would be reformatted, 1 file would be left unchanged.
   ```

   `python -m black python/commands/jlcpcb_parts.py python/kicad_interface.py` from the root (so it picks up `line-length = 100` from `pyproject.toml`) clears it.

3. **A test for the degraded path.** Something that constructs `JLCPCBPartsManager` against an unwritable directory and asserts the object is usable with `conn is None` rather than raising. Without it, a future refactor silently restores the crash — and since this bug only shows up in restricted environments, CI is the only place it will ever get caught.

If you drop commit content for items 1 and address 2 and 3, this is a small, clearly-scoped fix and it will move quickly. If the `sys.path` line is load-bearing for how you personally run the server, say so and we will work out the packaging fix separately rather than leaving you without a workaround.

Note it merges cleanly into current `main` despite being 168 commits behind, so no rebase is required.

mixelpixx:
Ping — no activity since 2026-06-29; the 07-03 review is still outstanding.

Recap: the `skip` ModuleNotFoundError fix and the JLCPCB graceful degradation are both wanted. What blocked it was the two extras bundled alongside — the `_fix_perms.py` debug script and the `sys.path` hack. Dropping those two leaves a small, clean PR I will merge.

Two updates since that review:

1. **CI now actually runs the Python suite** (#334). It was a no-op in four ways — every gate ended in `|| echo`, `pytest python/` pointed at the source tree so it collected 7 tests instead of ~1550, and Actions was disabled repo-wide. So when you push, you will get real black/mypy/pytest results for the first time. Not something you did wrong; it just means the bar is now enforced rather than implied.
2. Main has moved substantially (v2.4.0 plus about eight merges today), so rebase first.

If you would rather not continue, no problem — say so and I will close it, or cherry-pick just the `skip` import fix with attribution if it is still needed on current main. Let me know either way.

mixelpixx:

mixelpixx:
Closing in favor of #385, which lands the valuable core of this PR with credit (Co-authored-by): the graceful degradation of JLCPCBPartsManager when its data directory is unwritable or its database unopenable — construction happens at import, so that failure took down the whole server, exactly as you reported. Per the earlier review, two hunks are deliberately not taken: _fix_perms.py (the server should not chmod its way past permission problems) and the sys.path.insert change (it cannot resolve a missing site-packages package — your 'No module named skip' crash is an interpreter-mismatch issue at install time, now covered by setup verification per #350/#370). Thank you for the report and the fix.

Issue #256, opened 2026-06-15
Errors with writing in KICAD 10


AKCore:
1. Pathological formatting — it wrote the schematic with a single 110,828-character line. 
   
  2. Broken instance data any components it added had (project "project")
  and (path "/") instead of (project "ProjectName") and the real root-sheet path (path "/07577d12-…"). KiCad couldn't
  resolve those symbols to the current sheet, so as soon as you edited/redrew, it crashed.

scorp508:
@AKCore I opened a PR last night which may help this 'project' issue and other KiCAD 10 items. I was noticing any attempt to reposition a component added via MCP would trigger a KiCAD crash.

https://github.com/mixelpixx/KiCAD-MCP-Server/pull/265

Integer-Conversion-Error:
have the same issue, it was working so well then it nuked the whole project, everything unrecoverable. turns out it was churning out the whole sch file as a 12,000 character S-expression single line

edit(typo): "10,000 line lines" --> "12,000 character single line"

mixelpixx:
Status update: #265 is now merged to `main`, and it directly addresses the second problem reported here — the broken instance data. Components placed via `add_schematic_component` / `batch_add_components` now get the real project name (derived from the `.kicad_pro`), the real root-sheet uuid path instead of `(path "/")`, per-pin `(pin "N" (uuid ...))` entries, and the full KiCad 10 field set in canonical order. The fix was verified by round-tripping the generated block through `kicad-cli sch upgrade` and asserting the structure is identical to what KiCad itself writes, on both KiCad 10.0.4 (author) and 10.0.0 (my review). That is the mechanism that made KiCad crash when you edited or dragged a placed symbol, so if you can, please retest against current `main` and report back.

To set expectations clearly: the **first** problem in this report — the pathological single-line output (the 110,828-character line, and the "10,000-line S-expression lines" @Integer-Conversion-Error hit) — is **not** fixed by #265. That comes from a different code path (the sexpdata-based writers used by some wire/label operations serialize the whole file without line breaks), and it needs its own fix. Leaving this issue open to track that half. If you hit the single-line problem again, noting which tool call produced it (wire, label, move, etc.) would help pin down which writer is responsible.

mixelpixx:
Second update: with #284 merged just now, both problems in this report are fixed on current main.

[...]

PR #265, opened 2026-06-30
fix: emit structurally-complete symbol instances (no crash on drag)

scorp508:
## Problem
Placing a symbol via `add_schematic_component` (or `batch_add_components`) wrote a
symbol-instance block that KiCad loads and renders fine but **crashes when the placed
symbol is dragged** in the editor. Both tools route through
`DynamicSymbolLoader.create_component_instance`, whose instance block used placeholder
values and omitted fields that real KiCad output includes:

- literal `(project "project")` instead of the real project name
- hardcoded `(path "/")` instead of the real sheet-UUID path
- omitted `body_style` / `exclude_from_sim` / `in_pos_files`
- omitted per-pin `(pin "N" (uuid ...))` entries
- v9-style hide markers inside `(effects ...)` and missing `(show_name no)` /
  `(do_not_autoplace no)`

## Fix
Reproduce KiCad 10 (schematic format `20260306`, generator `eeschema`) authored output:

- **Real project name** derived from the nearest `.kicad_pro` stem
- **Real instance path**:
  - flat / root schematic: `/<root-sheet-uuid>` (the schematic's own top-level uuid)
  - hierarchical child sheet: `/<root-uuid>/<sheet-block-uuid>...`, reconstructed by
    walking the root project's sheet tree
- **Full KiCad-10 header field set** in canonical order
  (lib_id, at, [mirror], unit, body_style, exclude_from_sim, in_bom, on_board,
  in_pos_files, dnp, uuid, properties, pins, instances)
- **Per-pin `(pin "N" (uuid ...))`** for the placed unit (plus shared unit-0 pins),
  string-sorted to match KiCad's ordering
- **v10 property attributes** (`show_name no` / `do_not_autoplace no`) with top-level
  `(hide yes)` instead of `hide` inside `(effects ...)`

Because both `add_schematic_component` and `batch_add_components` use the same writer,
the single change fixes both.

## Verification (KiCad 10.0.4)
Ground truth was captured by upgrading a bundled demo (`interf_u`) with
`kicad-cli sch upgrade`.

1. `kicad-cli sch erc <file>` parses the generated schematic with no structural/parse
   error.
2. Round-trip: `kicad-cli sch upgrade` rewrites the file to canonical form and the
   placed symbol is **structurally identical** (s-expression tree equal modulo fresh
   uuids) — i.e. KiCad does not rewrite the instance.
3. Multi-unit pin selection (placed unit + shared unit-0 pins, string-sorted) and the
   hierarchical-path chain are unit-tested.
4. Both `add_schematic_component` and `batch_add_components` covered (shared writer).

> Note: the drag-crash itself requires the KiCad GUI to confirm interactively; the
> `kicad-cli` round-trip equivalence to KiCad-authored output is the structural proxy.

## Tests
- New `tests/test_symbol_instance_completeness.py` (header field set, v10 properties,
  real project name + path, per-pin uuids, multi-unit selection, hierarchical path, and
  a `kicad-cli` ERC + round-trip integration test that skips when the CLI is absent).
- Full affected suite green: 132 passed. Pre-commit hooks (black, isort, flake8, mypy)
  pass.

## PR Checklist
- [x] All tests pass locally
- [x] No merge conflicts

### DIFF
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -529,6 +529,191 @@ class DynamicSymbolLoader:
         ry = -dx * math.sin(rad) + dy * math.cos(rad)
         return round(rx, 3), round(ry, 3)
 
+    # ------------------------------------------------------------------ #
+    # Instance-block helpers (project name, hierarchical path, pin uuids) #
+    # ------------------------------------------------------------------ #
+
+    @staticmethod
+    def _read_root_uuid(content: str) -> str:
+        """Return a schematic's own top-level (uuid ...) value, or '' if absent."""
+        m = re.search(r'\(uuid\s+"?([0-9a-fA-F-]+)"?\)', content)
+        return m.group(1) if m else ""
+
+    def _resolve_project_name(self, schematic_path: Path) -> str:
+        """Return the KiCad project name recorded in (instances (project "<name>" ...)).
+
+        KiCad uses the project (``.kicad_pro``) stem, not the literal string
+        ``"project"``. Derive it from the nearest ``.kicad_pro`` (searching the
+        schematic's directory then a few parents), falling back to the schematic's
+        own stem when no project file is found.
+        """
+        try:
+            sch = Path(schematic_path).resolve()
+            search_dirs = [sch.parent] + list(sch.parent.parents)[:3]
+            for directory in search_dirs:
+                pros = sorted(directory.glob("*.kicad_pro"))
+                if pros:
+                    return pros[0].stem
+            return sch.stem
+        except Exception:
+            return Path(schematic_path).stem
+
+    def _iter_child_sheets(self, content: str) -> List[Tuple[str, str]]:
+        """Yield (sheet_block_uuid, sheet_file_rel) for each (sheet ...) in a schematic.
+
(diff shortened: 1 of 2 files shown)


mixelpixx (approved):
Reviewed in depth — this is exactly the fix #256's second complaint needed, and the verification methodology (order-sensitive s-expression round-trip through kicad-cli sch upgrade) is the right bar for code that writes into users' schematics. Verified locally on top of current main: your 11 new tests pass with the round-trip running against a real KiCad 10.0.0 kicad-cli (not skipped), the full suite is green apart from the 8 known Java-environmental failures, no existing test depends on the old writer output, and all 38 tests that exercise the writer pass. The fallback design matters most here and it is right: every failure path (no .kicad_pro, no root uuid, sheet-walk failure) degrades to the pre-PR behavior instead of writing something new and broken.

Merging. Follow-ups noted, none blocking:
1. When multiple .kicad_pro files sit in one directory, _resolve_project_name takes the alphabetically first; preferring a stem that matches the schematic/root stem would be more robust.
2. The kicad-cli round-trip covers the flat/unit-1/unrotated case; extending it to a rotated+mirrored multi-unit hierarchical placement would lock in the harder paths.
3. The other half of #256 (pathological single-line/110k-char output from the sexpdata-based writers) is explicitly out of scope here and stays tracked on that issue.

PR #266, opened 2026-06-30
fix: make the yaqwsx JLCPCB download work on a stock install (volume auto-detect + robust 7-Zip resolution)

scorp508:
Makes the **yaqwsx JLCPCB download path** (`download_jlcpcb_database source=yaqwsx`)
actually work on a stock install. Two related fixes in the same flow:

---

## 1. Auto-detect the yaqwsx volume count (was a hardcoded cap of 30)
The yaqwsx source is a split 7-Zip archive (`cache.zip` + `cache.z01 … cache.zNN`) that
needs **every** volume to extract. The loop capped volumes at `YAQWSX_MAX_VOLUMES = 30`,
but the live archive has more (and the count changes as the catalog is rebuilt). Over the
cap → incomplete archive → `RuntimeError("7z extraction failed")` and no `cache.sqlite3`.

**Fix:** iterate with `itertools.count(1)` and rely on the existing first-404 break to stop
at the true last volume. `YAQWSX_MAX_VOLUMES` becomes a high runaway-loop **safety guard**
(999), not the expected count — auto-detected, never truncated again.

## 2. Resolve 7-Zip robustly (was PATH-only)
The extraction step needs a 7z CLI. On Windows 7-Zip installs to
`C:\Program Files\7-Zip\7z.exe` but is **not** added to `PATH`, so `_find_7z()` (PATH-only)
failed even with 7-Zip 26.01 installed:
`RuntimeError: yaqwsx fallback requires a 7z CLI (7z/7zz/7za) which was not found`.
Same class of bug as the kicad-cli resolver (#267) — same pattern applied.

**Fix:** a shared resolver (`utils/seven_zip.py`) wired into `_find_7z()`:
`$SEVEN_ZIP`/`$SEVENZIP_PATH` override → `PATH` (7z/7zz/7za/7zr) → known per-OS install dirs
(Windows `%ProgramW6432%`/`%ProgramFiles%`/`(x86)\7-Zip` + literals; macOS Homebrew/usr-local;
Linux usr/bin, usr-local, snap/flatpak). Install-dir hits return an **absolute path**; the
result is cached; the env override is re-checked each call. When nothing resolves,
`download_yaqwsx` raises a **clear, actionable** message listing every location tried plus the
install / PATH / `SEVEN_ZIP` options — not the bare "not found".

---

## Verification
**Volume auto-detect**
- Unit tests: fetches all 40 simulated volumes and stops at the `z41` 404 (past the old
  30/80 caps); a re-run with all volumes present re-downloads nothing.
- Live probe confirms the 404 break finds the true last volume (validates auto-detection).

**7-Zip resolution (Windows, 7-Zip 26.01 at `C:\Program Files\7-Zip`, NOT on PATH)**
1. Auto-detect resolves `…\7-Zip\7z.exe` (previously raised "not found"). ✅
2. 7-Zip on PATH → still works (PATH checked before install dirs). ✅
3. Bogus `$SEVEN_ZIP` → `None` + the clear multi-location error (no silent fallback). ✅
4. No 7-Zip anywhere → the new error naming install / PATH / env options. ✅
5. Resolved binary runs: `7z` prints its `7-Zip 26.01 (x64)` banner. ✅

> Note: the full ~421 MB download + extraction + ~7M-part build still requires network +
> a 7z CLI; the volume-fetch and 7-Zip-resolution fixes (the two bugs) are covered by the
> unit tests, the live volume probe, and the real-binary checks above. Conversion
> (`convert_source_sqlite`) is covered by existing tests.

New tests: `tests/test_seven_zip_resolver.py` + 7-Zip cases in `tests/test_jlcpcb_downloader.py`.
Full suite: **1093 passed, 12 skipped.** Pre-commit hooks (black, isort, flake8, mypy) pass.

## PR Checklist
- [x] All tests pass locally
- [x] No merge conflicts

### DIFF
--- a/python/commands/jlcpcb_downloader.py
+++ b/python/commands/jlcpcb_downloader.py
@@ -22,6 +22,7 @@ Data is MIT-licensed via CDFER and yaqwsx/jlcparts; the underlying catalog facts
 originate from JLCPCB / LCSC and are subject to their terms.
 """
 
+import itertools
 import json
 import logging
 import os
@@ -37,7 +38,11 @@ logger = logging.getLogger("kicad_interface")
 
 CDFER_SQLITE_URL = "https://cdfer.github.io/jlcpcb-parts-database/jlcpcb-components.sqlite3"
 YAQWSX_BASE_URL = "https://yaqwsx.github.io/jlcparts/data"
-YAQWSX_MAX_VOLUMES = 30
+# Safety upper bound only — NOT the expected volume count. The yaqwsx split archive
+# (cache.z01 … cache.zNN) grows as the JLCPCB catalog expands, so the real last volume
+# is auto-detected by the 404 break in the download loop. This cap merely prevents an
+# unbounded loop if the 404 never arrives.
+YAQWSX_MAX_VOLUMES = 999
 
 ProgressFn = Optional[Callable[[str], None]]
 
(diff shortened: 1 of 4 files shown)


mixelpixx (approved):
Verified locally: all 25 downloader/resolver tests pass, black and mypy clean, full suite matches the known-environmental baseline. The design mirrors the kicad-cli resolver that just landed in #267 — override fails loudly rather than substituting a different binary, install-dir hits return absolute paths, cache with per-call env recheck — which is exactly the consistency I want across these external-tool resolvers. Demoting YAQWSX_MAX_VOLUMES from expected-count to runaway guard with the 404 break as the real terminator is the right inversion; the live archive's volume count changing over time made any hardcoded cap a time bomb. Merging.

PR #267, opened 2026-06-30
fix: resolve kicad-cli robustly instead of failing when not on PATH

scorp508:
## Problem
KiCad's Windows installer does **not** add `KiCad\<ver>\bin` to `PATH`. Every MCP tool
that shells out to `kicad-cli` then fails with a misleading **"kicad-cli not found in
PATH"** even though `kicad-cli.exe` is present in the install's `bin` directory. Observed
this session: `generate_netlist` returned `"Failed to generate netlist: kicad-cli not
found in PATH"`. The same failure hits the whole cli-backed family: `export_netlist`,
`export_gerbers`, `export_drill`, `export_pos`, `export_*` (pdf/svg/dxf/3d/bom/...),
`run_erc`/`run_drc`, the board 2D view, and schematic SVG/PDF export.

## Root cause
Resolution was bare-name/`PATH` only (`shutil.which("kicad-cli")` or spawning
`"kicad-cli"`), with several duplicated, out-of-date `_find_kicad_cli` helpers and no
fallback to the actual KiCad install location.

## Fix
Add one centralized resolver (`python/utils/kicad_cli.py`) and wire it into every
cli-backed tool. Resolution order:

1. **`$KICAD_CLI` / `$KICAD_CLI_PATH` override** — authoritative; an override that is set
   but invalid fails loudly rather than silently selecting a different binary.
2. **Adjacent to the running interpreter** — `Path(sys.executable).parent / kicad-cli`.
   The backend runs under KiCad's bundled Python (`KiCad\<ver>\bin\python`, which is how
   it imports `pcbnew`); `kicad-cli` lives in that same `bin` dir, making this the most
   reliable source for this server.
3. **`PATH`**.
4. **Known per-OS install locations** (newest KiCad version first; Windows
   `Program Files[ (x86)]\KiCad\*\bin`, macOS app bundle + Homebrew, Linux + Flatpak).

The resolved path is **cached** (the env override is re-checked every call so a changed
value isn't masked). When resolution fails, tools return a **clear, actionable error**
listing every location tried and how to set `KICAD_CLI`, replacing the bare
"not found in PATH".

This centralizes the duplicated `_find_kicad_cli` helpers (`kicad_interface`,
`design_rules`, `export`) onto the resolver, replaces the direct `"kicad-cli"` spawns in
`schematic_handlers` and `board/view`, and swaps all 47 bare error strings for the
detailed message. (The Node/TS layer does not spawn `kicad-cli` directly — it spawns this
Python backend — so no TS change is needed.)

## Verification (KiCad 10.0.4, with KiCad's bin NOT on PATH)
1. `export_netlist` against a real `.kicad_sch` **succeeds** (previously errored). ✅
2. Resolver finds `kicad-cli` via interpreter-adjacent / known-locations **without PATH**. ✅
3. Bogus `$KICAD_CLI` → tool fails with the **new clear error** (locations tried + env
   hint), and does not fall back to another binary. ✅
4. `kicad-cli` on `PATH` → still works (no regression). ✅
5. Resolved cli version (`10.0.4`) **matches the running KiCad**. ✅
- New `tests/test_kicad_cli_resolver.py` (14 tests, incl. an end-to-end `export_netlist`
  with PATH stripped of KiCad, and the bogus-override clear-error case).
- **Full suite: 1089 passed, 12 skipped.** Pre-commit hooks (black, isort, flake8, mypy)
  pass.

> Note: a one-line `board: Any` annotation was added to `SchematicHandlersMixin` so mypy
> can type-check that (edited) file standalone — a pre-existing per-file-mypy limitation
> surfaced because this PR touches the mixin.

## PR Checklist
- [x] All tests pass locally
- [x] No merge conflicts

### DIFF
--- a/python/commands/board/view.py
+++ b/python/commands/board/view.py
@@ -10,6 +10,7 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import pcbnew
 from PIL import Image
+from utils.kicad_cli import kicad_cli_not_found_message, resolve_kicad_cli
 
 logger = logging.getLogger("kicad_interface")
 
@@ -138,7 +139,6 @@ class BoardViewCommands:
         - "file": image is written next to the .kicad_pcb file and ``filePath`` is returned.
         """
         import glob
-        import shutil
         import subprocess
         import tempfile
 
@@ -173,12 +173,11 @@ class BoardViewCommands:
             layers: List[str] = params.get("layers", [])
             response_mode = params.get("responseMode", "inline")
 
-            kicad_cli = shutil.which("kicad-cli") or shutil.which("kicad-cli.exe")
+            kicad_cli = resolve_kicad_cli()
             if not kicad_cli:
                 return {
                     "success": False,
-                    "message": "kicad-cli not found in PATH",
-                    "errorDetails": "Install KiCad and ensure kicad-cli is on PATH",
+                    "message": kicad_cli_not_found_message(),
                 }
 
             with tempfile.TemporaryDirectory() as tmpdir:
@@ -287,8 +286,7 @@ class BoardViewCommands:
         except FileNotFoundError:
             return {
                 "success": False,
-                "message": "kicad-cli not found in PATH",
-                "errorDetails": "Install KiCad and ensure kicad-cli is on PATH",
+                "message": kicad_cli_not_found_message(),
             }
         except Exception as e:
             logger.error(f"Error getting board 2D view: {e}")
--- a/python/commands/design_rules.py
+++ b/python/commands/design_rules.py
@@ -7,6 +7,7 @@ import os
 from typing import Any, Dict, List, Optional, Tuple
 
 import pcbnew
+from utils.kicad_cli import kicad_cli_not_found_message, resolve_kicad_cli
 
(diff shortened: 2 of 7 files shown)


mixelpixx (approved):
Verified on a Windows machine that reproduces the exact problem (KiCad 10 installed, bin not on PATH): the resolver's tiers behave as designed, the override-fails-loudly semantics are right (silently substituting a different binary than the one the user named would be worse than failing), and the actionable error listing every location tried is a big step up from the bare 'not found in PATH' repeated 47 times. Centralizing the three drifted _find_kicad_cli copies is the part that keeps this fixed. All 12 resolver tests pass here, black and mypy clean, full suite matches the known-environmental baseline.

One real-world gap, not blocking because it is shared with the rest of the repo: this machine's KiCad lives at C:\KiCad\10.0\bin (the installer lets you pick any root, and short no-space roots are a common choice), which tier 4 misses since it only scans Program Files. platform_helper.py and library.py make the same assumption, so I'm filing a separate issue to unify KiCad-root discovery (add C:\KiCad-style roots and/or the registry uninstall keys) across all three rather than asking this PR to patch one of them. Merging.

PR #268, opened 2026-06-30
fix: fall back to SWIG/file backend for reads when IPC has no open document

scorp508:
## Problem (observed this session)
Some file-answerable read tools routed to the realtime **IPC** backend even when the
live KiCad GUI had **no document open**, producing confusing or wrong results — while
`open_project` had succeeded on the SWIG/file backend:

- `get_component_properties` returned `"Component J1 not found"` (the part exists in the
  file; IPC just had nothing loaded) — conflating *no board open* with *absent*.
- `get_board_info` returned `success:true` with **zeroed counts AND an embedded `error`
  string** — a contract violation (a failure must be `success:false`).

## Fix
Use the existing backend infrastructure (session pinning, `_backend`/`_realtime` tags)
for graceful fallback:

1. **Detect the no-open-document condition** in the IPC read handlers.
   `_ipc_get_board_info` now treats a zeroed size carrying an `error` as a failure (no
   more false success); `_ipc_get_component_properties` distinguishes *no board open*
   from a genuine miss. Both signal it via a `_no_open_document` marker.
2. **Reactive SWIG/file fallback** in `handle_command`: when a file-answerable read
   (`get_board_info`, `get_component_properties`) returns `_no_open_document`, it is
   re-served from the on-disk `.kicad_pcb` via SWIG and tagged `_backend:"swig"` with a
   `_backend_note` explaining the redirect.
3. **Accurate structured errors** when neither backend can answer (`success:false` with a
   distinct message) — never `success:true` with zeroed data + an error.

When a board **is** open in the live KiCad, the IPC realtime path is unchanged
(`_backend:"ipc"`), and a genuinely-absent component still reports "not found" via IPC.
The fallback set is extensible — other reads opt in by having their `_ipc_*` handler emit
`_no_open_document`.

## Verification (tests/test_ipc_swig_read_fallback.py)
1. No live document → `get_component_properties` succeeds via SWIG fallback. ✅
2. No live document → `get_board_info` file-fallback, or a clean `success:false`; never a
   false-success zeroed payload. ✅
3. Board open in live KiCad → IPC realtime path preserved (`_backend:"ipc"`). ✅
4. Genuinely-absent component → "not found", distinct from the no-board case. ✅
5. Every response reports which backend answered (`_backend`). ✅
- Full suite: **1084 passed, 12 skipped.** Pre-commit hooks (black, isort, flake8, mypy)
  pass.

## Parallel-review / conflict notes
- This PR and #267 both edit `kicad_interface.py` but in **non-overlapping regions**; a
  local trial merge of both branches is **conflict-free** (the only shared hunk is an
  identical isort import reorder, which merges cleanly).
- The pre-existing mypy `has-type` on `SchematicHandlersMixin.self.board` (surfaced here
  via import-following) is suppressed with a **scoped `pyproject.toml` mypy override**
  rather than editing `schematic_handlers.py`, keeping this PR's file set disjoint from
  #267.

## PR Checklist
- [x] All tests pass locally
- [x] No merge conflicts

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,3 +41,11 @@ module = [
     "PIL.*",
 ]
 ignore_missing_imports = true
+
+# SchematicHandlersMixin relies on attributes (e.g. self.board) provided by the host
+# KiCADInterface it is mixed into, so checking it standalone — or as a followed import
+# from another module — cannot determine their types. Suppress the resulting has-type
+# noise here so unrelated modules that merely import it are not blocked by it.
+[[tool.mypy.overrides]]
+module = ["commands.schematic_handlers"]
+disable_error_code = ["has-type"]
--- a/python/kicad_interface.py
+++ b/python/kicad_interface.py
@@ -36,8 +36,8 @@ if sys.platform == "win32":
 
 import sexpdata
 from annotations import AnnotationLoader
-from commands.wire_manager import WireManager
 from commands.schematic_handlers import SchematicHandlersMixin
+from commands.wire_manager import WireManager
 from resources.resource_definitions import RESOURCE_DEFINITIONS, handle_resource_read
 
 # Import tool schemas, resource definitions, and IPC API annotations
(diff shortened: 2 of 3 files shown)


mixelpixx (approved):
Reviewed against the #255 session-pinning logic specifically, since this touches the same dispatch path. The interaction is correct: the fallback only fires inside the branch already gated by _session_allows_ipc(), so a SWIG-pinned session never routes these reads to IPC in the first place, and an IPC-pinned session whose GUI closes its document gets a read-only re-serve from the file — no pin mutation, no save, no divergent-state risk. The _no_open_document marker being confined to reads (FILE_ANSWERABLE_READS) is the right scope constraint, and fixing the success:true-with-embedded-error contract violation in _ipc_get_board_info is worth having on its own.

Also appreciated: the mypy override for the mixin's has-type noise clears the two long-standing errors, so mypy python/ is now fully clean on main.

One nit for a follow-up, not blocking: both _ipc_* handlers now tag every exception with _no_open_document, so an IPC failure that is not actually a no-document condition still gets the 'live KiCad reported no open document' note. For reads the fallback behavior is right regardless; only the note text can mislead. Verified locally: your 9 new tests plus the 32 backend pinning/metadata tests pass, full suite matches the known-environmental baseline. Merging.

Issue #243, opened 2026-06-11
create_schematic: injects hidden _TEMPLATE_R/C/D instances at off-screen coordinates


Dewieinns:
## Environment
- KiCad 10.0.1, Windows 11
- MCP server: main post-PR-#212

## Severity
**Medium** — contaminates every new schematic; causes ERC duplicates and BOM pollution.

## Symptom
Every schematic created via `create_schematic` contains three off-screen component instances:

```
(property "Reference" "_TEMPLATE_R" (at -100 -102.54 0))
(property "Reference" "_TEMPLATE_C" (at -100 -107.46 0))
(property "Reference" "_TEMPLATE_D" (at -100 -117.46 0))
```

They are not visible at default zoom but appear in `list_schematic_components`, ERC results, and the BOM.

## Root cause
`commands/schematic.py` copies `python/templates/template_with_symbols.kicad_sch` as the base for every new schematic. That template file hard-codes `_TEMPLATE_R`, `_TEMPLATE_C`, and `_TEMPLATE_D` placed at `(-100, ~-102 to -117)` — far off the visible sheet area.

These instances exist because `dynamic_symbol_loader.py` → `load_symbol_dynamically()` creates a template instance at `(-200, -200)` to pre-seed `lib_symbols`. That helper was called during an earlier bootstrap phase and its output was baked into the template file.

## Expected behaviour
A newly created schematic should be empty. If pre-seeding `lib_symbols` is desired, the library definitions should be embedded in the `(lib_symbols ...)` block only — **not** as placed instances.

## Fix
Replace `template_with_symbols.kicad_sch` with a truly blank schematic (like `templates/empty.kicad_sch` or `templates/minimal.kicad_sch`) as the copy source, OR strip any instance whose reference matches `_TEMPLATE_*` immediately after the template copy.

## Workaround
After calling `create_schematic`, call `delete_schematic_component` for `_TEMPLATE_R`, `_TEMPLATE_C`, and `_TEMPLATE_D` before doing any real work.

Dewieinns:
Related: #221 reports the same root cause via `create_project` — the template file is pre-seeded with symbol instances that bleed into every new schematic/project. Fixing the template (or stripping `_TEMPLATE_*` instances after copy) would resolve both.

PR #269, opened 2026-06-30
fix: write blank KiCad 10 schematic on create_project (#221)

### LINKED ISSUE #221: `create_project` writes an outdated `.kicad_sch` (KiCad 9 format) pre-seeded with a `Device:R` symbol
**Environment**

## 2. `create_project` writes an outdated `.kicad_sch` (KiCad 9 format) pre-seeded with a `Device:R` symbol

**Severity:** Medium.

**Actual:** `EspDinIoT.kicad_sch` header is
`(kicad_sch (version 20250114) (generator "KiCAD-MCP-Server") ...)` and the file
(~22 KB) already contains a `Device:R` library symbol in `lib_symbols`.

**Expected (KiCad 10):** a blank schematic at the current format version
(eeschema writes `(version 20260306) (generator "eeschema")
(generator_version "10.0")`) with an empty `lib_symbols` for a new project.

**Note:** KNOWN_ISSUES.md ("Template File Corruption — Fixed v2.2.2") says the
20250114 version was *intentionally restored* — that is the KiCad **9** format.
Under KiCad **10** it is stale, and seeding a `Device:R` into a blank project is
surprising.

**Contrast (works correctly):** the sibling `.kicad_pcb` from the same
`create_project` call IS current and real:
`(version 20260206) (generator "pcbnew") (generator_version "10.0")`. So PCB
creation uses real pcbnew, while project/schematic creation use hand-written
templates — hence the version skew between the two files in one project.

boovaragan:
## Summary

Fixes #221.

`create_project` produced a "new" project that was neither blank nor in the
current KiCad format. Two defects, both fixed here:

1. **Pre-seeded components.** `create_project` copied
   `template_with_symbols_expanded.kicad_sch`, which contains **placed**
   `_TEMPLATE_R`, `_TEMPLATE_C`, and `_TEMPLATE_D` symbol instances. Every new
   project therefore opened with three stray components on the canvas.
   `create_project` now copies `empty.kicad_sch`, which has **no placed
   component instances**.

2. **Outdated schematic format.** Both the template and the inline fallback
   hardcoded the KiCad 9 era format `(version 20250114)`. Bumped to
   `20260306` (KiCad 10.0.4) so the generated schematic opens cleanly in
   KiCad 10 without a format-upgrade prompt.

### Note on `lib_symbols`

`empty.kicad_sch` retains the `Device:R/C/LED` definitions in its
`lib_symbols` block. These are **library definitions, not placed components** —
they render nothing, appear in no BOM/ERC, and KiCad prunes unused entries on
the first save. They are also the shared geometry source for ~20 schematic test
fixtures, so the file is deliberately left as the "empty project" (no placed
parts) rather than a literally empty file.

## Out of scope for this PR

- The non-conformant `.kicad_pro` stub (#220) — separate root cause, tracked
  independently.
- `create_schematic`'s use of the symbol-seeded template (#243).

## Test plan

- [x] All schematic test fixtures that consume `empty.kicad_sch` pass under
      `--timeout=60 --timeout-method=thread` (bbox/pin/angle set matches clean
      `main` at 46 passed; 126 more across component/label/snap files; net/wire
      /delete group green).
- [x] Baseline diff (clean `main` worktree vs. branch) confirms the change
      introduces no new failures.
- [ ] CI passes on the PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,15 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **`create_project` now writes a blank KiCad 10 schematic** (issue #221).
+  Previously it copied `template_with_symbols_expanded.kicad_sch`, which
+  contains placed `_TEMPLATE_R/C/D` symbol instances, so every "new" project
+  opened with three stray components on the canvas. It also hardcoded the
+  KiCad 9 era schematic format `(version 20250114)`. `create_project` now
+  copies `empty.kicad_sch` (no placed components) and the schematic format
+  version is bumped to `20260306` (KiCad 10.0.4) in both the template and the
+  inline fallback.
+
 - **`rotate_component` now treats `angle` as an absolute target rotation**,
   matching its schema description. Previously the IPC backend added the
   supplied angle to the current rotation, so two consecutive
--- a/python/commands/project.py
+++ b/python/commands/project.py
@@ -61,13 +61,16 @@ class ProjectCommands:
             board.SetFileName(board_path)
             pcbnew.SaveBoard(board_path, board)
 
-            # Create schematic from template (use expanded template with symbol definitions)
+            # Create schematic from the empty template (no pre-seeded symbols).
+            # A new project must start as a genuinely blank schematic; the
+            # symbol-seeded templates injected hidden _TEMPLATE_R/C/D instances
+            # that bled into every new project (see issue #221).
             schematic_path = project_path.replace(".kicad_pro", ".kicad_sch")
             template_sch_path = os.path.join(
                 os.path.dirname(os.path.abspath(__file__)),
                 "..",
                 "templates",
-                "template_with_symbols_expanded.kicad_sch",
+                "empty.kicad_sch",
             )
 
             if os.path.exists(template_sch_path):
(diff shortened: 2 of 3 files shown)


boovaragan:
Closing this PR — after rebasing on the latest `main`, I realized this change is not safe to merge as written, and it overlaps work already done in b3d7f61.

**Why I'm closing it:**

1. **The fallback-writer bump is already done, and better.** b3d7f61 ("emit KiCad 10 header in fallback writers") already updates the `create_project`/`create_schematic` fallback to `version 20260306` with `generator "eeschema"` / `generator_version "10.0"`. My change to the same line kept the old `generator "KiCAD-MCP-Server"`, so it's both conflicting and inferior.

2. **Repointing `create_project` to `empty.kicad_sch` would break `add_schematic_component`.** I confirmed that `add_schematic_component` clones a placed `_TEMPLATE_*` instance from the schematic (`component_schematic.py` ~L221, `template_symbol.clone()`), and raises *"Template symbol `_TEMPLATE_R` not found. The schematic must be created from `template_with_symbols.kicad_sch`"* when those instances are absent. A project created from `empty.kicad_sch` has no `_TEMPLATE_*` instances, so component insertion would fail on every newly created project. This is exactly the "changing them affects the component-cloning system" coupling called out in b3d7f61's message — I hadn't accounted for it.

My fixture tests passed only because none of them exercise the `create_project → add_schematic_component` path, so they didn't catch the regression.

**On the still-open part of #221:** new projects created via `create_project` still carry the visible `_TEMPLATE_R/C/D` instances, since the clone system depends on them. The clean fix seems to be decoupling `add_schematic_component` from *placed* instances — e.g. cloning from the `lib_symbols` definitions (or an off-canvas/hidden source) — after which new projects can start genuinely empty. If that matches your intended direction for the separately-tracked template work, I'm happy to take a proper run at it in a fresh PR. Apologies for the noise here.

PR #270, opened 2026-06-30
fix: add_schematic_component emits (pin N (uuid)) entries for ERC (#241)

### LINKED ISSUE #241: add_schematic_component: missing (pin "N" (uuid ...)) entries blocks ERC
## Environment
- KiCad 10.0.1, Windows 11

## Severity
**High** — ERC cannot bind nets to pins; every placed component reports unconnected pins.

## Symptom
Every component instance added via `add_schematic_component` is missing the `(pin "N" (uuid ...))` entries that KiCad requires to associate wires with pins. The ERC reports all pins as unconnected even when wires are correctly drawn.

## Root cause
`dynamic_symbol_loader.py` → `create_component_instance()` builds the instance block as a string template. The template includes `lib_id`, `at`, `unit`, `uuid`, and `property` blocks but emits **no `(pin ...)` entries**:

```python
instance_block = f"""  (symbol (lib_id "{full_lib_id}") (at {x} {y} {angle}) (unit {unit})
    (in_bom yes) (on_board yes) (dnp no)
    (uuid "{new_uuid}")
    (property "Reference" ...)
    ...
    (instances ...)
  )"""
```

KiCad's own save format writes one `(pin "number" (uuid "..."))` entry per pin defined in the embedded `lib_symbol`, e.g.:

```
  (symbol (lib_id "Device:R") (at 100 100 0) (unit 1)
    ...
    (pin "1" (uuid "a1b2c3d4-..."))
    (pin "2" (uuid "e5f6g7h8-..."))
  )
```

Without these, the schematic is invalid for ERC purposes.

## Fix
After writing the instance block, extract pin numbers from the embedded `lib_symbols` definition for `library_name:symbol_name` and append one `(pin "N" (uuid "..."))` per pin with a freshly generated UUID. The pin numbers can be read from the `(pin ... (number "N") ...)` blocks already injected by `inject_symbol_into_schematic`.

## Workaround
Post-process the `.kicad_sch` file with a script that reads pin numbers from the `lib_symbols` block and injects `(pin "N" (uuid "..."))` entries into each matching instance.

boovaragan:
## Summary

Fixes #241. Components added via `add_schematic_component` were missing the `(pin "N" (uuid ...))` entries KiCad requires to bind wires to pins, so ERC reported **every** pin as unconnected even when wires were drawn correctly.

`DynamicSymbolLoader.create_component_instance` synthesizes the placed `(symbol ...)` block as a string template with `lib_id`, `at`, `unit`, `uuid`, `property`, and `instances` — but emitted no per-pin entries. KiCad's own save format writes one `(pin "number" (uuid "..."))` per pin defined in the embedded `lib_symbol`. Without them the schematic is invalid for ERC purposes.

## Changes

- **`python/commands/dynamic_symbol_loader.py`**
  - New helper `_extract_lib_pin_numbers(...)` reads pin numbers from the already-injected `lib_symbols` definition. Pins live in per-unit sub-symbols named `<symbol>_<unit>_<bodystyle>`; pins for a placed `(unit N)` are gathered from sub-symbols whose unit is **N or 0** (unit 0 is common to all units), so multi-unit parts get the correct pins. Duplicate numbers (e.g. from an alternate DeMorgan body style) are dropped while preserving order.
  - `create_component_instance` now appends one `(pin "N" (uuid "..."))` per pin — each with a freshly generated UUID — after the `(instances ...)` block.

- **`tests/test_add_schematic_component.py`** — new `TestCreateComponentInstancePinEntries`:
  - a placed resistor gets exactly pins `1`/`2`, a placed LED gets `1`/`2`
  - each pin gets a distinct UUID
  - the resulting schematic still round-trips via `sexpdata`

- **`CHANGELOG.md`** — Bug Fixes entry under `[Unreleased]`.

## Testing

```
pytest tests/test_add_schematic_component.py::TestCreateComponentInstancePinEntries \
       tests/test_add_schematic_component.py::TestCreateComponentInstanceUnit \
       tests/test_add_schematic_component.py::TestCreateComponentInstanceSubSheet \
       --timeout=60 --timeout-method=thread
# 13 passed
```

All pre-commit hooks (black, isort, prettier, flake8, mypy) pass.

## Notes

This only touches the synthesis path in `DynamicSymbolLoader` (the path the live `add_schematic_component` tool uses) and is independent of the `_TEMPLATE_*` / template-rewrite work tracked under #221/#220.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,20 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **`add_schematic_component` emits `(pin "N" (uuid ...))` entries** (#241):
+  `DynamicSymbolLoader.create_component_instance` built the placed `(symbol ...)`
+  block with `lib_id`, `at`, `uuid`, properties, and `instances`, but no per-pin
+  entries. KiCad needs one `(pin "N" (uuid ...))` per pin to bind wires to pins,
+  so ERC reported every pin as unconnected even when wires were drawn correctly.
+  Each placed instance now carries one entry per pin, with pin numbers read from
+  the embedded `lib_symbols` definition for the placed `(unit ...)` (sub-symbols
+  with unit 0 are common to all units) and a freshly generated UUID per pin.
+
 - **Fallback schematic writer emits the KiCad 10 header** (#221, partial): the
   template-missing fallback in `create_schematic` and `create_project` wrote the
   stale KiCad 9 header `(version 20250114) (generator "KiCAD-MCP-Server")`. It
   now writes `(version 20260306) (generator "eeschema") (generator_version
-  "10.0")`, matching what eeschema writes for a new file. This covers only the
+"10.0")`, matching what eeschema writes for a new file. This covers only the
   fallback path; the main templates (which still carry the KiCad 9 version and
   the `_TEMPLATE_*` clone-source instances used by `add_schematic_component`)
   are tracked separately because rewriting them touches the component-cloning
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -513,6 +513,58 @@ class DynamicSymbolLoader:
         except Exception:
             return {}
 
+    def _extract_lib_pin_numbers(
+        self,
+        schematic_path: Path,
+        library_name: str,
+        symbol_name: str,
+        unit: int = 1,
+    ) -> list:
+        """
+        Return the ordered, de-duplicated list of pin numbers for the requested
+        unit, read from the lib_symbols definition (which must already have the
+        symbol injected).
+
+        Pins live in per-unit sub-symbols named '<symbol>_<unit>_<bodystyle>'.
+        Sub-symbols with unit 0 are common to every unit, so pins for a placed
+        '(unit N)' are gathered from sub-symbols whose unit is N or 0. Duplicate
+        numbers (e.g. from an alternate DeMorgan body style) are dropped while
+        preserving first-seen order. Returns an empty list on failure.
+        """
+        try:
+            with open(schematic_path, encoding="utf-8") as f:
+                content = f.read()
+
+            lib_start = content.find("(lib_symbols")
+            if lib_start == -1:
+                return []
+
+            sym_start = content.find(f'(symbol "{library_name}:{symbol_name}"', lib_start)
+            if sym_start == -1:
+                return []
+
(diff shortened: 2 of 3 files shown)


mixelpixx:
Heads up before you spend more time on this branch: #265 merged today, and it rewrites `create_component_instance` into a complete KiCad 10 instance writer — including per-pin `(pin "N" (uuid ...))` entries for the placed unit plus shared unit-0 pins, which is the core of what this PR adds for #241. The two PRs solved overlapping problems from different angles, and #265 landed first because it also carried the #256 corruption fix.

Could you re-check this PR against current `main`? Two possible outcomes:

1. If #265's pin-uuid emission fully covers the ERC scenario in #241, the most useful remaining contribution would be just your ERC-focused test additions (rebased onto the new writer), so #241 gets closed with a regression test that targets the ERC symptom specifically — the #265 tests assert structure, not ERC behavior on your reproduction.
2. If your approach covers cases #265 misses (different pin selection, alternate defs, whatever you find), rebase and narrow the diff to exactly that delta.

Either way this will need a rebase since both branches modify the same region of `dynamic_symbol_loader.py`. Sorry for the collision — the corruption report forced the ordering.

boovaragan:
Closing this — #265 landed the same fix first, and I've verified it fully covers #241 on current `main`.

**Production fix is already there.** `DynamicSymbolLoader.create_component_instance` now emits one `(pin "N" (uuid ...))` per pin via `_extract_symbol_pins` (`python/commands/dynamic_symbol_loader.py`), including the per-unit selection with unit-0 sub-symbols treated as common to every unit. That's the same mechanism this PR added (`_extract_lib_pin_numbers`), so my change is redundant and only re-collides on that region — which is the conflict flagged here.

**Coverage is already there too.** `tests/test_symbol_instance_completeness.py::TestPerPinUuids` asserts pin entries are emitted with UUIDs for `Device:R` and exercises multi-unit pin selection (unit-0 shared pins), which is broader than the single-unit R/LED cases in this PR. The only assertion here not already covered is per-pin UUID uniqueness, and that's inherent to `uuid.uuid4()` — not worth a separate PR.

One thing #265 didn't add is an **ERC-behavior** regression test that runs against the actual #241 reproduction (wire a placed part, assert ERC reports the pins as connected) rather than asserting structure. If that'd be useful I'm happy to open a small follow-up for it — though it depends on `kicad-cli` ERC being runnable in CI, so I'd want to confirm that's available before committing to it.

Thanks for the heads-up on the ordering — no worries about the collision. Closing in favor of #265.

Issue #225, opened 2026-06-03
[Request] Add close project command


ptr727:
Request to add a command to close the current project.

When kicad exists or closes a project it writes changes to disk, those changes can overwrite changes made by claude.\
Today my agents.md instructions tells claude to ask me to close the project, let it make edits, stash or commit chhanges, then claude reopens the project via mcp.\
It would be much easier if mcp can drive open and close, then claude can orchestrate file and mcp based edits without user involvement.

mixelpixx:
Good request — there's no `close_project`/unload handler today (we have create/open but no symmetric close). Reasonable to add: save-or-warn, drop the in-memory board (and IPC board), clear state. Tracking as a feature to add after the create_project/backend fixes (#220-#223) land, since it shares the lifecycle plumbing. Thanks!

mixelpixx:
Implemented by #271 (merged): close_project with save-or-warn semantics — save=true (default) saves first and refuses to close if the save fails; save=false closes with a warning when in-memory changes are discarded. It drops the SWIG and IPC board state and clears the per-project session pin, so open -> close -> file edits -> open round-trips cleanly.

PR #271, opened 2026-06-30
feat: add close_project command (#225)

boovaragan:
## Summary

Implements #225 — adds a `close_project` command, the symmetric counterpart to `open_project` / `create_project`. As you noted in the issue, this shares the create_project/backend lifecycle plumbing from #220–#223, which has since landed.

Today an agent has to choreograph closing manually (ask the user to close in KiCad, let them edit/stash/commit, then reopen via MCP) because the MCP keeps an in-memory board that a later save can clobber. `close_project` lets the agent release the project itself, so it can mix file-based and MCP-based edits without user involvement.

## Behavior

Matches the design you sketched (save-or-warn, drop the in-memory board + IPC board, clear state):

- **`save` (bool, default `true`)** — saves the board before closing.
  - If `save=true` and the save **fails**, the close is **refused** (`success: false`) so work is never silently lost; the board stays loaded for a retry.
  - If `save=false` and the board has unsaved changes, the close proceeds with a `warnings` entry noting the in-memory changes were discarded.
- **Drops board + clears state**: `self.board = None` propagated to every command handler via `_update_command_handlers()`, plus `ipc_board_api`, the session-backend pin (`session_backend` / `session_board_path`, #223), the disk signature, auto-save status, and project paths — returning the interface to its fresh `__init__` no-project state.
- **Idempotent**: closing with nothing loaded returns `success: true` with `closed: false`.

## Changes

- **`python/kicad_interface.py`** — new `_handle_close_project` + `_clear_project_state` helper; registered `close_project` in `command_routes`.
- **`src/tools/project.ts`** — `close_project` MCP tool with optional `save` boolean.
- **`src/tools/registry.ts`** — added `close_project` to `directToolNames` (project lifecycle).
- **`python/schemas/tool_schemas.py`** — `PROJECT_TOOLS` schema entry.
- **`tests/test_close_project.py`** — 5 unit tests (idempotent no-op, save-and-clear, `save=false` warns on a dirty board, save-failure refuses to close, save-exception refuses to close).
- **`CHANGELOG.md`** — New MCP Tools entry.

### Incidental: a pre-existing mypy blocker on `kicad_interface.py`

While committing I hit a **pre-existing** mypy failure that reproduces on untouched `main` and blocks *any* commit touching `kicad_interface.py` (the pre-commit hook forbids `--no-verify`, and CI runs the same mypy):

```
python/commands/schematic_handlers.py:2642: error: Cannot determine type of "board"  [has-type]
python/commands/schematic_handlers.py:2643: error: Cannot determine type of "board"  [has-type]
```

`SchematicHandlersMixin` reads `self.board`, but `board` is only defined on the concrete `KiCADInterface`; when mypy follows the import into the mixin it hits an attribute-resolution cycle. I added a type-check-time-only declaration (`if TYPE_CHECKING: board: Any`) to the mixin — no runtime effect — which is exactly the contract the mixin's own module docstring already describes ("every `self.board` … is supplied by the interface"). Happy to split this into its own PR if you'd prefer it separate.

## Testing

```
pytest tests/test_close_project.py --timeout=60 --timeout-method=thread   # 5 passed
npm run build                                                             # clean
npm run test:ts                                                           # 24 passed
```

All pre-commit hooks (black, isort, prettier, flake8, mypy, eslint) pass.

## Notes

`close_project` saves via the SWIG `save_project` path (the in-memory board the MCP owns), consistent with how `save_project` behaves for SWIG-pinned sessions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ All notable changes to the KiCAD MCP Server project are documented here.
   template-missing fallback in `create_schematic` and `create_project` wrote the
   stale KiCad 9 header `(version 20250114) (generator "KiCAD-MCP-Server")`. It
   now writes `(version 20260306) (generator "eeschema") (generator_version
-  "10.0")`, matching what eeschema writes for a new file. This covers only the
+"10.0")`, matching what eeschema writes for a new file. This covers only the
   fallback path; the main templates (which still carry the KiCad 9 version and
   the `_TEMPLATE_*` clone-source instances used by `add_schematic_component`)
   are tracked separately because rewriting them touches the component-cloning
@@ -107,6 +107,16 @@ All notable changes to the KiCAD MCP Server project are documented here.
   the failure on a real-world user schematic.
 
 ### New MCP Tools
+
+- `close_project` (#225) — Symmetric counterpart to `open_project` /
+  `create_project`. Optionally saves the board (`save`, default `true`), then
+  drops the in-memory board (SWIG + IPC) and clears all per-project session
+  state (session-backend pin, disk signature, project paths). Lets an agent
+  hand control back so the user — or the agent itself — can edit project files
+  directly without a later MCP save clobbering those changes, which previously
+  required manual open/close choreography. If `save=true` and the save fails,
+  the close is refused so work is never silently lost; if `save=false` on a
+  board with unsaved changes, the close proceeds with a warning.
 
 - `add_gnd_stitching_vias` — Drop GND stitching vias across the board with
   collision checking against every non-GND segment, via, and pad on every
--- a/python/commands/schematic_handlers.py
+++ b/python/commands/schematic_handlers.py
@@ -18,11 +18,10 @@ import subprocess
 import tempfile
 import traceback
 from pathlib import Path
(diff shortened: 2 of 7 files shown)


mixelpixx (approved):
This is exactly the design sketched in #225, and the integration details are the part I checked hardest: _clear_project_state clears the #223/#255 session pin (session_backend / session_board_path), the IPC board API, the disk signature, and propagates the dropped board to every command handler — so a subsequent open really does start from the fresh-init state, and no stale pin can route a later command to the wrong backend. The save-failure-refuses-to-close semantics are the right default for an agent-facing tool: silently losing work is the one unrecoverable outcome.

Your branch conflicted trivially with today's merges (both sides added the mixin board annotation for the same pre-existing mypy failure you documented — #268 landed a pyproject override and main also gained a bare class-level annotation). Since maintainer edits are allowed I pushed the resolution keeping main's variant; your TYPE_CHECKING version was equally correct, there is just no need for both.

Verified locally on the resolved merge: your 5 close_project tests plus the 24 backend pinning/read-fallback tests pass, full suite matches the known-environmental baseline, black/mypy clean, tsc and vitest green. Merging; #225 closes with this.

PR #272, opened 2026-07-01
fix: locate pins against their own unit for multi-unit symbols (#239)

### LINKED ISSUE #239: get_schematic_pin_locations returns wrong coordinates for multi-unit components
get_schematic_pin_locations (and get_pin_location) returns all pin world-coordinates computed from the placement of the first unit found in the schematic file, regardless of which unit each pin actually belongs to. For a three-unit NE5532 placed at y=175 (unit A), y=210 (unit B), y=245 (power), every pin is reported at y≈245 because the power unit happened to appear first in file order.

Steps to reproduce

Place a multi-unit symbol (e.g. Amplifier_Operational:NE5532) as three separate units at distinct positions — e.g. unit 1 at (300, 175), unit 2 at (300, 210), unit 3 at (300, 245).
Call get_schematic_pin_locations with the component reference.
Observe: all 8 pins are reported near y=245 (the power unit position) regardless of which unit they belong to.
Expected behaviour

Pin 3 (unit A) → y≈172. Pin 5 (unit B) → y≈207. Pin 8 (power unit) → y≈237.

boovaragan:
## Summary

Fixes #239 — `get_schematic_pin_locations` / `get_pin_location` returned wrong
coordinates for multi-unit components.

A multi-unit part (e.g. an `NE5532` dual op-amp) is placed as several
`(symbol ...)` instances that share one reference designator but each carry
their own `(unit N)`, position, and rotation. The pins for unit N live in the
`<name>_<unit>_<body>` sub-symbol inside `lib_symbols`.

The locator ignored units entirely: it looked up the symbol transform with
`WireDragger.find_symbol`, which returns the **first** `(symbol)` instance in
file order, and applied that single position to every pin. So for the three-unit
NE5532 in the issue — units at y=175, y=210, y=245 — every pin came back near
y≈245 because the power unit happened to be written first.

## Fix (all in `python/commands/pin_locator.py`)

1. **Tag each pin with its owning unit.** `parse_symbol_definition` now tracks
   the enclosing sub-symbol while recursing and records `pin_data["unit"]` from
   the `_<unit>_<body>` suffix (unit 0 = "common to all units", used as the
   default context so single-unit parts are unchanged).
2. **Locate each pin against its unit's placed instance.** `_get_symbol_transform`
   takes an optional `unit`; a new `_find_symbol_instance_for_unit` helper scans
   the placed `(symbol)` instances and returns the one whose `(unit N)` matches
   (same Reference parsing and trailing-`_` tolerance as `find_symbol`).
   `get_pin_location` and `get_pin_angle` resolve the pin's unit and re-read the
   transform for that instance, so position **and** rotation/mirror come from the
   correct unit. If a unit isn't placed (or the pin is unit 0), it falls back to
   the first instance — preserving the original single-unit behaviour.

## Tests

New `tests/test_pin_locator_multi_unit.py` places a 2-unit symbol at two
distinct positions and asserts:

- `parse_symbol_definition` tags pins 1–3 as unit 1 and pins 5–7 as unit 2;
- unit-1 pins resolve against the unit-1 instance and unit-2 pins against the
  unit-2 instance (explicitly checking a unit-2 pin is **not** at the unit-1
  position, the exact failure in #239).

```
pytest tests/test_pin_locator_multi_unit.py                 # 4 passed
pytest tests/test_pin_locator_and_component.py \
       tests/test_pin_locator_y_flip.py \
       tests/test_pin_locator_duplicate_pin_defs.py \
       tests/test_get_pin_angle.py \
       tests/test_net_label_pin_snapping.py                 # no regressions
```

Single-unit parts are unaffected; the change is additive (`pin_data` gains a
`unit` key) and the default lookup path is unchanged when no unit is requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,22 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **Pin locations respect the owning unit of multi-unit symbols** (#239):
+  `get_schematic_pin_locations` / `get_pin_location` located every pin against
+  whichever placed `(symbol)` instance appeared first in file order, so for a
+  multi-unit part (e.g. a dual op-amp placed as separate units at different
+  positions) all pins collapsed onto that one unit's coordinates. Pins are now
+  tagged with their owning unit while parsing `lib_symbols` (the
+  `<name>_<unit>_<body>` sub-symbol), and each pin is located against the placed
+  instance carrying the matching `(unit N)` — with rotation/mirror read from that
+  same instance. Single-unit parts are unaffected (they fall back to the first
+  instance as before).
+
 - **Fallback schematic writer emits the KiCad 10 header** (#221, partial): the
   template-missing fallback in `create_schematic` and `create_project` wrote the
   stale KiCad 9 header `(version 20250114) (generator "KiCAD-MCP-Server")`. It
   now writes `(version 20260306) (generator "eeschema") (generator_version
-  "10.0")`, matching what eeschema writes for a new file. This covers only the
+"10.0")`, matching what eeschema writes for a new file. This covers only the
   fallback path; the main templates (which still carry the KiCad 9 version and
   the `_TEMPLATE_*` clone-source instances used by `add_schematic_component`)
   are tracked separately because rewriting them touches the component-cloning
--- a/python/commands/pin_locator.py
+++ b/python/commands/pin_locator.py
@@ -7,6 +7,7 @@ Uses S-expression parsing to extract pin data from symbol definitions.
 
 import logging
 import math
+import re
 import tempfile
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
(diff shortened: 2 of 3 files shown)


mixelpixx (approved):
Excellent fix — the unit-tagging while walking lib_symbols sub-symbols is the right mechanism, and the fallback to the first instance keeps single-unit parts byte-identical in behavior.

Heads up: today's merge of #259 rewrote get_pin_angle's math (it now derives the bearing via WireDragger.pin_world_xy instead of hand-rolled angle arithmetic), which conflicted with your unit-aware transform in the same lines. Since you allow maintainer edits, I pushed the merge resolution to your branch rather than bouncing it back: the unit's instance now supplies the FULL transform — position included, not just rotation/mirror as your original unpack did — because pin_world_xy anchors the endpoint at (sym_x, sym_y), and computing a multi-unit pin's bearing from the first instance's origin would be wrong in exactly the multi-unit cases this PR fixes.

Verified after resolution on this machine with a real KiCad 10.0.0 (kicad-cli on PATH, so the eeschema-truth oracle actually ran): all 27 pin-geometry tests pass — your 3 new multi-unit tests plus the full pin_world_xy netlist-oracle matrix — and the full suite matches the known-environmental baseline. black and mypy clean. Merging; #239 closes with this.

PR #273, opened 2026-07-01
Fail fast on concurrent calls (no more apparent freezes) + fix IPC socket fd leak

stefangordon:
## Problem

The KiCAD bridge is single-threaded — commands are serialized over one stdio connection and one IPC socket (`callKicadScript` → `requestQueue` → `processNextRequest`). When an operation is already in flight, a second concurrent tool call sits **silently in the queue** behind it, potentially for up to the 10-minute long-op timeout used by `export_*`, `run_drc`, and `sync_schematic_to_board`. With no feedback, this presents to MCP clients as a **frozen server**.

This is easy to hit in practice: LLM agents frequently batch multiple tool calls in a single turn, which the MCP client dispatches concurrently. The first slow call blocks all siblings, and the agent reports "the MCP froze."

Separately, `IPCBackend.disconnect()` leaks file descriptors: it dropped the kipy `KiCad` reference without closing the underlying `pynng` socket. Long-lived sessions that reconnect accumulate fds (observed a bridge climb past 20+ open sockets over a session).

## Changes

**1. Fail fast with an actionable error on concurrent calls.** When an operation is in flight, new commands are rejected immediately instead of queueing indefinitely. The error explains that the server is serialized, names the in-flight command and how long it has been running, and tells the caller to wait and retry sequentially:

> KiCAD MCP server is busy: still processing "export_gerber" (running for 42s). This server handles ONE KiCAD operation at a time — commands are serialized over a single connection to KiCad. Do not issue KiCAD tool calls in parallel; wait for the in-flight operation to finish, then retry "get_board_info". Exports, DRC, and schematic sync can take a while on large designs.

An agent that receives this backs off and serializes its calls, instead of hanging. The startup warm-up is labeled (`startup warm-up (...)`) so calls during that brief window get an accurate message too.

**2. Close the IPC socket on disconnect.** kipy exposes no public `close()`, but holds a persistent `pynng` socket at `KiCad._client._conn`. `disconnect()` now closes it explicitly, guarded with `getattr`/try-except so it degrades gracefully if kipy internals change.

## Verification

- **Concurrency:** fired two concurrent `get_backend_state` calls at a fresh server over raw MCP stdio — one succeeds, the other returns the busy error naming the in-flight command. Sequential calls are unaffected (the first resolves before the second is issued, so the guard never trips).
- **fd leak:** ran 10 connect/disconnect cycles — fds stay flat (delta +2 from baseline, stable) vs. growing ~3 per cycle before the fix.

## Note / possible follow-up

The 10-minute blocking behavior of long ops is architectural (serialized stdio). This PR makes it *visible and non-hanging* rather than eliminating it. A future change could run long exports without holding the request loop (e.g. a job/poll model) so unrelated queries stay responsive during a big export.

### DIFF
--- a/python/kicad_api/ipc_backend.py
+++ b/python/kicad_api/ipc_backend.py
@@ -133,8 +133,22 @@ class IPCBackend(KiCADBackend):
             return "unknown"
 
     def disconnect(self) -> None:
-        """Disconnect from KiCAD."""
+        """Disconnect from KiCAD.
+
+        kipy does not expose a public close(), but it holds a persistent pynng
+        socket at ``KiCad._client._conn``. Simply dropping the reference orphans
+        that socket and leaks its OS file descriptors (they are not released
+        promptly by GC), so long-lived sessions that reconnect accumulate fds.
+        Close the underlying socket explicitly, tolerating any internal changes.
+        """
         if self._kicad:
+            try:
+                conn = getattr(getattr(self._kicad, "_client", None), "_conn", None)
+                if conn is not None and hasattr(conn, "close"):
+                    conn.close()
+                    logger.debug("Closed underlying KiCAD IPC socket")
+            except Exception as e:
+                logger.debug(f"Error closing IPC socket during disconnect: {e}")
             self._kicad = None
             self._connected = False
             logger.info("Disconnected from KiCAD IPC")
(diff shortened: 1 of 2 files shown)


stefangordon:
Closing in favor of two focused PRs, one concern each per CONTRIBUTING: #275 (fail fast on concurrent calls) and #276 (close IPC socket on disconnect / fd leak). Same changes, split for easier review.

Issue #235, opened 2026-06-07
get_board_2d_view: KiCad 10 requires --layers; pcbPath fallback broken


Dewieinns:
## Follow-up to #209 (resolved ✅)

PR #212 fixes the three KiCad 10 breakages reported in #209 and is confirmed working on KiCad 10.0.1 / Windows 11. Thanks for the fast turnaround.

Two more issues surfaced during verification on the same environment.

---

### Symptom
Calling `get_board_2d_view` without specifying `layers` returns:
```json
{ "success": false, "message": "kicad-cli SVG export failed" }
```

### Root cause
`kicad-cli pcb export svg` in KiCad 10 requires at least one layer — passing no `--layers` argument (the current code path when `layers` is empty) exits with:
```
At least one layer must be specified
Exit code: 1
```

### Fix
When `layers` is empty, query the board's enabled layers via `pcbnew.LoadBoard()` before building the command, falling back to a standard set if that fails:

```python
if not layers:
    try:
        import pcbnew as _pcb
        _b = _pcb.LoadBoard(pcb_path)
        layers = [
            _b.GetLayerName(i)
            for i in range(_pcb.PCB_LAYER_ID_COUNT)
            if _b.IsLayerEnabled(i)
        ]
    except Exception:
        layers = [
            "F.Cu", "B.Cu", "Edge.Cuts",
            "F.Silkscreen", "B.Silkscreen",
            "F.Mask", "B.Mask", "F.Fab", "B.Fab",
        ]
cmd += ["--layers", ",".join(layers)]
```

Verified working on KiCad 10.0.1 / Windows 11.

---

### Symptom
Omitting `pcbPath` (to use the currently loaded board) returns:
```json
{ "success": false, "message": "pcbPath required" }
```
even when a board is loaded.

### Expected behavior
The tool description says `pcbPath` falls back to the currently loaded board if omitted. This fallback doesn't appear to work — the path check fails before reaching the loaded-board branch.

Passing `pcbPath` explicitly works fine as a workaround.

---

## Environment
- OS: Windows 11 Pro 10.0.26200
- KiCad: 10.0.1
- MCP server: `main` as of 2026-06-03 (post-PR-#212 merge)

mixelpixx:
[...]

PR #274, opened 2026-07-01
Rasterize schematic view with cairosvg so renders fit inline size caps

stefangordon:
## Problem

`get_schematic_view` / `get_schematic_view_region` can't produce an image under an MCP client's inline cap (~1MB), and `width`/`height` have no effect on the payload.

Root cause in `_svg_to_png` (`python/commands/schematic_handlers.py`): it tries pymupdf → inkscape → ImageMagick, **none of which are declared dependencies**. On a stock install the only image library present is **cairosvg** (which *is* in `requirements.txt`), so every converter fails, `_svg_to_png` returns `None`, and the handler silently falls back to returning the raw full-sheet SVG as text — a complete page (title block, grid, every path/font) that routinely exceeds 1MB and, being vector, ignores width/height.

## Change

Use **cairosvg** as the primary converter (already a declared dependency; honors `output_width`/`output_height`). pymupdf/inkscape/convert remain as fallbacks. Adds `tests/test_svg_to_png_render.py` covering: returns PNG (not None), honors requested dimensions, and smaller dimensions shrink the payload.

| Output | base64 size |
|---|---|
| Raw SVG (old fallback) | multi-MB, ignores width/height |
| cairosvg PNG @ 1200x900 | ~0.020 MB |
| cairosvg PNG @ 800x600 | ~0.012 MB |

## Testing / notes

New test passes; lint clean (Black/isort). Note: `mypy` reports a **pre-existing** `has-type` error on unrelated board-loading code in this file (present on `main`); left untouched to keep this PR single-purpose. Some other schematic tests hang/error in a KiCad-less environment on `main` too — unrelated to this change.

### DIFF
--- a/python/commands/schematic_handlers.py
+++ b/python/commands/schematic_handlers.py
@@ -22,7 +22,6 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import pcbnew
 import sexpdata
-
 from commands.library_schematic import LibraryManager as SchematicLibraryManager
 from commands.schematic import SchematicManager
 from commands.wire_manager import WireManager
@@ -31,16 +30,34 @@ logger = logging.getLogger("kicad_interface")
 
 
 def _svg_to_png(svg_path: str, width: int, height: int) -> Optional[bytes]:
-    """Convert SVG to PNG. No cffi dependency.
+    """Convert SVG to PNG at the requested pixel dimensions.
 
     Priority:
-      1. pymupdf (fitz) — bundled MuPDF renderer, pure Python, no system deps
-      2. Inkscape CLI — accurate KiCAD SVG rendering
-      3. ImageMagick convert — broad availability fallback
+      1. cairosvg — the declared dependency (see requirements.txt); honors
+         output_width/output_height and reliably rasterizes KiCAD SVGs.
+      2. pymupdf (fitz) — bundled MuPDF renderer, if installed.
+      3. Inkscape CLI — accurate KiCAD SVG rendering, if installed.
+      4. ImageMagick convert — broad availability fallback.
     Returns PNG bytes or None if all converters fail.
+
+    Rasterizing here is essential: returning the raw KiCAD SVG instead
+    produces a full-sheet vector document (title block, grid, every path
+    and font) whose text easily exceeds an MCP client's inline size cap,
+    and width/height have no effect on it.
     """
     import subprocess
     import tempfile
+
+    try:
+        import cairosvg
+
+        return cairosvg.svg2png(
+            url=svg_path,
+            output_width=int(width),
+            output_height=int(height),
+        )
+    except Exception:
+        pass
 
     try:
         import fitz
--- a/tests/test_svg_to_png_render.py
+++ b/tests/test_svg_to_png_render.py
@@ -1,0 +1,74 @@
+"""Tests for ``_svg_to_png`` used by ``get_schematic_view``.
+
+Regression guard for the bug where the converter chain (pymupdf → inkscape
+→ ImageMagick) had no available backend on a stock install, so the function
+returned ``None`` and the handler silently fell back to emitting the raw
+full-sheet SVG — which ignores width/height and blows past MCP clients'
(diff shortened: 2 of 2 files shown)


mixelpixx (changes requested):
The problem statement is exactly right — the converter chain has no declared dependency, so a stock install silently falls back to the oversized raw SVG, and your size table makes the impact clear. The direction (make the first converter one we actually ship) is correct too. But I have to request changes on the specific choice of cairosvg, because I ran this branch on a stock Windows setup and the fix does not work there:

```
cannot load library 'libcairo-2.dll': error 0x7e.  Additionally,
ctypes.util.find_library() did not manage to locate a library called 'libcairo-2.dll'
```

All three of your new tests fail on Windows for this reason (`png is None`). The `cairosvg` pip package is only a wrapper — it needs the native cairo runtime, which pip does not install and stock Windows does not have (it typically arrives via the GTK runtime or MSYS2, neither of which we can assume). So on Windows — the platform most of this repo's setup tooling targets (`setup-windows.ps1`, the Windows troubleshooting guide) — cairosvg imports, raises at first use, and the chain falls through to the same broken state this PR is fixing. It works on Linux because distros ship libcairo2, which is also why our Linux-only CI would pass and hide this.

Concrete suggestion that keeps your design intact: make **pymupdf** the declared, first-priority converter instead. It was already the chain's #1 precisely because it bundles the MuPDF renderer with no system dependencies, and it ships binary wheels for Windows/macOS/Linux — `pip install pymupdf` is sufficient everywhere. The change would be:

1. Add `pymupdf>=1.24` to `requirements.txt` (it is the only converter in the chain that is pip-complete on all three platforms).
2. Keep the chain order `pymupdf -> cairosvg -> inkscape -> convert` (cairosvg stays a fine second choice where cairo exists).
3. Your tests then pass unmodified on all platforms, and they become a real regression guard for the "stock install must have a working converter" invariant — which is the most valuable part of this PR.

If you prefer to keep cairosvg first for output-size or fidelity reasons, that also works, but then `requirements.txt` alone cannot guarantee a working converter on Windows and the tests need a skip guard for missing native cairo — which would defeat the invariant the tests exist to enforce, so I'd rather have pymupdf declared.

Everything else verified locally: merges cleanly with current main, black and mypy clean, and the fallback behavior (raw SVG) is unchanged when every converter is missing.

mixelpixx:
Ping — no activity since 2026-07-01, and the 07-03 review is still the outstanding item.

The objection is unchanged and is the same one on #277: this inserts cairosvg as the **priority-1** rasterizer ahead of the existing pymupdf/fitz chain (`schematic_handlers.py:61`), and cairo's native DLL is absent on a stock Windows install. `tests/test_cairo_dll_preload_windows.py` exists on main because this has already bitten the project.

The smallest change that makes this mergeable: keep pymupdf primary and add cairosvg as a *fallback* rather than ahead of it. Your own `test_svg_to_png_render.py` asserts that a converter is available without asserting which, so it should pass unchanged under that ordering — this may be a one-line reorder.

Note I landed the `--layers` half of #277 separately in #337, with your authorship preserved, because it was blocking issue #235 and PR #324. This PR and #277's second commit are the remaining pieces.

If you have moved on, that is completely fine — say so (or just don't reply) and I will close #274, #275, #277 and #278 as partially-landed rather than leave them hanging. Your `--layers` fix and the three PRs you already had merged are appreciated either way.

Heads-up if you do return: main moved a lot today. CI now genuinely runs the Python suite (#334 — it was a no-op, `pytest python/` collected 7 tests instead of ~1550), so rebase before pushing.

mixelpixx:

mixelpixx:
Closing in favor of #384, which lands this with credit — your commits' diagnosis and tests, with the one change requested in review applied: pymupdf as the declared first-priority converter (binary wheels everywhere, unlike cairosvg's native cairo runtime which stock Windows lacks), cairosvg second. Your tests are kept as written apart from the docstring. Thank you — the 'stock install must have a working converter' invariant those tests pin was the most valuable part of this PR, and it survives intact.

Issue #251, opened 2026-06-11
autoroute: MCP framework timeout kills call before Freerouting finishes; worker appears to fail


Dewieinns:
## Environment
- KiCad 10.0.1, Windows 11
- Freerouting 2.2.4 (JAR), Java 26
- MCP server: main post-PR-#212
- Board: >20 nets

## Severity
**Medium** — UX issue; surfaces as a false failure, but Freerouting may have completed.

## Symptom
Calling `autoroute` on a board with more than ~20 nets returns a timeout error from the MCP framework before Freerouting finishes. The `.ses` output file may exist on disk and be valid, but the MCP tool has already reported failure.

## Root cause
`commands/freerouting.py` → `autoroute()` has an internal `timeout` parameter (default 300s, line 302). However, the **MCP framework's own call timeout** (Node.js side, typically 30s) fires first for any non-trivial board. The Python subprocess is killed or the response is dropped, even though Freerouting is still running or has already produced a `.ses` file.

## Expected behaviour
- Either make the tool fire-and-forget (return immediately with a job ID) and provide a separate `check_autoroute_status` / `get_autoroute_result` polling tool.
- Or stream progress updates during routing so the MCP connection stays alive.
- At minimum, surface the `.ses` path in the error response so the caller can call `import_ses` manually if the file exists.

## Workaround
1. Note the `.ses` output path from the `autoroute` parameters before calling.
2. After the timeout error, check whether the `.ses` file exists on disk.
3. If it does, call `import_ses` directly with that path.
4. Use `Monitor` against the Freerouting output directory to wait for the `.ses` to appear.

mixelpixx:
Fixed in #335, now on main.

Your diagnosis was exactly right: the Node bridge had a 30 s default and `autoroute` was missing from its long-running list, so it abandoned the call while the Python side was still working through its own 300 s budget. Hence a failure report against a `.ses` that existed and was valid.

Rather than just moving `autoroute` onto the fixed 600 s tier, the timeout is now derived from what you actually ask for. The Python side treats `timeout` as **per attempt** and repeats it `attempts` times for best-of-N, so a flat ceiling would still have cut you off — `attempts: 5` alone is a 1500 s routing budget. It now computes `attempts × timeout + 120 s` of overhead (DSN export, JVM start per attempt, SES import, board save), floored at 600 s. So a bigger board just needs a bigger `timeout`, and it will be honoured.

[...]

Issue #248, opened 2026-06-11
sync_schematic_to_board: per-footprint FootprintLoad() cost makes non-trivial boards hang (partly fixed — library cache landed in #310)


Dewieinns:
## Environment
- KiCad 10.0.1, Windows 11
- MCP server: main post-PR-#212
- Board: ~40 schematic components, empty PCB (4 mounting holes only)

## Severity
**Critical** — blocks the entire rebuild flow; kills the Python worker.

## Symptom
Calling `sync_schematic_to_board` on a board with ~40 footprint-bearing schematic components times out after 600s. All subsequent MCP calls then return:

```
Python process for KiCAD scripting is not running
```

The PCB file is left unchanged.

## Root cause
`_handle_sync_schematic_to_board` calls `_add_missing_footprints_from_schematic`, which:

1. Instantiates `LibraryManager(project_path=project_dir)` — this constructor scans and parses the full `fp-lib-table` (100+ standard KiCad libraries, potentially thousands of `.pretty` directory entries) **on every sync call**.
2. For each of the 40 schematic components, calls `pcbnew.FootprintLoad(library_path, fp_name)` — each call can take 100–500 ms on a cold library.
3. The combined 40 × FootprintLoad + library scan can easily exceed the MCP framework's internal timeout, causing the Node.js side to kill the connection while the Python subprocess is still working, leaving the worker in a broken state.

## Expected behaviour
- Either stream progress so the caller can see the operation is still running.
- Or complete in a reasonable time (comparable to KiCad UI's F8, which runs in < 5s for the same board).

## Fix ideas
- Cache `LibraryManager` across calls so the `fp-lib-table` is not re-parsed on every `sync_schematic_to_board` invocation.
- Batch `FootprintLoad` calls and/or load libraries once rather than per-footprint.
- Add a configurable timeout parameter and return a partial-progress result rather than letting the worker crash.

## Workaround
Use KiCad UI directly: `File → Open → <project>.kicad_pro`, switch to PCB Editor, press **F8** (`Tools → Update PCB from Schematic`). The native operation completes in seconds for the same board.

mixelpixx:
[...]

PR #275, opened 2026-07-01
Fail fast with a clear error when a KiCAD operation is already in flight

stefangordon:
## Problem

The bridge is single-threaded — commands are serialized over one stdio connection and one IPC socket (`callKicadScript` → `requestQueue` → `processNextRequest`). While one command is in flight, a second concurrent tool call sits **silently in the queue** behind it, for up to the 10-minute long-op timeout used by `export_*` / `run_drc` / `sync_schematic_to_board`. With no feedback, this presents to MCP clients as a **frozen server** — and it's easy to hit because agents often batch tool calls in a single turn, which the client dispatches concurrently.

## Change (server.ts only)

When an operation is already in flight, reject new commands immediately with an actionable message that names the in-flight command and how long it has run, and tells the caller the server is serialized and to retry sequentially. The startup warm-up is labeled so calls during that brief window get an accurate message too. Sequential usage is unaffected (the first call resolves before the next is issued, so the guard never trips).

Example message:
> KiCAD MCP server is busy: still processing "export_gerber" (running for 42s). This server handles ONE KiCAD operation at a time … wait for the in-flight operation to finish, then retry "get_board_info".

## Relationship to open issues

Relates to #248 and #251, where a slow/blocking operation makes subsequent calls appear to hang. This surfaces the serialization and stops the pile-up; it does **not** by itself fix the underlying long-op behavior described in those issues (e.g. the `sync_schematic_to_board` algorithm in #248). Flagging rather than claiming to close them.

## Testing

`npm run test:ts` passes (24/24). Verified manually over raw MCP stdio: two concurrent `get_backend_state` calls → one succeeds, the other returns the busy error naming the in-flight command; sequential calls behave normally. Lint clean (Prettier/ESLint).

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -229,6 +229,10 @@ export class KiCADMcpServer {
     reject: Function;
   }> = [];
   private processingRequest = false;
+  /** Name of the command currently being processed by the Python bridge (for busy diagnostics). */
+  private inFlightCommand: string | null = null;
+  /** Epoch ms when the in-flight command started (for busy diagnostics). */
+  private inFlightStartedAt: number | null = null;
   private responseBuffer: string = "";
   private currentRequestHandler: {
     resolve: Function;
@@ -600,7 +604,6 @@ export class KiCADMcpServer {
       await this.runWarmup(120_000);
       logger.info("Warm-up complete — pcbnew/wxApp initialised");
 
-
       // Write a ready message to stderr (for debugging)
       process.stderr.write("KiCAD MCP SERVER READY\n");
 
@@ -634,14 +637,14 @@ export class KiCADMcpServer {
   private async waitForReady(timeoutMs: number): Promise<void> {
     return new Promise((_resolve, reject) => {
       const timeout = setTimeout(() => {
-        reject(new Error(
-          `Python process did not send READY within ${timeoutMs / 1000} s`
-        ));
+        reject(new Error(`Python process did not send READY within ${timeoutMs / 1000} s`));
       }, timeoutMs);
-      this.readyPromise.then(() => {
-        clearTimeout(timeout);
-        _resolve();
-      }).catch(reject);
+      this.readyPromise
+        .then(() => {
+          clearTimeout(timeout);
+          _resolve();
+        })
+        .catch(reject);
     });
   }
 
@@ -668,30 +671,32 @@ export class KiCADMcpServer {
       const timeoutHandle = setTimeout(() => {
         logger.warn(
           `Warm-up timed out after ${timeoutMs / 1000} s — ` +
-          "continuing without full initialisation"
+            "continuing without full initialisation",
         );
         this.responseBuffer = "";
         this.processingRequest = false;
         this.currentRequestHandler = null;
+        this.inFlightCommand = null;
+        this.inFlightStartedAt = null;
         resolve();
       }, timeoutMs);
 
       // Use the existing request infrastructure to avoid race conditions
       // with the persistent stdout handler.
       this.processingRequest = true;
+      this.inFlightCommand = "startup warm-up (pcbnew/wxApp initialization)";
+      this.inFlightStartedAt = Date.now();
       this.currentRequestHandler = {
         resolve: (result: any) => {
           clearTimeout(timeoutHandle);
           this.processingRequest = false;
           this.currentRequestHandler = null;
+          this.inFlightCommand = null;
+          this.inFlightStartedAt = null;
           if (result?.success) {
-            logger.info(
-              `Warm-up succeeded: pcbnew ${result.version} (${result.elapsed_s}s)`
-            );
+            logger.info(`Warm-up succeeded: pcbnew ${result.version} (${result.elapsed_s}s)`);
           } else {
-            logger.warn(
-              `Warm-up returned failure: ${result?.message || "unknown"} — continuing`
-            );
+            logger.warn(`Warm-up returned failure: ${result?.message || "unknown"} — continuing`);
           }
           resolve();
         },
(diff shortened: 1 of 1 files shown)


mixelpixx (changes requested):
The diagnosis is right (silent queue behind a 10-minute export reads as a frozen server), the in-flight bookkeeping is clean, and the busy message naming the running command and its elapsed time is exactly what a caller needs. But I can't take the guard as written, because it rejects unconditionally whenever anything is in flight — and that turns the most common concurrent pattern from "works" into "errors".

Concrete scenario, straight from your own problem statement: agents batch tool calls in a single turn, so the client dispatches, say, `get_board_info` and `get_backend_state` concurrently. Today the second call waits ~100ms in the queue behind the first and both succeed. With this change the second call fails instantly with the busy error, and the agent burns a retry round-trip — on every batched turn, forever. The failure mode you're fixing (stuck behind a long export) and the case you'd be breaking (momentarily behind a fast read) are both "processingRequest is true"; the guard needs to distinguish them, and elapsed/expected duration is the signal that does it.

Two designs that keep your diagnostics and fix the freeze without breaking batching:

1. **Bounded wait (preferred):** enqueue as today, but attach a timer to each queued item. If it hasn't started within N seconds (5s, say), reject it with exactly your busy message. Fast reads drain in milliseconds and never see the timer; anything stuck behind a DRC/export gets your actionable error instead of a silent 10-minute wait. This also keeps the existing requestQueue meaningful — as written, the guard makes it structurally impossible for the queue to ever hold more than the one in-flight item, so the queue machinery becomes dead code, which is a hint the design is fighting the existing architecture rather than extending it.

2. **Long-op cutoff:** reject immediately only when the in-flight command is a known long-runner (you already special-case these for timeouts: export_*, run_drc, sync_schematic_to_board) or has been running longer than a few seconds; otherwise queue as today.

Either way, please add a vitest for the semantics (queued-then-served for two fast concurrent calls; rejected-with-busy-message when the wait bound is exceeded) — `server.ts` currently has no coverage of the queue behavior, and this is exactly the kind of semantics change that regresses silently.

Happy to re-review quickly once the guard is wait-bounded — the message text and the warm-up labeling can stay exactly as they are.

mixelpixx:
Checking in here too — and apologies for two notifications in one pass, this is the other half of your queue alongside #277.

Nothing has changed on the branch since the 07-03 review, so the one blocking item still stands. Restating it with the reasoning, since the diagnosis in this PR was correct and it is only the remedy that needs adjusting:

**The diagnosis is right.** A silent queue behind a ten-minute export is genuinely indistinguishable from a hung server, and the in-flight bookkeeping you added is the right instrumentation — `inFlightCommand` and `inFlightStartedAt`, cleared on every exit path including the warm-up timeout. The error text is genuinely good: it names the blocking command, reports how long it has been running, and explains that operations are serialized. That is a real improvement over silence.

**The problem is that the reject is unconditional.** The current handler rejects the moment `processingRequest` is true:

```
Rejecting "${command}" — busy with "${running}"${elapsedStr}
```

That converts every concurrent call into a hard failure, including ones that would have succeeded after a few hundred milliseconds. MCP clients routinely dispatch several tool calls in one batch — that is normal client behavior, not misuse — so this turns a working (if slow) path into a broken one. The failure mode moves from "feels frozen" to "randomly errors under normal use", which is worse, because at least the first one eventually returns a result.

**What would unblock it: a bounded wait before rejecting.** Queue the request and wait up to some ceiling — a few seconds is plenty to absorb ordinary batching — and only reject with your (excellent) diagnostic message if the wait expires. Short overlaps then just work, while a genuinely stuck ten-minute export still produces the clear error instead of an infinite hang. The queue infrastructure is already there in `requestQueue`; this is mostly about not short-circuiting it.

Alongside that, **vitest coverage** for the two behaviors that matter: a request that overlaps briefly and then succeeds, and one that exceeds the ceiling and rejects with the diagnostic. `tests-ts/` has the pattern to follow.

Worth noting this one has grown more relevant, not less — #305 (draft) also touches request-ordering hardening in `src/server.ts`, and the intent has been to land this first so that work builds on it rather than colliding. So it is not stale, it is queued.

The branch still merges cleanly into current `main` despite being 168 commits behind, so no rebase is needed — just the bounded-wait change and the tests.

mixelpixx:
Ping — no activity since 2026-07-01; the 07-03 review is still outstanding.

Recap of the one objection: rejecting **immediately and unconditionally** whenever `processingRequest` is true converts ordinary short overlaps into hard errors. MCP clients batch tool calls routinely, so "feels frozen" becomes "randomly errors" — which is worse, because it is nondeterministic. What is wanted is a **bounded wait**: queue on the existing `requestQueue` and reject only on expiry.

Two things you should know before picking this up again:

1. **#305 touches the same fields from the opposite direction.** It removes the state-clearing from the timeout handler so the queue stays blocked until Python's late response arrives, specifically to prevent response misattribution. Your change makes the busy state maximally fatal; that one makes it potentially permanent. Landed together and unmodified, one hung command would turn every subsequent call into an immediate hard error, forever. Both PRs actually converge on the same missing primitive — a bounded timer — which is why I would rather land one coherent change than two opposed ones.
2. **The timeout code has moved.** #335 extracted the per-command timeout policy out of `src/server.ts` into a new `src/command-timeout.ts` as a pure function (so it is unit-testable), while fixing #251 where `autoroute` was abandoned at 30 s. Your rebase will need to account for that, and it is probably the natural home for a bounded-wait constant too.

If you have moved on, no problem — say so and I will close this with #274/#277/#278 as partially-landed. The diagnosis in this PR was correct even if the remedy needs softening.

mixelpixx:

```bash
git fetch origin
```

mixelpixx:
Closing with thanks — the problem this PR diagnosed turned out to be bigger than its fix, and it is now solved at the level below. You were right that overlapping operations were dangerous; the mechanism was worse than a hang: the ID-free single-slot bridge could resolve request B with request A's late response, silently desyncing everything after (see #373). #382 fixes that with end-to-end request correlation — overlap now queues and every response is identified, so the fail-fast this PR proposed is no longer needed and would reject calls the queue handles fine. Your diagnosis contributed to getting there.

PR #277, opened 2026-07-01
Fix get_board_2d_view: pass --layers and rasterize inline PNG via cairosvg

### LINKED ISSUE #235: get_board_2d_view: KiCad 10 requires --layers; pcbPath fallback broken

stefangordon:
## Problem

`get_board_2d_view` fails to return an image, for two independent reasons — both must be fixed for the tool to work:

1. **Export flag.** KiCad 9+ `pcb export svg` fails with *"At least one layer must be specified"* (no file written) unless `--layers` is passed, but the layers list defaulted to `[]` and the flag was only appended when non-empty.

2. **Inline PNG conversion.** `board/view.py` had its own copy of `_svg_to_png` that tried pymupdf → inkscape → ImageMagick — none of which are declared dependencies. On a stock install (only `cairosvg`, which *is* in `requirements.txt`) all converters failed, it returned `None`, and inline PNG mode failed with a misleading *"kicad-cli SVG export failed"* even though the plot itself succeeded (file/SVG mode worked). Same root cause as the schematic render path.

## Fix

1. Default `layers` to a readable 2D view (`F.Cu, B.Cu, F.SilkS, B.SilkS, Edge.Cuts`) when the caller specifies none; explicit layers still override.
2. Use `cairosvg` as the primary SVG→PNG converter (pymupdf/inkscape/convert remain fallbacks).

## Tests

- `--layers` is always present in the export command.
- `_svg_to_png` returns a PNG at the requested dimensions.

Verified end-to-end on a real board: inline PNG produced (~0.03 MB).

Fixes #235. Related: #209.

### DIFF
--- a/python/commands/board/view.py
+++ b/python/commands/board/view.py
@@ -170,7 +170,17 @@ class BoardViewCommands:
                     "message": f"Unsupported format '{fmt}'. Use 'png', 'jpg', or 'svg'.",
                     "errorDetails": f"Got: {fmt}",
                 }
-            layers: List[str] = params.get("layers", [])
+            # `pcb export svg` requires at least one layer on KiCad 9+ — omitting
+            # it fails with "At least one layer must be specified" and no output.
+            # Default to a readable 2D view (copper + silkscreen + board outline)
+            # when the caller doesn't specify layers.
+            layers: List[str] = params.get("layers") or [
+                "F.Cu",
+                "B.Cu",
+                "F.SilkS",
+                "B.SilkS",
+                "Edge.Cuts",
+            ]
             response_mode = params.get("responseMode", "inline")
 
             kicad_cli = shutil.which("kicad-cli") or shutil.which("kicad-cli.exe")
--- a/python/commands/board/view.py
+++ b/python/commands/board/view.py
@@ -15,16 +15,33 @@ logger = logging.getLogger("kicad_interface")
 
 
 def _svg_to_png(svg_path: str, width: int, height: int) -> Optional[bytes]:
-    """Convert SVG to PNG. No cffi dependency.
+    """Convert SVG to PNG at the requested pixel dimensions.
 
     Priority:
-      1. pymupdf (fitz) — bundled MuPDF renderer, pure Python, no system deps
-      2. Inkscape CLI — accurate KiCAD SVG rendering
-      3. ImageMagick convert — broad availability fallback
+      1. cairosvg — the declared dependency (see requirements.txt); honors
+         output_width/output_height and reliably rasterizes KiCAD SVGs.
+      2. pymupdf (fitz) — bundled MuPDF renderer, if installed.
+      3. Inkscape CLI — accurate KiCAD SVG rendering, if installed.
+      4. ImageMagick convert — broad availability fallback.
     Returns PNG bytes or None if all converters fail.
+
+    Without a working converter, get_board_2d_view's inline PNG mode fails
+    (historically surfaced as a misleading "kicad-cli SVG export failed"),
+    even though the kicad-cli plot itself succeeded — file/SVG mode works.
     """
     import subprocess
     import tempfile
+
+    try:
+        import cairosvg
+
+        return cairosvg.svg2png(
+            url=svg_path,
+            output_width=int(width),
+            output_height=int(height),
+        )
+    except Exception:
+        pass
 
     try:
         import fitz
--- a/tests/test_get_board_2d_view_save_to_file.py
+++ b/tests/test_get_board_2d_view_save_to_file.py
@@ -227,3 +227,29 @@ def test_kicad10_export_cmd_uses_file_output_and_mode_single(tmp_path):
     out_idx = argv.index("--output")
     out_arg = argv[out_idx + 1]
     assert out_arg.endswith(".svg")  # a file path, not a directory
+
+
+# ---------------------------------------------------------------------------
+# --layers regression (KiCad 9+ requires at least one layer)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.unit
+def test_board_svg_export_always_passes_layers(tmp_path):
+    """`pcb export svg` must receive --layers with a non-empty spec even when the
+    caller specifies no layers. KiCad 9+ fails with "At least one layer must be
+    specified" (and writes no file) otherwise."""
+    cmd, root, board_path = _make_view_cmd(tmp_path)
+    fake_result = MagicMock(returncode=0, stderr="", stdout="")
+
+    with (
+        patch("shutil.which", return_value="/usr/bin/kicad-cli"),
+        patch("subprocess.run", return_value=fake_result) as run,
+    ):
+        cmd.get_board_2d_view({"pcbPath": str(board_path), "format": "svg"})
+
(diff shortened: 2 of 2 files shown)


mixelpixx (changes requested):
Part 1 (the --layers default) is verified and ready: all nine layer-flag tests pass here on top of current main, the default layer set is sensible, and explicit layers still override. That alone fixes #235's headline failure.

Part 2 has the same cross-platform problem I just wrote up in detail on #274: the cairosvg pip package needs a native cairo runtime that stock Windows does not have, so on this Windows machine test_svg_to_png_converts_via_cairosvg fails (cannot load library 'libcairo-2.dll') and the converter falls through to the same undeclared chain this PR is replacing. It works on Linux because distros ship libcairo2 — which our Linux-only CI would happily confirm while Windows users stay broken.

Same suggestion as there: declare pymupdf in requirements.txt as the first-priority converter (binary wheels on all three platforms, no system deps) and keep cairosvg as the second choice. Since both PRs share this converter decision, whichever lands first should carry the requirements.txt change and the other can rebase on it.

If you'd rather get #235 closed quickly, splitting the --layers fix out would merge today — but given the small diff I'm fine holding one PR for the converter swap too. Your call.

mixelpixx:
Nudging this one because it has become the blocker for a second PR, and the fix on your side looks like it is about one command.

Recap of where the review left it on 07-03: **Part 1 (the `--layers` default) is verified and approved.** Part 2 (cairosvg as the primary rasterizer) is the blocked half — `cairosvg` needs a native cairo DLL that is not present on a stock Windows install, so making it priority 1 in the converter chain moves the failure rather than fixing it. The suggestion there was to keep pymupdf as the primary (it bundles its own renderer and needs no system libraries) and treat cairosvg as a fallback.

**Why this is newly urgent: #324 independently implemented the same `--layers` fix.** It replaces the identical expression on the identical line (`view.py:170`), `params.get("layers", [])` becomes `params.get("layers") or [...]`. The only difference is the default list — yours is `F.Cu, B.Cu, F.SilkS, B.SilkS, Edge.Cuts`, theirs omits `B.SilkS`. Merging the two branches against each other produces exactly one conflict, that hunk.

We have asked #324 to drop its copy, because yours is the one that was reviewed and yours is the one that carries the nine layer-flag regression tests. But that leaves #324 (a genuine `.kicad_sch` corruption fix, otherwise green) waiting on this PR to actually land the layers half.

**The good news is that your commits already split cleanly along exactly the right line:**

```
aa5875d  get_board_2d_view: always pass --layers to `pcb export svg`      <- approved half
5aeb716  get_board_2d_view: rasterize inline PNG via cairosvg (...)       <- blocked half
```

So dropping `5aeb716` leaves precisely the approved change, with its tests, and nothing else. If you push that, this can merge essentially immediately and #324 unblocks behind it. The cairosvg work then comes back as its own PR with the pymupdf-primary ordering, reviewed on its own merits.

Two practical notes:

1. **No rebase is actually required for conflicts** — I checked, and the branch still merges cleanly into current `main` despite being 168 commits behind. So do not feel obliged to untangle a rebase; a refresh is nice-to-have, not a gate.
2. One thing a refresh *would* pick up: `main` now resolves the CLI through `resolve_kicad_cli()` rather than the inline `shutil.which("kicad-cli") or shutil.which("kicad-cli.exe")` your branch was written against. Git merges it cleanly because your diff does not touch that line, so this is informational rather than a problem.

If you would rather not do the split yourself, say the word and we can cherry-pick `aa5875d` into a small PR with you credited as author, the same way #215 was carved out of #182 — happy to take that on rather than let a reviewed fix sit.

mixelpixx:
Heads-up: I have cherry-picked your `aa5875d` (`--layers`) into #337 with your authorship preserved on the commit. It has been sitting approved-in-substance since 2026-07-03 and was blocking two other things — issue #235, and PR #324, which independently carries the same edit as a drive-by. Yours is the better version (it includes `B.SilkS` and comes with the regression test), so that is the one being landed.

I verified the underlying behaviour against real KiCad 10.0.0 first: without `--layers`, `pcb export svg` prints "At least one layer must be specified" and writes no file at all. So this was silently producing nothing whenever the caller omitted layers — worth fixing regardless of the rest.

**This PR stays open for the second commit.** The cairosvg objection from the 07-03 review is unchanged: `5aeb716` inserts cairosvg as priority 1 ahead of the existing pymupdf/fitz chain, and cairo's native DLL is absent on a stock Windows install. `tests/test_cairo_dll_preload_windows.py` exists on main because this has already bitten the project once.

If you want that half to land, the smallest change is to keep pymupdf primary and add cairosvg as a *fallback* rather than ahead of it — your new test asserts a converter is available without asserting which, so it should pass unchanged under that ordering. Rebase onto current main (which now has real CI — see #334) and I will review promptly.

If you would rather not, no problem: say so and I will close #277 as partially-landed. Same question applies to #274, #275 and #278, which have all been quiet since 07-01 — happy to keep any of them open if you plan to return to them.

mixelpixx:

```bash
git fetch origin
```

mixelpixx:
Closing with thanks — both halves of this PR are now landed or resolved elsewhere: part 1 (the --layers fix) was cherry-picked into #337 with your authorship preserved back in July, and the rasterization question is settled by #384 (pymupdf declared first, cairosvg second — see the review discussion on #274 for why). Nothing of value here was dropped.

PR #278, opened 2026-07-01
Inherit the library symbol's Footprint when placing a schematic instance

stefangordon:
## Problem

A footprint set on a library symbol (e.g. via `create_symbol`) doesn't propagate to placed instances. `create_component_instance` wrote the instance's `Footprint` property straight from its `footprint` argument (default `""`) and never consulted the library symbol, so placements stayed unfootprinted until an explicit `edit_schematic_component` set the field.

KiCad itself inherits the symbol's `Footprint` onto the instance at placement time.

## Fix

When the caller doesn't pass a footprint, default it to the library symbol's own `Footprint` value, read from the already-injected `lib_symbols` block. Explicit footprint arguments still override.

## Tests

`tests/test_symbol_footprint_inherit.py`:
- placement with no footprint inherits the library symbol's value
- an explicit footprint argument still overrides

### DIFF
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -529,6 +529,35 @@ class DynamicSymbolLoader:
         ry = -dx * math.sin(rad) + dy * math.cos(rad)
         return round(rx, 3), round(ry, 3)
 
+    def _extract_lib_property_value(
+        self, schematic_path: Path, library_name: str, symbol_name: str, prop_name: str
+    ) -> str:
+        """Return a property's VALUE from the injected lib_symbols definition.
+
+        Used to inherit fields (e.g. Footprint) from the library symbol onto a
+        placed instance, matching KiCad's own placement behavior. Returns "" if
+        the symbol or property isn't found.
+        """
+        try:
+            import re
+
+            with open(schematic_path, encoding="utf-8") as f:
+                content = f.read()
+            lib_start = content.find("(lib_symbols")
+            if lib_start == -1:
+                return ""
+            sym_start = content.find(f'(symbol "{library_name}:{symbol_name}"', lib_start)
+            if sym_start == -1:
+                return ""
+            sym_block = self._extract_paren_block(content, sym_start)
+            m = re.search(
+                r'\(property\s+"' + re.escape(prop_name) + r'"\s+"([^"]*)"',
+                sym_block,
+            )
+            return m.group(1) if m else ""
+        except Exception:
+            return ""
+
     def create_component_instance(
         self,
         schematic_path: Path,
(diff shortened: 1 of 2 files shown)


mixelpixx:
Heads up: #265 merged today and rewrites `create_component_instance` in `dynamic_symbol_loader.py` (complete KiCad 10 instance blocks — real project name, real sheet-uuid path, per-pin uuids, canonical field order). Your footprint-inherit change touches the same function, so this branch will need a rebase.

The feature itself looks complementary rather than conflicting — inheriting the library symbol's Footprint property is orthogonal to the instance-block structure — so after rebasing, the property-value plumbing should slot into the new writer cleanly (properties are emitted around line 805-830 of the rewritten file). One thing to double-check after the rebase: #265's writer normalizes the Datasheet property from "~" to an empty string and adds a Description property, so if your test asserts on absolute property indices or counts, update it to match the new shape.

mixelpixx:
Ping — no activity since 2026-07-01, and this one is blocking a filed bug.

Issue #300 (footprint left empty on placed instances even when the library symbol has a default) is **exactly what this PR implements**, and it is still live on main — `dynamic_symbol_loader.py` continues to emit the caller's `footprint` verbatim, with `schematic_handlers.py` defaulting it to `""` and no library lookup. So there is a user waiting on this.

The blocker is mechanical: the branch is CONFLICTING. It was authored before #265 rewrote `create_component_instance` into a full KiCad 10 instance writer, so the function you are patching no longer looks the same. That is the whole ask — rebase onto current main and re-point the change.

Worth knowing before you do: `dynamic_symbol_loader.py` has since become the busiest file in the repo. #324 is in flight against its property-writing path, and I filed #336 for an escape-awareness bug in `_extract_lib_property_value`. If you rebase and something looks like it moved for no reason, that is why.

If you have moved on, that is fine — tell me (or don't) and I will either close this as partially-landed or take the fix myself against #300, crediting you for the diagnosis. I would rather not leave a filed bug blocked on a quiet branch indefinitely, so if I hear nothing I will pick it up.

Also, for the record: I landed the `--layers` half of your #277 in #337 with your authorship preserved — it was blocking #235 and PR #324.

mixelpixx:
Heads-up: this branch already needed a rebase, and **#339 has now added a second, unrelated source of conflict** — a repo-wide formatting normalization (black, isort and prettier over `tests/` and `src/tools/`, which the old `black --check python/` gate never covered).

Flagging it so the two do not get confused with each other when you rebase:

- **Formatting conflicts** are noise. Take either side of the hunk and run `npm run format:py` (plus `npx prettier --write .` if it is in `src/`). I verified the normalization is semantically inert before merging — every changed Python file AST-compared against the prior commit, giving 0 real changes, and an identical test result (1551 passed).
- **Substantive conflicts** are the pre-existing ones from main moving under this branch, and those are the ones worth reading carefully.

If a formatting-looking conflict turns out to be more than whitespace, tell me — that would mean I got something wrong in #339.

Also relevant to your next push: **CI now actually enforces formatting** via `pre-commit run --all-files` in the `code-quality` job. Running `pre-commit install` once (or `npm run format:py` before pushing) keeps you green. Worth knowing that until today CI's Python side was a complete no-op and Actions was disabled repo-wide — so if this PR ever felt like it was held to a standard nothing was checking, that is because it was.

Your outstanding review items are unchanged; nothing in #339 touches the substance here.

mixelpixx:
Closing in favor of #381, which lands your fix with credit (Co-authored-by, and your test file kept as written). As announced on #300 on 07-26, I said I would take this over if the branch stayed quiet — it has been silent since 07-01 and the grace period expired 08-02.

One substantive note on why #381 is not a straight rebase of this branch: your _extract_lib_property_value predates #348 and uses the non-escape-aware quoted-value regex, which would now regress the escaping fixes — #381 takes only the 4-line inheritance hook and routes it through the current escape-aware extractor and (via #354) the escaping writer, with a new test pinning that interaction. The diagnosis and the shape of the fix are yours; thank you for both, and for #274/#275/#277 besides — this repo got real value from all four.

PR #280, opened 2026-07-01
Bound IPC connect so a busy KiCad falls back to SWIG instead of hanging

stefangordon:
## Problem

kipy dials with `pynng.Req0(dial=..., block_on_dial=True)`, and `block_on_dial=True` has **no timeout** (only send/recv are capped, at 2s). When KiCad's GUI thread is busy and not servicing its socket — e.g. auto-reloading a library/table the MCP just wrote, or any modal/progress state — `IPCBackend.connect()` can block until the MCP client's ~240s watchdog fires. To the caller it looks like a frozen server; only afterward does the session fall back to SWIG.

## Fix

Run each kipy open+ping in a daemon thread with a wall-clock join timeout (`KICAD_IPC_CONNECT_TIMEOUT`, default 5s). A stuck attempt is abandoned (the daemon thread dies with the process) and `connect()` fails fast, so callers fall back to the SWIG/file backend in seconds instead of ~240s. A healthy connect is unaffected (sub-second).

## Tests

`tests/test_ipc_connect_timeout.py`:
- a simulated stuck dial is bounded to the timeout (not the 30s sleep)
- a healthy connection returns fast
- a fast failure propagates unchanged

## Related

#248 / #251 (operations that appear to hang). This bounds the IPC *connect* path specifically; it's complementary to serializing/queue behavior.

### DIFF
--- a/python/kicad_api/ipc_backend.py
+++ b/python/kicad_api/ipc_backend.py
@@ -17,12 +17,22 @@ Key Benefits over SWIG:
 import logging
 import os
 import platform
+import threading
+import time
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional
 
 from kicad_api.base import APINotAvailableError, BoardAPI, ConnectionError, KiCADBackend
 
 logger = logging.getLogger(__name__)
+
+# Hard wall-clock cap for a single IPC connection attempt. kipy dials with
+# pynng's block_on_dial=True, which has NO timeout (only send/recv are capped),
+# so if KiCad's GUI thread is busy (e.g. reloading a library we just wrote to,
+# or any modal/progress state) and not servicing its socket, a connect can hang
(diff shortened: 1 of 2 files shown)


mixelpixx (approved):
The daemon-thread bound is the right call given pynng's block_on_dial has no timeout, and the env-overridable cap with SWIG fallback matches how the rest of the backend degrades. Verified locally on top of current main (including #276, which auto-merges cleanly in the neighboring region): your timeout tests plus the backend pinning and read-fallback suites all pass, black and mypy clean.

One nit worth a small follow-up, not blocking: if a stuck dial eventually completes after the join gave up, the orphaned worker finishes creating a live KiCad object whose socket nobody owns — which is exactly the fd leak #276 just closed for the disconnect path. Having the worker close the connection when it discovers it lost the race (e.g. check a 'gave_up' flag before storing the result, close otherwise) would make the two fixes airtight together. Merging.

PR #283, opened 2026-07-02
fix(schematic): drag net labels with a moved/rotated pin to prevent silent re-netting

ravishivt:
## Summary

When `move_schematic_component` / `rotate_schematic_component` drag connected
wires to follow a symbol's pins, a net label sitting exactly on a moved pin is
now carried along with it — fixing a bug where the label was left behind and
could **silently re-net** onto a neighbouring signal.

## Why

`WireDragger.drag_wires()` moved wire endpoints and junctions to follow rotated/
moved pins, but not net labels. A `label` / `global_label` / `hierarchical_label`
coincident with a pin was left at the old coordinate while the pin's wiring moved
away. The label then detached from the pin's net and, if another net's wire
occupied that coordinate, silently merged onto it.

Observed: rotating a diode moved a drain test-point label onto the **+12V** rail.
This is a connectivity-correctness bug — the change is invisible unless you diff
the netlist, which is exactly what makes it dangerous.

## Changes

- **`wire_dragger.py`**: `drag_wires()` gains a third pass that drags
  `label` / `global_label` / `hierarchical_label` sitting on a moved pin, using
  the same coordinate-match tolerance already used for wire endpoints and
  junctions. Returns `labels_moved`.
- **`schematic_handlers.py`**: move and rotate handlers report `labelsMoved`.
- **`src/tools/schematic.ts`**: the move/rotate tool responses surface how many
  labels were relocated, so the caller is told when a label moved. Also clarifies
  the rotate tool's `angle` as an **absolute** target (final orientation), not a
  relative increment like KiCad's UI `R` key.

## Test plan

- `pytest tests/test_wire_dragger_labels.py` — 4 passing: a label follows a moved
  pin; global/hierarchical labels also move; a label *not* on a pin is untouched;
  no false positives when no labels are present.
- Existing `test_rotate_schematic_mirror.py`, `test_ipc_rotate_absolute.py` pass.
- flake8 + isort clean. (black must be run via a Python ≥3.10 pre-commit env.)

Independent of the canonical-formatting fix (separate PR); the two branches touch
different regions of `wire_dragger.py` / `schematic_handlers.py` and merge cleanly.

### DIFF
--- a/python/commands/schematic_handlers.py
+++ b/python/commands/schematic_handlers.py
@@ -1684,6 +1684,7 @@ class SchematicHandlersMixin:
                 "wiresMoved": drag_summary.get("endpoints_moved", 0),
                 "wiresRemoved": drag_summary.get("wires_removed", 0),
                 "wiresSynthesized": drag_summary.get("wires_synthesized", 0),
+                "labelsMoved": drag_summary.get("labels_moved", 0),
             }
 
         except Exception as e:
@@ -1767,6 +1768,7 @@ class SchematicHandlersMixin:
                 "mirror": effective_mirror,
                 "wiresMoved": drag_summary.get("endpoints_moved", 0),
                 "wiresRemoved": drag_summary.get("wires_removed", 0),
+                "labelsMoved": drag_summary.get("labels_moved", 0),
             }
 
         except Exception as e:
--- a/python/commands/wire_dragger.py
+++ b/python/commands/wire_dragger.py
@@ -27,6 +27,9 @@ _K = {
         "xy",
         "wire",
         "junction",
+        "label",
+        "global_label",
+        "hierarchical_label",
         "property",
         "stroke",
         "width",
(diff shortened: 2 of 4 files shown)


mixelpixx (approved):
Silent re-netting is the worst kind of schematic bug — invisible until the netlist diverges — and the fix is scoped exactly right: a third drag pass reusing the same coordinate-match tolerance as the endpoint and junction passes, so a label only moves when it sat on a pin that moved. The int-normalization of rotation angles (90 not 90.0) also composes well with #284's byte-canonical goal.

Verified locally on the merged tree (your branch auto-merged clean with today's #284): your 4 label tests, the rotate/mirror and IPC-rotate suites, and the eeschema netlist-oracle suite all pass; full suite matches the known-environmental baseline; black, mypy, tsc, vitest all clean.

One investigation note that is NOT about this PR: while verifying I found a pre-existing, order-dependent test-pollution bug — running test_move_with_wire_preservation.py together with test_pin_world_xy_eeschema_truth.py in one pytest invocation makes 20 tests fail (identically on untouched main; each file passes alone and the full suite dodges the ordering). If you run subsets locally and see the oracle tests fail, that's why. I'll file it separately. Merging.

PR #284, opened 2026-07-02
fix(schematic): write .kicad_sch in KiCad's canonical format, not minified

ravishivt:
## Summary

Schematic write tools now emit `.kicad_sch` in KiCad's canonical multi-line
("pretty") format instead of a single minified line, so tool edits produce
minimal, reviewable diffs that match what eeschema's **Save** writes byte-for-byte.

## Why

Every schematic write tool parses the file with `sexpdata` and serialized it back
with `sexpdata.dumps()`, which emits the **entire file as one line**. A one-symbol
edit therefore rewrote a whole sub-sheet onto a single line:

- `rotate_schematic_component` on one diode turned an untouched 12,951-line
  sub-sheet into **1 line** — `git diff` showed ~12,951 deletions / 1 insertion,
  impossible to review.
- KiCad's UI always re-expands to the multi-line canonical format on save, so a
  minified file flips back and forth on every manual edit — perpetual whole-file
  churn and merge pain.

Keeping KiCad-pretty (not embracing minification) is the only option that doesn't
fight the editor: you can't stop eeschema from writing multi-line.

## Changes

- **`python/utils/sexpr_format.py`** (new): a port of KiCad's `Prettify()`
  (`FORMAT_MODE::DEFAULT`, the mode eeschema uses for `.kicad_sch`) plus
  `dumps()`, a drop-in for `sexpdata.dumps()`. It re-parses its own output and
  falls back to the compact form on any round-trip mismatch, so a formatter
  change can never corrupt schematic data.
- **`wire_manager.py`, `schematic_handlers.py`, `schematic_snap.py`,
  `schematic_batch.py`**: every `sexpdata.dumps()` write site now calls
  `kicad_dumps`.
- **`schematic.py`**: `save_schematic()` reformats kicad-skip's output through
  `prettify()` (kicad-skip emits a semi-minified layout).
- **`wire_dragger.py`**: normalize integral rotation angles to `int` (`90`, not
  `90.0`) to match KiCad's output exactly.
- **`scripts/kicad_sch_reformat.py`** (new): re-expand files minified by older
  versions — in-place, `--check` (CI/pre-commit), or `--stdin` (git clean
  filter). Reuses the same formatter, so results match tool writes.
- **`tests/`**: `test_sexpr_format.py` covers a synthetic golden, xy-run wrapping,
  escaped quotes/backslashes, idempotency and data round-trip; plus a real
  `kicad-cli`-produced fixture (`canonical_schematic.kicad_sch`) with a full
  `lib_symbols` def and a placed symbol `(instances (project ...))` block.

## Compatibility

Not a breaking change:

- **Data / semantics are unchanged.** The formatter only moves whitespace outside
  quoted strings; the parsed s-expression is identical, and KiCad opens minified
  and pretty files the same way. Unlike `kicad-cli sch upgrade`, it never touches
  `(version ...)` or project names — it is pure whitespace, so no schema/version
  upgrade is triggered.
- **No API change.** No tools or params are renamed/removed.
- **Nothing reformats spontaneously.** A file is only rewritten when a write tool
  actually edits it — exactly when the old code would have minified it anyway.
  There is no repo-wide sweep and no reformat-on-read.

What existing users see on their next tool write:

- Files currently in KiCad's pretty format (the common case): a **minimal** diff
  that matches eeschema, instead of the whole file collapsing to one line — a
  strict improvement, and it ends the UI-expands / tool-minifies churn for anyone
  who edits in both KiCad and these tools.
- Files that were already minified by an older version of this server: a
  **one-time** expansion back to pretty, then stable. That is cosmetic and
  desirable; `scripts/kicad_sch_reformat.py` lets a user do that expansion as a
  deliberate standalone commit rather than have it ride along inside an unrelated
  edit.

Byte-identical output is verified against KiCad 10. `Prettify()` has been
layout-stable across KiCad 7–10, so other versions should match; if a version
ever formats differently the only effect is minor cosmetic churn — never data
changes, since `dumps()` falls back to the compact form on any round-trip
mismatch.

## Test plan

- `pytest tests/test_sexpr_format.py` — 16 passing, including an end-to-end test
  that drives a real write tool (`WireManager.add_wire`) and asserts the file on
  disk is canonical (multi-line, tab-indented, round-trips).
- `dumps(loads(f)) == f` byte-for-byte on the real `kicad-cli` fixture and on
  local project sub-sheets (incl. a 223 KB / 12,951-line one).
- A real `rotate_schematic_component` now yields a ~6-line diff (2 wire endpoints
  + 1 angle) instead of 12,951 deletions / 1 insertion; `kicad-cli sch erc`
  before/after is identical — no connectivity change.
- Existing `test_rotate_schematic_mirror.py`, `test_ipc_rotate_absolute.py` pass.
- flake8 + isort clean. (black must be run via a Python ≥3.10 pre-commit env.)

Independent of the net-label re-netting fix (separate PR); the two branches touch
different regions of `wire_dragger.py` / `schematic_handlers.py` and merge cleanly.

### DIFF
--- a/python/commands/schematic.py
+++ b/python/commands/schematic.py
@@ -5,6 +5,7 @@ import uuid
 from typing import Any, Optional
 
 from skip import Schematic
+from utils.sexpr_format import prettify
 
 logger = logging.getLogger("kicad_interface")
 
@@ -107,6 +108,13 @@ class SchematicManager:
         try:
             # kicad-skip uses write method, not save
             schematic.write(file_path)
+            # kicad-skip emits a semi-minified layout; reformat to KiCad's
+            # canonical pretty format so tool writes match eeschema's "Save"
+            # and produce minimal, reviewable diffs.
+            with open(file_path, "r", encoding="utf-8") as f:
+                content = f.read()
+            with open(file_path, "w", encoding="utf-8") as f:
+                f.write(prettify(content))
             logger.info(f"Saved schematic to: {file_path}")
             return True
         except Exception as e:
--- a/python/commands/schematic_batch.py
+++ b/python/commands/schematic_batch.py
@@ -658,6 +658,7 @@ class SchematicBatchCommands:
         try:
             import sexpdata
             from sexpdata import Symbol
+            from utils.sexpr_format import dumps as kicad_dumps
 
             data = sexpdata.loads(sch_path.read_text(encoding="utf-8"))
             changed = False
@@ -681,7 +682,7 @@ class SchematicBatchCommands:
                         continue
                 new_data.append(item)
             if changed:
-                sch_path.write_text(sexpdata.dumps(new_data), encoding="utf-8")
+                sch_path.write_text(kicad_dumps(new_data), encoding="utf-8")
         except Exception as e:
             logger.warning(f"replace cleanup failed: {e}")
 
--- a/python/commands/schematic_handlers.py
+++ b/python/commands/schematic_handlers.py
@@ -22,10 +22,10 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import pcbnew
 import sexpdata
-
 from commands.library_schematic import LibraryManager as SchematicLibraryManager
 from commands.schematic import SchematicManager
 from commands.wire_manager import WireManager
+from utils.sexpr_format import dumps as kicad_dumps
 
 logger = logging.getLogger("kicad_interface")
 
@@ -1675,7 +1675,7 @@ class SchematicHandlersMixin:
             WireManager.sync_junctions(sch_data)
 
             with open(schematic_path, "w", encoding="utf-8") as f:
-                f.write(sexpdata.dumps(sch_data))
+                f.write(kicad_dumps(sch_data))
 
             return {
                 "success": True,
@@ -1758,7 +1758,7 @@ class SchematicHandlersMixin:
             WireManager.sync_junctions(sch_data)
 
             with open(schematic_path, "w", encoding="utf-8") as f:
-                f.write(_sexpdata.dumps(sch_data))
+                f.write(kicad_dumps(sch_data))
 
             return {
                 "success": True,
@@ -1989,7 +1989,7 @@ class SchematicHandlersMixin:
                 item[at_idx] = [_SYM_AT, float(new_x), float(new_y), rotation]
 
                 with open(schematic_path, "w", encoding="utf-8") as f:
-                    f.write(_sexpdata.dumps(sch_data))
+                    f.write(kicad_dumps(sch_data))
 
                 return {
                     "success": True,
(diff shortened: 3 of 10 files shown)


mixelpixx (approved):
This closes the second half of #256 (the first half — broken instance data — landed in #265 yesterday), and the design earns the trust a schematic-writer change needs. What I verified, beyond reading the port line by line:

1. **The safety valve works and is the right shape.** `dumps()` re-parses its own pretty output and byte-compares against the compact form, falling back to the (ugly but exact) compact string on any mismatch. That makes formatter-induced corruption structurally impossible at every swapped write site — the failure mode degrades to the status quo, never to data loss. This is the same degrade-not-corrupt principle #265 followed, and it is exactly what code that writes into users' schematics should do.
2. **Independent end-to-end on real KiCad 10.0.0 (this machine):** took a real, loadable schematic, minified it through `sexpdata` (what the old writers produced), restored it through your `dumps()` — the output loads in `kicad-cli` and passes ERC with zero errors. Also confirmed the kicad-cli fixture restores byte-identically after a minify round-trip, and that `prettify` is idempotent.
3. **Blast radius:** all 208 tests across the wire/schematic write paths (wire preservation, junctions, connectivity, hierarchy, add-component, and the #265 instance round-trip) pass on the merged tree; full suite matches the known-environmental baseline; mypy clean.

Your branch had a one-line import collision with yesterday's #267 merge; since maintainer edits are allowed I pushed the trivial resolution (kept both imports) rather than bouncing it back.

Three follow-ups, none blocking:

1. **The fixture doesn't actually load in KiCad.** `kicad-cli sch erc tests/fixtures/canonical_schematic.kicad_sch` reports "Failed to load schematic" on KiCad 10.0.0 here (not a path artifact — a prettified real schematic loads fine the same way, and adding `embedded_fonts` doesn't fix it). Its role in the tests is byte-layout ground truth, so this doesn't undermine the suite, but a canonical fixture that KiCad itself rejects will confuse the next person who reaches for it. Worth regenerating from a loadable design, and ideally adding a kicad-cli-gated loadability assertion like the one #265's tests use.
2. **`save_schematic` calls `prettify()` without the round-trip guard** that `dumps()` has (`python/commands/schematic.py`). The input there comes from kicad-skip rather than sexpdata, but the equivalent insurance is cheap: parse both texts with `sexpdata.loads` and compare `sexpdata.dumps` of each, keep kicad-skip's original on mismatch. That would extend the corruption-impossible property to the one write site that currently trusts the formatter unconditionally.
3. `black scripts/kicad_sch_reformat.py` — it's outside CI's `black --check python/` scope, but keeping it formatted spares the next contributor a surprise when pre-commit touches it.

Merging. Thanks for the reformat script too — that gives users with already-minified files from older versions a clean repair path, which #256's reporters will want.

PR #291, opened 2026-07-03
feat: add update_symbol_from_library MCP tool

karu2003:
## Summary

Adds MCP tool **update_symbol_from_library** — programmatic equivalent of KiCad's Update Symbol from Library. Refreshes embedded lib_symbols cache entries in .kicad_sch files from a shared .kicad_sym library.

Reuses DynamicSymbolLoader for library parsing. Supports single file, explicit path list, or batch over projectsDir/*/*.kicad_sch.

## Motivation

When a shared symbol library is updated (footprint fields, pin layout, graphics), placed instances in schematics keep stale lib_symbols copies until manually updated in KiCad. For multi-sheet / multi-project workflows this is tedious and error-prone. This tool automates the refresh at scale.

## Behaviour / edge cases

| Case | Handling |
|------|----------|
| Power symbols (GND, +3V3, …) | Library uses (power) wrapper; schematic expects inline (on_board yes) (power) flag. _adapt_library_block_for_schematic flattens the wrapper. |
| Mirror-cache entries (__m0, __m90, …) | KiCad-generated sub-symbols — skipped, must not be overwritten from the base library. |
| Missing lib_symbols entry | Injected at end of (lib_symbols …) if (lib_id …) references the symbol. |
| Optional repair | repairMirrorFromBackup + backupDir restores corrupted __m* blocks from a pre-update backup. |

## Tool parameters

- schematicPath / schematicPaths / projectsDir — target schematics (one required)
- libraryName — required sym-lib-table nickname (e.g. Device, project_lib)
- symbols — optional list to limit which symbol names are updated
- backupDir + repairMirrorFromBackup — optional mirror-block recovery

## Test plan

- [x] npm run build passes locally
- [ ] Call update_symbol_from_library with libraryName on a schematic backed by a shared .kicad_sym
- [ ] Power symbols load without parse errors
- [ ] Mirror symbols (__m*) unchanged
- [ ] Open updated schematic in KiCad 9/10; run ERC
- [ ] kicad-cli sch export pdf succeeds on updated schematic


karu2003:
@mixelpixx — ready for review when you have time.

This is a small, focused addition (one Python module + wiring). Branch is rebased on current `main` (`4f9828b`), single commit, `npm run build` green locally.

**Real-world validation:** used on a 10-schematic project with a shared `XXX_components.kicad_sym` library. Two pitfalls surfaced during development and are handled in the tool:

1. **Power symbols** — copying `.kicad_sym` blocks verbatim breaks schematic parsing (`(power) { … }` wrapper). The adapter flattens to inline `(power)` flag.
2. **Mirror-cache symbols** (`__m0`, `__m90`, …) — must be left alone; overwriting them from the base library also breaks parsing. These are skipped; optional `repairMirrorFromBackup` can restore from a pre-update backup if needed.

Happy to add unit tests with a minimal `.kicad_sch` + `.kicad_sym` fixture if you'd like that before merge — the logic is pure file/text manipulation and testable without a live KiCad session.

mixelpixx (changes requested):
Fit first, since we now gate on it explicitly: this one passes easily — it is the programmatic equivalent of KiCad's own Update Symbol from Library, the stale-embedded-symbol problem is real for any shared-library workflow, and reusing DynamicSymbolLoader keeps the parsing surface single. Happy to carry it. Two must-fixes before merge, both of which will feel familiar from #261:

1. **No Python schema.** update_symbol_from_library is registered in the TS layer (schematic-batch.ts) but appears nowhere in python/schemas/tool_schemas.py, so the Python server's own tools/list falls through to the no-schema branch and logs a warning at runtime — the exact parity gap we fixed on #261's declutter tool. Please add the entry (port it from your TS definition) to the appropriate *_TOOLS list.

2. **No tests for ~200 lines of paren-matched text surgery on user schematics.** The helpers are pure functions on strings and very testable. The critical invariant to lock in: refreshing a lib_symbols definition must not disturb the placed instances — in particular the per-pin (pin "N" (uuid ...)) entries and (instances ...) blocks that #265 introduced, since those are what keep ERC and netlists working. A test that places a symbol via DynamicSymbolLoader.add_component, refreshes its library definition through your tool, and asserts the instance block is byte-identical would cover the scariest failure mode. Please also include a power-symbol case since the (power) wrapper flattening is bespoke logic.

One nit: the modified schematic is written back with plain write_text of the spliced content. Since #284, tool writes should match KiCad's canonical layout — run the final content through utils.sexpr_format.prettify() (cheap, and idempotent on already-canonical files) so a refresh does not introduce layout churn against eeschema saves.

Re-review will be quick once those are in; the batch-over-projectsDir design and the mirror-block repair logic both read well.

mixelpixx:
Yes -- please add the tests. The three cases that matter most:

1. A refresh preserves placed-instance blocks (especially the per-pin `(pin "N" (uuid ...))` entries from #265 -- those are what keep ERC/netlists working after a refresh).
2. A power-symbol refresh does not break schematic parsing (the `(power)` wrapper flattening is your bespoke logic and needs a regression guard).
3. The Python schema entry in tool_schemas.py (item 1 from the review).

Happy to re-review as soon as those land. The branch rebase looks clean and the real-world validation on a 10-schematic project is great evidence.

mixelpixx:
Merged. All three review items landed exactly as asked: the Python schema (tools/list parity restored), the instance-preservation and power-symbol tests (the per-pin uuid fixture is precisely the invariant that matters), and prettify() on both write paths. The 17 tests all pass, plus the TS suite.

Three small things I handled during the merge rather than another round-trip:

1. A mypy error: `results` was inferred as `list[dict[str, str]]` from the first append (the error dict), so the later int-valued entries failed type checking. Annotated it as `list[dict[str, Any]]`.
2. One black reformat on update_symbol_from_library.py.
3. Added the CHANGELOG entry under [Unreleased] -- worth remembering for future PRs, since release notes are compiled from it.

Full suite is at 1330 passed. Requiring `libraryName` instead of a project-specific default was the right call -- thanks for the quick turnaround on the review.

PR #292, opened 2026-07-03
fix: resolve user-defined env vars from kicad_common.json

Alddp:
## Problem

KiCad allows users to define custom path variables in Preferences > Configure Paths, stored in kicad_common.json under environment.vars. Users reference these variables in sym-lib-table / fp-lib-table, for example:

```
(lib (name "seex") (uri "${SEEK}/seex.kicad_sym") ...)
```

The MCP server's `_resolve_uri` method only recognizes built-in KiCad variables (KICAD9_SYMBOL_DIR, KICAD_3RD_PARTY, KIPRJMOD, etc.) and never reads kicad_common.json, causing libraries with custom variables to fail resolution and silently disappear from MCP listings.

## Solution

- Added `load_kicad_env_vars()` static method to `PlatformHelper` to read environment.vars from kicad_common.json
- Called this method in `_resolve_uri` of both `library.py` and `library_symbol.py` to merge user-defined variables
- Cross-platform support: macOS, Linux, Windows KiCad configuration paths

## Tests

Added `tests/test_kicad_common_env_resolution.py` with 4 regression tests:

1. `test_symbol_manager_resolves_user_defined_env_var` - Verifies symbol library manager resolves custom variables
2. `test_library_manager_resolves_user_defined_env_var` - Verifies footprint library manager resolves custom variables
3. `test_sym_lib_table_entry_with_user_var_is_loaded` - End-to-end test for sym-lib-table parsing
4. `test_user_defined_and_builtin_vars_coexist` - Verifies custom and built-in variables coexist

All tests pass with no regressions.

## Related Files

- `python/utils/platform_helper.py` - Added `load_kicad_env_vars()`
- `python/commands/library.py` - Calls `load_kicad_env_vars()`
- `python/commands/library_symbol.py` - Calls `load_kicad_env_vars()`
- `tests/test_kicad_common_env_resolution.py` - New test file

### DIFF
--- a/python/commands/library.py
+++ b/python/commands/library.py
@@ -12,6 +12,8 @@ import re
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
+from utils.platform_helper import PlatformHelper
+
 logger = logging.getLogger("kicad_interface")
 
 
@@ -152,6 +154,9 @@ class LibraryManager:
             "KICAD8_3RD_PARTY": self._find_kicad_3rdparty_dir(),
             "KICAD_3RD_PARTY": self._find_kicad_3rdparty_dir(),
         }
+
+        # Merge user-defined env vars from kicad_common.json
+        env_vars.update(PlatformHelper.load_kicad_env_vars())
 
         # Project directory
         if self.project_path:
--- a/python/commands/library_symbol.py
+++ b/python/commands/library_symbol.py
@@ -13,6 +13,8 @@ from dataclasses import asdict, dataclass
 from pathlib import Path
 from typing import Dict, List, Optional
 
+from utils.platform_helper import PlatformHelper
+
 logger = logging.getLogger("kicad_interface")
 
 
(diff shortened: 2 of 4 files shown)


mixelpixx (approved):
Fit is clear-cut: KiCad's own Preferences > Configure Paths variables are first-class configuration, and libraries registered through them silently vanishing from MCP listings is a real interop hole. The implementation is appropriately conservative — read-only, version-descending search across the three platforms' settings dirs, graceful empty-dict on any parse failure, and merged before KIPRJMOD so the project override still wins.

Verified locally on top of current main: your env-resolution tests plus the roots/paths suites pass (25 tests), mypy clean, full suite at the known-environmental baseline. I pushed two small maintainer edits: the import-line merge collision with #296 (kept both imports in library.py and library_symbol.py) and a black reformat of platform_helper.py.

One follow-up worth a future pass, not blocking: the settings-dir version list (8.0/9.0/10.0) is hardcoded in one more place now — when a KiCad 11 appears, this joins the sweep that #296's kicad_roots helper would naturally own (settings dirs are versioned the same way install roots are). Merging — welcome, and thanks for covering all three platforms in one go.

PR #294, opened 2026-07-03
fix: filter pins by unit in _find_pins_on_net for multi-unit symbols (#293)

### LINKED ISSUE #293: get_wire_connections: phantom cross-unit pins in net membership after #272 fix
## Summary

After PR #272 (correct query coordinates) and PR #259 (rotation/mirror fix), `get_wire_connections` returns the correct query point for each pin of a multi-unit component, but the connected-pins list still includes phantom pins from sibling units.

## Root cause

`wire_connectivity.py`, `_parse_symbol_instances_sexp` (~line 347): the `(unit N)` field is not extracted from placed symbol instances. Downstream in `_find_pins_on_net`, `get_symbol_pins` returns all units' pins (now correctly tagged with their unit by PR #272), but they are all transformed against each placed instance's position without filtering by unit. This creates phantom pin-to-net associations.

## Concrete example

`Amplifier_Operational:LM358`, 3 units, reference IC-1:
```
unit=1  pos=(143.51, 128.27)  rot=0
unit=2  pos=(143.51, 146.05)  rot=0
unit=3  pos=(176.53, 102.87)  rot=270
```

After PR #272 + #259, `get_wire_connections(reference="IC-1", pin="1")` correctly queries (151.13, 128.27) — unit 1's output position. The BFS finds the correct net. But the connected-pins list reports `IC-1/7` (unit 2's output), which belongs to a completely different net.

Full results vs `kicad-cli sch export netlist` ground truth:

| Pin | GT pins | MCP pins | Phantom |
|-----|---------|----------|---------|
| 1 | C1/2, IC-1/1, R23/2 | C1/1, IC-1/1, **IC-1/7**, R23/1 | IC-1/7 |
| 7 | C2/2, IC-1/7, R28, R29 | **IC-1/1**, IC-1/7, R28/2, C2/2, R29/1 | IC-1/1 |
| 3 | D1/1, D2/2, IC-1/3, R22/1 | R22/2, IC-1/3, **IC-1/5**, D2/1, D1/2 | IC-1/5 |
| 5 | D3/1, D4/2, IC-1/5, R26/1 | **IC-1/3**, IC-1/5, D4/2, R26/2, D3/1 | IC-1/3 |
| 8 | REG_VO (many) | REG_VO (many) | ✅ none |
| 4 | IC-1/4, R1-2/2 | IC-1/4, R1-2/1 | ✅ none |

Power pins (8 and 4, unit 3) are clean because unit 3 is placed far away — its pins don't land on any signal-net wires when incorrectly transformed against units 1/2's positions.

## Why signal pins collide

Units 1 and 2 share the same x-coordinate (143.51), differ only in y (128.27 vs 146.05). Output pins are both at relative (+7.62, 0). When unit 1's instance position is used to transform unit 2's output pin: (143.51+7.62, 128.27) = (151.13, 128.27) — lands exactly on unit 1's output wire. BFS follows that wire and incorrectly reports IC-1/7.

## Suggested fix

In `_parse_symbol_instances_sexp`: extract `(unit N)` into each instance dict.

In `_find_pins_on_net`: for each instance, filter `get_symbol_pins` results to only pins where `pin_data["unit"] == instance_unit or pin_data["unit"] == 0` before transforming and testing against wire positions.

The `unit` field is already present on each pin dict after PR #272 — this fix just needs to use it in `wire_connectivity.py` as well.

## Environment

- kicad-mcp v2.3.0 (post PR #272 + #259)
- KiCad 10.0.3, Ubuntu 24.04, Node v20.20.2

zerthimon:
## Summary

Fixes #293 — `get_wire_connections` reported phantom cross-unit pins in its net-membership list for multi-unit components.

`_find_pins_on_net` transformed **every** unit's pins against **every** placed instance of a multi-unit component. When two units share an x-coordinate and their pins have identical library offsets (the normal case for a dual op-amp), a sibling unit's pin lands exactly on this instance's wire and is reported as a false member of the net — e.g. an LM358's pin 7 (unit B output) appearing on unit A's output net alongside pin 1.

This is the net-membership counterpart to #272, which fixed the query-**coordinate** side of the same multi-unit gap. #272 added a per-pin `unit` tag in `parse_symbol_definition`; this PR consumes that tag in `wire_connectivity.py`.

## Root cause

`_parse_symbol_instances_sexp` (in `wire_connectivity.py`) parsed placed symbol instances but did not extract the `(unit N)` field. Downstream, `_find_pins_on_net` fetched all of a symbol's pins via `get_symbol_pins` and transformed them against each instance's position with no unit filtering, so pins from sibling units produced phantom net associations.

## Fix

Two changes, both in `python/commands/wire_connectivity.py`:

1. `_parse_symbol_instances_sexp` now records each instance's `(unit N)` (defaulting to `0`).
2. `_find_pins_on_net` filters the pin definitions to the instance's own unit plus unit 0 (the shared/common body) before transforming them.

Single-unit parts are unaffected: their instances carry no `unit` (or unit 0), which skips the filter entirely and preserves the original behaviour.

## Example (LM358, 3 units)

Real schematic, reference IC-1. Verified against `kicad-cli sch export netlist`:

| Pin | Before (phantom) | After |
|-----|------------------|-------|
| 1 | C1, IC-1/1, **IC-1/7**, R23 | C1, IC-1/1, R23 |
| 7 | **IC-1/1**, IC-1/7, R28, C2, R29 | IC-1/7, R28, C2, R29 |
| 3 | R22, IC-1/3, **IC-1/5**, D2, D1 | R22, IC-1/3, D2, D1 |
| 5 | **IC-1/3**, IC-1/5, D4, R26, D3 | IC-1/5, D4, R26, D3 |
| 8 | (clean) | (clean) |
| 4 | (clean) | (clean) |

Power pins 4/8 (unit 3) were already clean because unit 3 is placed far from units 1/2, so its mis-transformed pins didn't land on their wires.

## Tests

New `tests/test_wire_connectivity_multi_unit.py`: places a 2-unit symbol whose units share x and differ only in y, wires each unit's output to a distinct passive, and asserts each net contains only its own unit's pin — never the sibling's. Both tests **fail without the fix** (verified by stashing the change) and pass with it.

```
pytest tests/test_wire_connectivity_multi_unit.py                    # 2 passed
pytest tests/test_wire_connectivity.py \
       tests/test_pin_locator_multi_unit.py \
       tests/test_pin_locator_and_component.py \
       tests/test_net_connectivity.py \
       tests/test_label_at_pin_net_connections.py                    # 83 passed, no regressions
```

Black, isort, flake8, and mypy all pass on the changed files.

## PR Checklist

- [x] All tests pass locally
- [x] Documentation updated (CHANGELOG.md under [Unreleased])
- [x] No merge conflicts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **`get_wire_connections` no longer reports phantom cross-unit pins** (#293):
+  `_find_pins_on_net` transformed every unit's pins against every placed
+  instance of a multi-unit component, so a sibling unit's pin whose library
+  offset matched could land on this instance's wire and be reported as a
+  false member of the net (e.g. an LM358's pin 7 appearing on unit A's
+  output net alongside pin 1). `_parse_symbol_instances_sexp` now records
+  each instance's `(unit N)`, and `_find_pins_on_net` filters the pin
+  definitions to that unit (plus unit 0, common to all units) before
+  transforming them. Complements #272, which fixed the query-coordinate side
+  of the same multi-unit gap; the pin `unit` tags it added are what this fix
+  consumes. Single-unit parts are unaffected.
+
 - **`create_project` writes a conformant KiCad 10 `.kicad_pro`** (#220): the
   project file was a hand-rolled 122-byte stub containing only
   `board.filename` and a `sheets` entry with the literal id `"root"`, so
--- a/python/commands/wire_connectivity.py
+++ b/python/commands/wire_connectivity.py
@@ -365,6 +365,7 @@ def _parse_symbol_instances_sexp(
             "rotation": 0.0,
             "mirror_x": False,
             "mirror_y": False,
+            "unit": 0,
         }
 
         for sub in item[1:]:
@@ -385,6 +386,8 @@ def _parse_symbol_instances_sexp(
                         inst["mirror_x"] = True
                     elif mv == "y":
                         inst["mirror_y"] = True
+            elif tag == Symbol("unit") and len(sub) >= 2:
+                inst["unit"] = int(sub[1])
             elif tag == Symbol("property") and len(sub) >= 3:
                 prop_name = str(sub[1]).strip('"')
                 if prop_name == "Reference":
@@ -445,6 +448,22 @@ def _find_pins_on_net(
             if not pin_defs:
                 logger.debug(f"  {ref}: no pin definitions for lib_id={lib_id}")
                 continue
+
+            # For a multi-unit component, each placed instance carries a single
+            # (unit N) and only owns the pins defined in that unit's sub-symbol
+            # (plus unit 0, which is common to every unit). Without this filter,
+            # every unit's pins are transformed against every instance's
+            # position, so a sibling unit's pin can land on this instance's wire
+            # and be reported as a phantom member of the net (#293).
+            inst_unit = inst["unit"]
+            if inst_unit:
+                pin_defs = {
+                    num: pdata
+                    for num, pdata in pin_defs.items()
+                    if pdata.get("unit", 0) in (0, inst_unit)
+                }
+                if not pin_defs:
+                    continue
 
             sym_x = inst["x"]
             sym_y = inst["y"]
--- a/tests/test_wire_connectivity_multi_unit.py
+++ b/tests/test_wire_connectivity_multi_unit.py
@@ -1,0 +1,144 @@
+"""
+Regression tests for issue #293:
+
+    get_wire_connections reports phantom cross-unit pins in net membership
+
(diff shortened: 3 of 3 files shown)


mixelpixx (approved):
Fit and correctness both clear: this fixes an owner-tracked bug (#293) in an existing tool, and it applies exactly the unit-filter pattern #272 established on the coordinate side to the net-membership side — the two halves of the same multi-unit gap now use the same rule (pins in unit 0 or the instance's own unit only). Verified locally on top of current main: your multi-unit membership tests plus the wire-connectivity and #272 pin-locator suites all pass (49 tests), black/mypy clean, full suite at the known-environmental baseline. The LM358 false-membership example in the test file is a good permanent repro. I pushed a trivial CHANGELOG conflict resolution (both sides added entries; kept both). Merging — closes #293. Welcome aboard, and thanks for the precise root-cause writeup.

PR #295, opened 2026-07-03
fix: start new projects/schematics blank instead of seeding _TEMPLATE_ symbols (#221 A+C)

### LINKED ISSUE #221: `create_project` writes an outdated `.kicad_sch` (KiCad 9 format) pre-seeded with a `Device:R` symbol

### LINKED ISSUE #243: create_schematic: injects hidden _TEMPLATE_R/C/D instances at off-screen coordinates

boovaragan:
## Summary

Fixes #221 and closes #243 — this is **Part A + C** of the plan agreed in
#221, following the maintainer's merge of Part B (#288).

`create_project` and `create_schematic` copied seeded templates
(`template_with_symbols_expanded.kicad_sch` / `template_with_symbols.kicad_sch`)
that pre-loaded `Device:R/C/LED` `lib_symbols` **and** placed `_TEMPLATE_*`
symbol instances into every new file. Those seeds existed only as clone sources
for the legacy `ComponentManager.add_component` path — the live
`add_schematic_component` tool synthesizes its own `lib_symbols` via the dynamic
loader, and the legacy template fallback was removed in #288 — so the seeds only
leaked stale KiCad 9 content into user projects.

## Changes

**A — repoint both creation paths to a blank KiCad 10 template**

- New `python/templates/blank.kicad_sch`: header
  `(version 20260306) (generator "eeschema") (generator_version "10.0")`,
  empty `lib_symbols`, no placed symbols.
- `create_schematic` (`python/commands/schematic.py`) now copies
  `blank.kicad_sch` instead of `template_with_symbols.kicad_sch`.
- `create_project` (`python/commands/project.py`) now copies `blank.kicad_sch`
  instead of `template_with_symbols_expanded.kicad_sch` (both call sites
  repointed, as requested).

**C — keep the seeded template as a test fixture**

- `template_with_symbols.kicad_sch` is left **unchanged** in-repo; the
  pin-geometry oracle tests load it directly and stay green.

Left in place intentionally (per your guidance): the `_TEMPLATE_`-prefix filters
in the ERC/analysis/layout modules — removing them now would change behavior for
schematics created from the old template.

## Tests

New `tests/test_blank_template_no_seed.py` (3 tests):

- the bundled `blank.kicad_sch` is blank + KiCad 10;
- `create_schematic` produces a file with **no `_TEMPLATE_` references and no
  seeded `lib_symbols`**;
- `create_project` produces a blank schematic likewise.

The assertions run the real `shutil.copy` + file writes against a temp dir
(only the heavy `Schematic` / `pcbnew` deps are stubbed), so they exercise the
actual bytes written to disk. This is the direct regression test for #221 and
#243.

Existing `create_schematic` / `create_project` path + fallback-header tests
still pass, as do the fixture-based add/pin tests (including the #288
mirror-canary tests that seed from `template_with_symbols.kicad_sch`).

## Note

`template_with_symbols_expanded.kicad_sch` now has no runtime consumer (it was
only referenced by the `create_project` call site repointed here). I left it in
the repo rather than delete it in this PR — happy to remove it in a follow-up if
you'd prefer, since it's no longer wired to anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,21 @@ All notable changes to the KiCAD MCP Server project are documented here.
   `add_schematic_component` tool synthesizes its own `lib_symbols` via the
   dynamic loader (and the legacy fallback was removed in #288), so the seeds only
   leaked into user files. Both tools now copy a new blank KiCad 10 template
-  (`python/templates/blank.kicad_sch`: `(version 20260306) (generator
-"eeschema")`, empty `lib_symbols`, no placed symbols).
+  (`python/templates/blank.kicad_sch`: `(version 20260101) (generator
+  "eeschema")`, empty `lib_symbols`, no placed symbols).
   `template_with_symbols.kicad_sch` is kept unchanged in-repo as a test fixture.
   A regression test asserts a created schematic contains no `_TEMPLATE_`
   references and no seeded `lib_symbols` entries.
+
+- **Generated schematics use format version `20260101`, not `20260306`, so
+  every KiCad 10.0.x can open them**: KiCad refuses to load files that claim
+  a format version newer than the running build, and `20260306` is a later
+  10.0.x token — KiCad 10.0.0 (whose `kicad-cli sch upgrade` writes
+  `20260101`) reported "Failed to load schematic" on our output. All
+  templates, fallback writers, and test fixtures now use `20260101`, which
+  loads on every 10.0.x (newer builds silently upgrade on save). This also
+  makes `tests/fixtures/canonical_schematic.kicad_sch` loadable by
+  `kicad-cli`, which it previously was not.
 
 - **`create_project` writes a conformant KiCad 10 `.kicad_pro`** (#220): the
   project file was a hand-rolled 122-byte stub containing only
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **New projects and schematics start blank instead of seeding `_TEMPLATE_*`
+  symbols** (#221, #243): `create_project` and `create_schematic` copied
+  `template_with_symbols_expanded.kicad_sch` / `template_with_symbols.kicad_sch`,
+  which pre-seeded `Device:R/C/LED` `lib_symbols` and placed `_TEMPLATE_*`
+  instances into every new file. Those symbols existed only as clone sources for
+  the legacy `ComponentManager.add_component` path; the live
+  `add_schematic_component` tool synthesizes its own `lib_symbols` via the
+  dynamic loader (and the legacy fallback was removed in #288), so the seeds only
+  leaked into user files. Both tools now copy a new blank KiCad 10 template
+  (`python/templates/blank.kicad_sch`: `(version 20260306) (generator
+"eeschema")`, empty `lib_symbols`, no placed symbols).
+  `template_with_symbols.kicad_sch` is kept unchanged in-repo as a test fixture.
+  A regression test asserts a created schematic contains no `_TEMPLATE_`
+  references and no seeded `lib_symbols` entries.
+
 - **`create_project` writes a conformant KiCad 10 `.kicad_pro`** (#220): the
   project file was a hand-rolled 122-byte stub containing only
   `board.filename` and a `sheets` entry with the literal id `"root"`, so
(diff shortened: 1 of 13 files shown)


mixelpixx (approved):
This completes the plan exactly as agreed in #221 — blank KiCad 10 templates for both creation paths, template_with_symbols.kicad_sch retained untouched as the fixture the pin-geometry and mirror-canary tests load, and the regression test locking in the no-seed guarantee. Verified locally: all fixture-dependent suites pass (39 tests across the legacy manager, add-component, and pin-locator files), the eeschema netlist oracle runs green against real KiCad, and production add_schematic_component works on a blank-created schematic end to end.

One real problem surfaced during that end-to-end check, and since maintainer edits are allowed I pushed the fix to your branch rather than bouncing it: the template's format token 20260306 makes KiCad 10.0.0 refuse to load the file outright ('Failed to load schematic'). KiCad rejects files claiming a format version newer than the running build, and 20260306 is a later-10.0.x token — this machine's 10.0.0, whose own kicad-cli sch upgrade writes 20260101, cannot open it. Not your error: the issue text itself quoted 20260306 (from the reporter's newer build), our own fallback writers shipped the same token in v2.3.0, and it silently explained a previously-mysterious unloadable test fixture from #284. The pushed commit sweeps every template, fallback writer, and fixture to 20260101, which every 10.0.x loads (newer builds upgrade on save). Re-verified after the sweep: blank template loads on 10.0.0, blank project + component add round-trips with ERC 0 errors 0 warnings, the #284 fixture now loads in kicad-cli, full suite at the known-environmental baseline.

Merging — this closes #221 and #243 and finishes the whole June scaffolding cluster. Thanks for carrying A+C exactly to spec, and for the clean integration with #288 and #290 (the sheets-UUID flow through write_kicad_pro survived untouched).

PR #299, opened 2026-07-04
Fix schematic pin binding: instance pin UUIDs, grid snap, multi-unit pin location (#241, #239)

### LINKED ISSUE #239: get_schematic_pin_locations returns wrong coordinates for multi-unit components

### LINKED ISSUE #241: add_schematic_component: missing (pin "N" (uuid ...)) entries blocks ERC

bluebirdlboro:
## Summary

Components added via `add_schematic_component` produced schematics where wires and net labels did not bind to pins: ERC reported every pin unconnected and connections did not survive into the netlist. This PR fixes the root causes plus a couple of related robustness/perf fixes.

Verified end to end on Linux with KiCad 10 `kicad-cli` (real symbol/footprint libraries): after the fix a drawn wire's net reaches the exported netlist and `sync_schematic_to_board` assigns the pads correctly.

## Fixes

**1. Missing `(pin "N" (uuid ...))` in placed instances (fixes #241)**
`DynamicSymbolLoader.create_component_instance` emitted no per-pin UUID entries, which KiCad requires to associate wires/labels with pins. Now injects one `(pin "N" (uuid ...))` per pin, unit-aware (parsed from the injected `lib_symbols` sub-symbols `<name>_<unit>_<style>`).

**2. Off-grid placement**
Components were written at raw millimetre coordinates, leaving pins off the 1.27 mm connection grid so wire endpoints never coincided with pins. The symbol origin is now snapped to the 1.27 mm grid.

**3. Multi-unit pins located from the wrong unit (fixes #239)**
For multi-unit symbols (op-amps, logic gates), each unit is a separate placement sharing the reference. `get_schematic_pin_locations` computed every pin from whichever instance appeared first in file order. `WireDragger.find_symbol` now accepts a `unit`, and `PinLocator` maps each pin to its unit and reads that unit's placement.

**4. Clean-template option**
`create_schematic` gains a `template` parameter so the dynamic workflow can start from a template without the hidden `_TEMPLATE_*` clone-source symbols that otherwise add phantom ERC/netlist noise. Default template is unchanged, so the legacy `place_component` clone path is unaffected.

**5. Symbol-library parsing cache (perf)**
A fresh loader is created per component add and a fresh interface per session, so instance-level caches never survived: with 200+ libraries installed, each interface spawned a full-library warm thread and each add re-parsed a multi-MB `.kicad_sym`. Added process-wide caches (keyed by resolved file path), warm-once-per-process, and a `KICAD_SKIP_SYMBOL_WARMUP` opt-out. ~10x faster per add; interface construction no longer grows super-linearly.

## Tests

- New `tests/test_pin_locator_multiunit.py` (self-contained two-unit schematic), plus pin-UUID / grid-snap assertions in `tests/test_add_schematic_component.py`.
- Full suite green (1062 passed, 12 skipped).

### DIFF
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -613,6 +613,28 @@ class DynamicSymbolLoader:
             return []
 
     @staticmethod
+    def _extract_schematic_uuid(schematic_path: Path) -> Optional[str]:
+        """
+        Return the root schematic's top-level UUID, used to build KiCad-native
+        ``(instances (project ... (path "/<root-uuid>" ...)))`` annotation blocks.
+
+        The top-level ``(uuid ...)`` sits in the header before ``(lib_symbols ...)``,
+        so the first UUID in the file is the schematic's own.  Handles both the
+        GUI-saved quoted form ``(uuid "…")`` and this project's template form
+        ``(uuid <hex>)`` (see schematic.py).  Returns None on any failure so callers
+        fall back to the previous ``path "/"`` behaviour rather than crashing.
+        """
+        import re
+
+        try:
+            with open(schematic_path, encoding="utf-8") as f:
+                content = f.read()
+            m = re.search(r'\(uuid\s+"?([0-9a-fA-F-]{36})"?', content)
+            return m.group(1) if m else None
+        except Exception:
+            return None
+
+    @staticmethod
     def _rotate_offset(dx: float, dy: float, angle_deg: float) -> tuple:
         """
         Rotate a 2-D offset by angle_deg degrees (KiCad CCW-in-screen, i.e. Y-axis
@@ -699,6 +721,17 @@ class DynamicSymbolLoader:
         fp_x, fp_y, _, _ = _prop_at("Footprint", 0, 0, 0)
         ds_x, ds_y, _, _ = _prop_at("Datasheet", 0, 0, 0)
 
+        # --- annotation (instances) block ---------------------------------------
+        # KiCad reads the RefDes shown in the GUI from this block, not from the
+        # (property "Reference" ...) above.  It must name the project (= schematic
+        # file stem) and address the root sheet by its own UUID, otherwise KiCad 9's
+        # eeschema shows the component as unannotated (R?).  Native format (see
+        # demos/interf_u): (project "<stem>" (path "/<root-uuid>" ...)).
+        project_name = schematic_path.stem
+        root_uuid = self._extract_schematic_uuid(schematic_path)
+        instance_path = f"/{root_uuid}" if root_uuid else "/"
+        instance_project = project_name if root_uuid else "project"
+
         mirror_str = " (mirror y)" if mirror_y else ""
         instance_block = f"""  (symbol (lib_id "{full_lib_id}") (at {x} {y} {angle}){mirror_str} (unit {unit})
     (in_bom yes) (on_board yes) (dnp no)
@@ -716,8 +749,8 @@ class DynamicSymbolLoader:
       (effects (font (size 1.27 1.27)) (hide yes))
     ){pin_lines}
     (instances
-      (project "project"
-        (path "/"
+      (project "{instance_project}"
+        (path "{instance_path}"
           (reference "{reference}")
           (unit {unit})
         )
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -16,6 +16,18 @@ from typing import Dict, List, Optional, Tuple
 
 logger = logging.getLogger("kicad_interface")
 
+# Module-level caches shared across DynamicSymbolLoader instances.
+# Symbol libraries are effectively immutable during a process run, but a fresh
+# loader is created for every add_component call (see schematic_handlers), so an
+# instance-level cache never survives — every component add would otherwise
+# re-scan the sym-lib-table and re-read the multi-MB .kicad_sym file.  These
+# caches make library resolution and symbol extraction pay their cost once.
+# Call DynamicSymbolLoader.clear_library_caches() if a library changes on disk
+# mid-session.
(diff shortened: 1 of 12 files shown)


mixelpixx:
Thanks for the thorough diagnosis — every problem you identified was real. The awkward news is that this branch was authored against a main that has since moved a lot (16 merges in the past three days), and three of your four fixes have been merged from other angles while this sat in the queue:

1. Per-pin (pin "N" (uuid ...)) entries: merged via #265, which rewrote create_component_instance into a complete KiCad 10 instance writer (unit-aware pin extraction included). Issue #241 is closed.
2. Multi-unit pin locations: merged via #272 (unit-aware transforms in PinLocator/WireDragger). Issue #239 is closed.
3. Clean templates: merged via #295 this morning — create_schematic/create_project now start from a blank KiCad 10 template by default; the _TEMPLATE_* seeds are gone entirely, so a template parameter to opt out of them is no longer needed.

Your fix 2 — snapping the placement origin to the 1.27 mm grid — is NOT covered by any of those, and I just verified on current main that a component placed at raw coordinates (e.g. x=100.33) still lands off-grid, exactly the wire-binding failure you describe. That delta is genuinely wanted.

Suggestion: rebase onto current main and narrow the PR to the grid-snap fix plus its tests (your conftest/test additions that still apply are welcome too). One design question to address in the narrowed version: whether to snap silently or also report the snapped coordinates in the tool result, since a caller who asked for x=100.33 should be able to see it got 100.33 -> 99.06 or 101.6. Sorry for the collision — the queue moved unusually fast this week, and your netlist-verified end-to-end approach is exactly the standard we want. verify_wiring_poc.py should be dropped from the final diff (scratch script).

mixelpixx:
Thanks for the new commits. Before this can move, though, it still needs the restructuring from my earlier comment -- the branch remains based on a main from before the recent merge wave (your test count says 1062; main's suite is at 1330 today), so three of the original four fixes are still in this diff even though they are already merged from other angles (#265 for per-pin UUIDs, #272 for multi-unit pin locations, #295 for clean templates -- blank templates are now the default, so the new `template` parameter here would land as dead weight). `verify_wiring_poc.py` is also still in the diff.

What I want from this PR has actually grown, because your newest addition is good: the process-wide symbol-library cache addresses a real cost on main (every `SymbolLibraryManager` instantiation spawns its own warm thread and re-parses multi-MB `.kicad_sym` files -- I just confirmed the instance-level `_cache_lock`/warm-thread structure in `library_symbol.py` is exactly as you describe). So there are now two genuinely-new deltas worth landing, and they are unrelated to each other:

1. **Grid snap** for component placement (verified still broken on main: raw coordinates land off the 1.27 mm grid and break wire binding). Please include the snapped-coordinate reporting question from my earlier comment in the design.
2. **Process-wide symbol cache + warm-once + `KICAD_SKIP_SYMBOL_WARMUP`** (perf).

Please split these into two focused PRs, each rebased onto current main. Small, single-topic PRs merge same-day here; bundles sit in the queue -- this one has been the demonstration. If splitting is a hassle, do the grid-snap one first; the cache work will not go stale, and I am happy to review it second.

bluebirdlboro:
Superseded by two focused PRs as requested:

- **#319** — fix: snap component placement origin to 1.27 mm grid
- **#320** — perf: add module-level caches for library discovery and symbol extraction

Both rebased onto current main, split by topic, and with clean single-topic diffs.

PR #301, opened 2026-07-05
fix: resolve symdir (extends) parent from sibling shard (#282)

### LINKED ISSUE #282: symdir libraries: (extends) parent in sibling shard is not resolved, derived symbols extract incomplete
## Summary

#193 (merged) added KiCad 10 `.kicad_symdir` directory-library support: `DynamicSymbolLoader.find_library_file` resolves `<Lib>.kicad_symdir`, and `extract_symbol_from_library` reads the per-symbol shard `<Lib>.kicad_symdir/<Symbol>.kicad_sym`. Base symbols work end-to-end (verified against a stock Windows KiCad 10.0 install: `Device:R` extracts correctly from the symdir).

One gap remains for **derived symbols**: in the sharded format, a symbol that uses `(extends "Parent")` lives in its own file and the parent lives in a **sibling shard**. `_inline_extends_symbol(lib_content, ...)` searches only the single shard's content, so the parent is never found.

## Reproduction (stock KiCad 10.0, Windows)

`Device.kicad_symdir/Filter_EMI_C.kicad_sym` contains only `Filter_EMI_C` with `(extends "C_Feedthrough")`; `C_Feedthrough` is in `C_Feedthrough.kicad_sym` next to it.

```
extract_symbol_from_library("Device", "Filter_EMI_C")
-> WARNING Cannot resolve parent 'C_Feedthrough' for 'Filter_EMI_C' - stripping extends clause (symbol may be incomplete)
```

The result is a symbol shell without the parent's pins/graphics. Degraded but not corrupting: the extends clause is stripped (KiCad refuses `(extends)` inside a schematic's `lib_symbols`, so stripping keeps the file loadable), and only derived symbols are affected — `R`, `C`, `D`, `LED`, and other base symbols are fine.

## Suggested fix

In `extract_symbol_from_library`, when the library resolved to a directory and the child block contains `(extends "Parent")`, read the sibling shard `lib_path / f"{parent}.kicad_sym"` and pass that content to `_inline_extends_symbol` (or resolve the parent recursively — parents can themselves extend). Worth a unit test with a two-shard fixture directory.

boovaragan:
## Summary

Fixes #282. In KiCad 10's sharded `.kicad_symdir` libraries each symbol is its
own `<Symbol>.kicad_sym` file, so a derived symbol that uses `(extends "Parent")`
has its parent in a **sibling shard**, not in its own file.
`extract_symbol_from_library` read only the child's shard and handed that to
`_inline_extends_symbol`, which therefore never found the parent, stripped the
`(extends)` clause, and returned a symbol shell with no parent pins/graphics
(the issue's example: `Device:Filter_EMI_C`, which extends `C_Feedthrough`).

## Fix (`python/commands/dynamic_symbol_loader.py`)

- New `_resolve_symdir_extends(lib_dir, symbol_name, block)`: reads the parent
  shard `lib_dir/<Parent>.kicad_sym`, resolves it **recursively** (a parent may
  itself extend a grandparent in yet another shard), then merges the
  fully-resolved parent into the child via the existing single-level
  `_inline_extends_symbol`. A `_seen` set guards against cyclic `(extends)`
  chains, and a missing/unreadable parent shard falls back to the inliner's
  existing graceful strip-and-warn (the file stays loadable — KiCad refuses
  `(extends)` inside a schematic's `lib_symbols`).
- Small `_read_symdir_shard` helper for reading a shard.
- `extract_symbol_from_library` now branches the extends handling: **directory
  library → `_resolve_symdir_extends`**; single-file `.kicad_sym` library → the
  existing `_inline_extends_symbol` path, unchanged.

## Tests

New `tests/test_symdir_extends.py` builds a fake `.kicad_symdir` of shards
(pointing the loader's bundled-library search at a tmp dir) and asserts:

- a derived symbol inlines its parent's pins from the sibling shard, renames the
  parent's unit sub-symbols to the child, keeps the child's property overrides,
  and emits **no** `(extends)`;
- a three-shard `Base → Mid → Top` chain resolves end-to-end (grandparent pins
  reach the grandchild);
- a genuinely missing parent shard degrades gracefully (strips + no crash).

The fixture uses a non-installed library name so the test can't accidentally
resolve against a real KiCad `Device` library on the dev machine. Single-file
libraries are unaffected; existing `test_symbol_pins` /
`test_symbol_instance_completeness` / `test_loader_global_sym_lib_table` pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **Derived symbols in KiCad 10 `.kicad_symdir` libraries now inline their parent**
+  (#282): in the sharded directory format each symbol is its own
+  `<Symbol>.kicad_sym` file, so a symbol using `(extends "Parent")` has its parent
+  in a _sibling_ shard. `extract_symbol_from_library` handed only the child's shard
+  to the inliner, so the parent was never found and the `(extends)` clause was
+  stripped — producing a symbol shell with no parent pins/graphics (e.g.
+  `Device:Filter_EMI_C`, which extends `C_Feedthrough`). A new
+  `_resolve_symdir_extends` reads the parent shard, resolves it recursively (a
+  parent may itself extend a grandparent in another shard), and merges it via the
+  existing single-level inliner; a missing parent shard still degrades gracefully
+  (strips + warns, keeping the file loadable). Single-file `.kicad_sym` libraries
+  are unaffected.
+
 - **`create_project` writes a conformant KiCad 10 `.kicad_pro`** (#220): the
   project file was a hand-rolled 122-byte stub containing only
   `board.filename` and a `sheets` entry with the literal id `"root"`, so
@@ -145,7 +158,7 @@ the KiCad GUI connects later (reopen the project to adopt IPC).
   components via dynamic template injection** (#221, part B): when no placed
   `_TEMPLATE_*` donor existed, the legacy clone path used to call
   `DynamicSymbolLoader.load_symbol_dynamically`, which wrote a template
-  instance into the *file* mid-call and cloned onto a locally reloaded
+  instance into the _file_ mid-call and cloned onto a locally reloaded
   object — so callers following the normal add-then-save pattern saved
   their stale in-memory schematic, discarding the new component and
   leaving `_TEMPLATE_*` clutter behind. That branch is removed: template
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -330,6 +330,53 @@ class DynamicSymbolLoader:
 
         return first_line + "\n" + "\n".join(body_lines) + "\n" + last_line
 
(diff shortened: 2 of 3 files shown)


mixelpixx (approved):
Verified beyond the synthetic test trees: I staged a genuine KiCad 10 74xx.kicad_symdir (from a real 10.0.0 install) into the discovery path and ran the extraction on 7400, which carries (extends "74LS00") with the parent in a sibling shard — exactly the #282 shape. Result: a complete 14 KB symbol with all parent pins and unit bodies inlined and the extends clause resolved, where pre-fix code returned a pinless shell. Your 8 synthetic tests plus the add-component and instance-completeness suites all pass, black/mypy clean, full suite at the known-environmental baseline.

The recursive parent resolution with graceful degrade on a missing shard is the right shape — same degrade-not-corrupt principle the instance writer follows. One note for the record, not about this PR: on-machine discovery of the real symdir location only worked because I staged it manually; stock discovery still misses custom install roots like C:\KiCad, which is #286 and is fixed by your #296 — reviewing that next, and together they make this fix reachable end-to-end on such installs. I pushed a trivial CHANGELOG conflict resolution (both sides added entries at the top; kept both). Merging — closes #282.

PR #304, opened 2026-07-05
fix: use pin_world_xy in _find_pins_on_net for rotated symbols

### LINKED ISSUE #303: _find_pins_on_net mislocates pins on rotated symbols (uses a local transform instead of pin_world_xy)
## Summary

`_find_pins_on_net` in `wire_connectivity.py` computes each pin's world coordinate with its own inline transform instead of calling `WireDragger.pin_world_xy`. The inline block applies mirror **before** rotation, which diverges from `pin_world_xy` (the shared transform corrected in #259) for 90 and 270 degree rotations. Pins on rotated symbols are placed at the wrong coordinate, so they don't land on their own wires and are silently dropped from the pin list `get_wire_connections` returns.

## Divergence

For an asymmetric library pin point, comparing `pin_world_xy` against the `_find_pins_on_net` inline transform across all 16 rotation x mirror combinations, they disagree in exactly four:

| rotation | mirror       | result |
|----------|--------------|--------|
| 90       | none         | differ |
| 90       | x and y      | differ |
| 270      | none         | differ |
| 270      | x and y      | differ |

All 0/180 cases agree; 90/270 with exactly one mirror axis agree. So the bug affects any symbol rotated 90 or 270 with zero or two mirror axes.

## Reproduction

Schematic with an LM324 (`Amplifier_Operational:LM324`) whose gates are placed rotated 90 degrees, no mirror (e.g. reference IC-6, units 2/3/4).

`kicad-cli sch export netlist` ground truth:
```
Net-(IC-6-Pad7): IC-6/7, R10/1, R5/2, R6/2
```

`WireDragger.pin_world_xy` puts IC-6 pin 7 at (294.64, 119.38) - correct; `get_wire_connections` at that point returns the right net. But `_find_pins_on_net`'s inline transform computes pin 7 at (294.64, 134.62) - 15.24 mm off in y - so IC-6/7 is absent from the returned pin list even though it is on the net:

```
get_wire_connections at (294.64, 119.38):
  expected: ['IC-6/7', 'R10/1', 'R5/2', 'R6/2']
  actual:   ['R10/1', 'R5/1', 'R6/2']          <- IC-6/7 missing
```

On this schematic, eight IC-6 pins across units 2, 3 and 4 are mislocated this way (pins 5,6,7,9,10,12,13,14).

## Fix

Replace the inline transform in `_find_pins_on_net`:

```python
px, py = pdata["x"], pdata["y"]
py = -py
if mirror_x:
    py = -py
if mirror_y:
    px = -px
if sym_rot != 0:
    px, py = locator.rotate_point(px, py, sym_rot)
abs_x = sym_x + px
abs_y = sym_y + py
```

with the shared, #259-corrected transform:

```python
abs_x, abs_y = WireDragger.pin_world_xy(
    pdata["x"], pdata["y"], sym_x, sym_y, sym_rot, mirror_x, mirror_y
)
```

This is the same consolidation #259 already did for `WireManager._collect_pin_positions`; `_find_pins_on_net` was a second divergent copy that #259 didn't touch.

Verified: with `pin_world_xy`, the eight IC-6 pins resolve onto their own nets and match the netlist; single-unit and unrotated parts are unchanged.

## Relationship to #293/#294

Independent but adjacent. #293/#294 fixed *which* pins each instance owns (unit filtering). This is about *where* each owned pin is computed to be. Both are in `_find_pins_on_net`.

## Environment

- kicad-mcp, current main (post #259, #272, #294)
- KiCad 10.0.3, Ubuntu 24.04, Node v20.20.2

zerthimon:
## Summary

Fixes #303.

Fixes the rotated-symbol pin mislocation in `_find_pins_on_net`.

`_find_pins_on_net` computed each pin's world coordinate with a local transform that applied mirror before rotation. That ordering diverges from `WireDragger.pin_world_xy` (the shared transform corrected in #259) for 90 and 270 degree rotations, so pins on rotated symbols were placed at the wrong coordinate, failed to land on their own wires, and were silently dropped from the pin list `get_wire_connections` returns.

## Root cause

`WireManager._collect_pin_positions` was already routed through `pin_world_xy` by #259, but `_find_pins_on_net` carried a second, divergent copy of the transform math. For an asymmetric pin point the two disagree in exactly four of sixteen orientation combos: 90 and 270 rotation, each with zero or two mirror axes.

## Evidence

LM324 (`Amplifier_Operational:LM324`), gates placed rotated 90, reference IC-6. Ground truth from `kicad-cli sch export netlist`:

```
Net-(IC-6-Pad7): IC-6/7, R10/1, R5/2, R6/2
```

Before: `_find_pins_on_net` places IC-6 pin 7 at (294.64, 134.62) instead of (294.64, 119.38), so `get_wire_connections` returns `['R10/1', 'R5/1', 'R6/2']` - IC-6/7 missing. Eight IC-6 pins (units 2/3/4) were affected. After: all resolve onto their own nets and match the netlist.

## Change

Replaces the inline transform block with a `WireDragger.pin_world_xy` call, so pin geometry matches eeschema (Y-flip -> rotate -> mirror -> translate) everywhere.

## Tests

New `TestRotatedSymbolPinLocation` in `tests/test_wire_connectivity_multi_unit.py`: places a symbol rotated 90 degrees and asserts its output pin lands on its own net. Fails without the fix (verified by stashing the change).

Also corrects two assertions in the existing #293 tests. That fixture places its sense resistors rotated 90 degrees; with the transform fixed, the wire-end resistor pin is pin 1, not pin 2. The pre-fix `R*/2` values reflected the same rotation bug acting on those resistors, so this is the expected corrected geometry, not a behavior regression.

```
pytest tests/test_wire_connectivity_multi_unit.py           # 3 passed
pytest tests/                                                # 1322 passed
    (4 pre-existing failures on main, unrelated: Windows path
     tests on Linux + a SymbolLibraryManager._cache_lock bug)
```

Black, isort, flake8, mypy clean on the changed Python files.

## Relationship to #293/#294

Independent but adjacent. #294 fixed which pins each instance owns (unit filtering); this fixes where each owned pin is computed to be. Both touch `_find_pins_on_net`.

## Checklist

- [x] Code follows style guidelines (black, isort, flake8, mypy)
- [x] All tests pass locally (new test fails without the fix)
- [x] CHANGELOG.md updated under [Unreleased]
- [x] Commit message follows convention

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **`get_wire_connections` locates pins correctly on rotated symbols**:
+  `_find_pins_on_net` computed pin world coordinates with a local transform
+  that applied mirror before rotation, diverging from `WireDragger.pin_world_xy`
+  (the shared transform corrected in #259) for 90 and 270 degree rotations. Pins
+  on rotated symbols were placed at the wrong coordinate and dropped from their
+  own net's pin list (observed on an LM324 whose gates are placed rotated 90:
+  eight of its pins went missing from their nets). `_find_pins_on_net` now calls
+  `pin_world_xy` so pin geometry matches eeschema everywhere. Single-unit and
+  unrotated parts are unaffected.
+
 - **`get_wire_connections` no longer reports phantom cross-unit pins** (#293):
   `_find_pins_on_net` transformed every unit's pins against every placed
   instance of a multi-unit component, so a sibling unit's pin whose library
--- a/python/commands/wire_connectivity.py
+++ b/python/commands/wire_connectivity.py
@@ -15,6 +15,7 @@ from typing import Any, Dict, List, Optional, Set, Tuple
 
 import sexpdata
 from commands.pin_locator import PinLocator
+from commands.wire_dragger import WireDragger
 from sexpdata import Symbol
 
 logger = logging.getLogger("kicad_interface")
@@ -472,17 +473,20 @@ def _find_pins_on_net(
             mirror_y = inst["mirror_y"]
 
             for pin_num, pdata in pin_defs.items():
-                px, py = pdata["x"], pdata["y"]
-                # y-negate: lib_symbols y-up → schematic y-down
-                py = -py
-                if mirror_x:
-                    py = -py
-                if mirror_y:
-                    px = -px
-                if sym_rot != 0:
-                    px, py = locator.rotate_point(px, py, sym_rot)
-                abs_x = sym_x + px
-                abs_y = sym_y + py
+                # Use the shared symbol->world transform so pin geometry matches
+                # eeschema (Y-flip -> rotate -> mirror -> translate). A local copy
+                # of this math applied mirror before rotation, which disagrees with
+                # pin_world_xy for 90/270 rotations and mislocated pins on rotated
+                # symbols, dropping them from their own net's pin list.
+                abs_x, abs_y = WireDragger.pin_world_xy(
+                    pdata["x"],
+                    pdata["y"],
+                    sym_x,
+                    sym_y,
+                    sym_rot,
+                    mirror_x,
+                    mirror_y,
+                )
                 if _on_net(abs_x, abs_y):
                     key = (ref, pin_num)
                     if key not in seen:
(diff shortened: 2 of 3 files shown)


mixelpixx:
Merged. Clean work -- the issue write-up, the 16-combination divergence table, the LM324 reproduction, and the corrected test assertions for the #294 resistors were all exactly right.

One observation for context: this is the second call-site that was still carrying the pre-#259 transform math. The first was `WireManager._collect_pin_positions`, which #259 fixed by routing through `pin_world_xy`. This one in `_find_pins_on_net` got missed because it was a separate function in a separate file. Both are now on the shared transform.

1288 tests pass (up from 1287), black/mypy clean, no regressions.

Issue #249, opened 2026-06-11
autoroute: restore the board to its pre-export state when a run is cancelled or crashes


Dewieinns:
## Environment
- KiCad 10.0.1, Windows 11
- MCP server: main post-PR-#212

## Severity
**High** — any autoroute failure or cancellation kills the entire MCP session.

## Symptom
If `autoroute` is cancelled mid-run or Freerouting crashes on a complex board, all subsequent MCP calls fail with SWIG / NoneType errors. `get_backend_state` itself fails. `open_project` returns "SWIG proxy is dehydrated and recovery failed". The only remedy is a full VS Code / Claude Code restart.

## Root cause
The `autoroute` flow calls `pcbnew.ExportSpecctraDSN`, then spawns Freerouting as a subprocess and waits. If the subprocess is killed or crashes while Python's SWIG `board` object is mid-mutation, the internal board state can be left inconsistent. There is no recovery path: no `reset_backend` tool exists, and `open_project` re-uses the same corrupted SWIG state rather than reinitialising it from a fresh `pcbnew.LoadBoard` call.

Additionally, the MCP framework's call timeout (shorter than Freerouting's runtime — see the autoroute-timeout issue) can kill the Python-side coroutine while the Freerouting subprocess is still alive, leaving the board object partially mutated.

## Expected behaviour
- A `reset_backend` or `reload_board` tool should exist to recover without restarting the host.
- `open_project` should always reinitialise from disk via a fresh `pcbnew.LoadBoard`, not attempt to re-use an existing (potentially dehydrated) proxy.
- Autoroute should be cancellable cleanly, restoring the board to its pre-export state on failure.

## Workaround
Full restart of VS Code / Claude Code after any autoroute failure or cancellation.

mixelpixx:
Status after the 2026-07-02 merge wave — two of the three expected behaviors now exist, keeping open for the third.

1. 'A reset_backend or reload_board tool should exist': #271 added close_project. close_project (save=false) followed by open_project drops the SWIG board, the IPC board API, and the session backend pin, then reloads from disk — a full backend reset without restarting the host. If you can retest an autoroute crash on current main, this cycle should now recover where it previously required a VS Code restart.
2. 'open_project should reinitialise from disk': open_project's load path goes through _safe_load_board with dehydration recovery, and the #255 session-pinning work made its state reset explicit (a failed reopen clears the stale pin rather than reusing old IPC context).
[...]

PR #305, opened 2026-07-06
[codex] Harden MCP request handling and schematic replacement

wendell21-z:
## Summary

- connect the MCP stdio transport before the Python worker finishes initialization
- keep timed-out requests synchronized until late Python responses are consumed
- reject active and queued requests when the Python worker exits
- route warm-up through the normal serialized request queue
- apply configured Python, KiCad, and log paths instead of silently ignoring them
- back up schematic files before component replacement and restore the original on failure
- add regression coverage for component replacement rollback

## Root cause

Timed-out Python commands cleared the active response handler and advanced the queue even though the worker continued running. A late response could therefore be delivered to an unrelated request. Component replacement also wrote the symbol deletion before loading the replacement, so loader failures could leave the schematic missing the original component. In addition, MCP transport connection was delayed until worker initialization completed, and several accepted configuration fields were not applied.

## User impact

Requests now remain correctly ordered across timeouts, worker exits fail pending calls explicitly, Codex can complete the MCP handshake during worker startup, and failed symbol replacements preserve both the original schematic and a durable backup under .mcp-backups.

## Validation

- npm run build
- npm run test:ts: 24 passed
- targeted Python tests: 14 passed
- pre-commit checks: trailing whitespace, EOF, line endings, large files, merge conflicts, Black, isort, Prettier, flake8, mypy, and ESLint passed
- live MCP stdio initialize handshake passed against KiCad 10.0.4
- broader Python run reached 631 passed and 1 skipped before interruption; one pre-existing Windows path case-deduplication test failed under the macOS-hosted Windows simulation


mixelpixx (changes requested):
Fit: yes on all four topics -- request-ordering correctness and pre-destructive-write recovery are core reliability work, and the validation rigor here (live handshake against KiCad 10.0.4, full lint battery) is appreciated, as is the codex disclosure. I verified the claims that needed verifying: KICAD_PYTHON is genuinely consumed at spawn (server.ts resolvePythonPath reads it before the spawn call), logger.setLogDir exists, and the config fields were indeed accepted-but-ignored before. Two of the four changes are clean keeps: rejecting active and queued requests on worker exit is plainly correct, and routing warm-up through the normal request queue removes bespoke handler-poking that was race-prone by construction.

One must-fix, one question, and a split request:

**Must-fix: the timeout change trades response-misattribution for a permanent deadlock on a hung worker.** Keeping the handler installed after rejecting the caller is the right instinct -- a late response resolving the next unrelated request is a real corruption vector. But the new code keeps the queue blocked until Python's late response arrives, and this repo has three documented cases where the worker hangs while staying alive: #247 (SWIG proxy dehydration), #248 (sync_schematic_to_board hangs >600 s), #249 (post-autoroute dehydration). Under any of those, today's behavior degrades (misattribution risk) but the server keeps working; with this change, one hung command bricks the entire request queue forever -- the worker-exit handler never fires because the process never exits. Please add a bounded reaper: after rejecting the caller, if the late response has not arrived within a recovery window (say 2x the original timeout), kill the Python worker outright -- your own new exit handler then rejects the queue cleanly and the supervisor can restart. Killing is honest here: after a hang the worker's state is unknowable anyway.

**Question on transport-before-READY:** connecting the transport first is the right direction (a 120 s READY wait before connect would starve any client handshake timeout). But it changes the failure mode when Python never becomes ready: the client now sees a live, handshaken server whose every call will queue or fail. Please confirm the waitForReady rejection still tears the server down with a nonzero exit rather than leaving that zombie -- a log line is not enough for MCP clients that only watch the process.

**Split request:** the schematic_batch.py rollback is independently mergeable and nearly ready -- please move it to its own PR so it does not wait on the transport work. Two notes for that PR: the project already has a checkpoint convention (snapshot_project writes to <project>/snapshots/), so either nest the per-file backups there or add a comment explaining why .mcp-backups is separate, so we do not accrete parallel backup mechanisms; and the 10-file cap plus the backup path in the success result are both good calls.

Sequencing note: #275 (stefangordon) touches the same queue with a bounded-wait proposal for the busy-reject. Whichever lands second needs to re-verify against the first; flagging so neither of you is surprised.

mixelpixx:
Ping — this has been a draft since 2026-07-11 with the must-fix from the 07-06 review still outstanding. Recapping so it is actionable, and flagging two changes on main that affect it.

**The must-fix, unchanged:** keeping the handler installed after a timeout is the right instinct — a late response resolving an unrelated next request is a real corruption vector. But as written the queue stays blocked until Python's late response arrives, and this repo has three documented cases where the worker hangs *while staying alive* (#247 SWIG proxy dehydration, #248 sync hanging >600 s, #249 post-autoroute dehydration). Under any of those, one hung command bricks the request queue permanently — your worker-exit handler never fires because the process never exits. Needed: a bounded reaper that kills the worker after a recovery window, so the exit handler can clean up.

Also still open from that review: please confirm that a `waitForReady` rejection tears the server down with a nonzero exit rather than leaving a handshaken-but-dead server, since you now connect the transport first.

**Two things moved on main:**

1. **#335 extracted the timeout policy out of `src/server.ts`** into a new `src/command-timeout.ts` as a pure, unit-tested function, while fixing #251 (`autoroute` was being abandoned at 30 s because it was missing from the long-running list). Your branch is CONFLICTING and that is now part of why. The good news is that a bounded-reaper constant has an obvious home there.
2. **#334 made CI real.** The Python job was a no-op — `pytest python/` collected 7 tests instead of ~1550, every gate ended in `|| echo`, and Actions was disabled repo-wide. Your PR's own validation was more thorough than CI was, which is a little embarrassing on our side; it is fixed now.

**Standing offer:** the `schematic_batch.py` rollback half is independently mergeable and nearly ready — split it into its own PR and I will merge it without waiting on the queue work. Two notes for that one: the project already has a checkpoint convention (`snapshot_project` writes to `<project>/snapshots/`), so either nest the per-file backups there or comment on why `.mcp-backups` is separate; and the 10-file cap plus surfacing the backup path in the result were both good calls.

If you have moved on, that is fine — say so and I will close this, keeping the rollback idea on the backlog. The diagnosis was sound; it is only the recovery behaviour that needs bounding.

mixelpixx:
Heads-up: this branch already needed a rebase, and **#339 has now added a second, unrelated source of conflict** — a repo-wide formatting normalization (black, isort and prettier over `tests/` and `src/tools/`, which the old `black --check python/` gate never covered).

Flagging it so the two do not get confused with each other when you rebase:

- **Formatting conflicts** are noise. Take either side of the hunk and run `npm run format:py` (plus `npx prettier --write .` if it is in `src/`). I verified the normalization is semantically inert before merging — every changed Python file AST-compared against the prior commit, giving 0 real changes, and an identical test result (1551 passed).
- **Substantive conflicts** are the pre-existing ones from main moving under this branch, and those are the ones worth reading carefully.

If a formatting-looking conflict turns out to be more than whitespace, tell me — that would mean I got something wrong in #339.

Also relevant to your next push: **CI now actually enforces formatting** via `pre-commit run --all-files` in the `code-quality` job. Running `pre-commit install` once (or `npm run format:py` before pushing) keeps you green. Worth knowing that until today CI's Python side was a complete no-op and Actions was disabled repo-wide — so if this PR ever felt like it was held to a standard nothing was checking, that is because it was.

Your outstanding review items are unchanged; nothing in #339 touches the substance here.

mixelpixx:
Closing with thanks — superseded. The request-handling hardening this draft aimed at is now owned by the #373 correlation work (#382: request IDs end-to-end, stale-response discard, per-PID logs), and the schematic-replacement half was overtaken by the #265 instance-writer rewrite and the #348/#354 escaping fixes. The branch has been conflicting since July with the must-fix items unaddressed; if you return and find something in your approach the merged work still lacks, a fresh focused PR is welcome.

PR #308, opened 2026-07-06
feat: add_symbol_property + add_library_symbol_property MCP tools

karu2003:
## Summary

Two new MCP tools for adding/updating custom properties on symbols:

### `add_library_symbol_property` — for `.kicad_sch` schematics
Add or update a custom property (Manufacturer, MPN, LCSC, etc.) on a symbol definition in the `lib_symbols` section of a schematic. Makes the property available to all instances of that symbol.

- Uses text surgery + `prettify()` for canonical KiCad formatting
- Preserves placed instances (pin UUIDs, references)
- 12 tests

### `add_symbol_property` — for `.kicad_sym` library files
Add or update a custom property directly on a symbol in a .kicad_sym library file. Useful for batch-adding fields (Manufacturer, Supplier, JLCPCB Part #, etc.) to library symbols.

### Files changed
- `python/commands/add_schematic_symbol_property.py` — schematic tool
- `python/commands/add_library_symbol_property.py` — library tool
- `python/kicad_interface.py` — command routing + imports
- `python/schemas/tool_schemas.py` — MCP schemas
- `src/tools/schematic-batch.ts` — TS registration (schematic)
- `src/tools/symbol-creator.ts` — TS registration (library)
- `tests/test_add_schematic_symbol_property.py` — 12 tests

### DIFF
--- a/python/commands/add_library_symbol_property.py
+++ b/python/commands/add_library_symbol_property.py
@@ -1,0 +1,106 @@
+"""Add or update a property on a symbol definition in the lib_symbols section."""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+from typing import Any
+
+from utils.sexpr_format import prettify
+
+
+def _find_symbol_in_lib_symbols(
+    content: str, library_name: str, symbol_name: str
+) -> tuple[int, int, str] | None:
+    """Return (start, end, block_text) for a symbol in lib_symbols, or None."""
+    full = f"{library_name}:{symbol_name}"
+    lib_start = content.find("(lib_symbols")
+    if lib_start == -1:
+        return None
+
+    marker = f'(symbol "{full}"'
+    sym_start = content.find(marker, lib_start)
+    if sym_start == -1:
+        return None
+
+    depth = 0
+    for i in range(sym_start, len(content)):
+        if content[i] == "(":
+            depth += 1
+        elif content[i] == ")":
+            depth -= 1
+            if depth == 0:
+                return sym_start, i, content[sym_start : i + 1]
+
+    return None
+
+
+def _has_property(symbol_block: str, prop_name: str) -> bool:
+    """Check if a property already exists in a symbol block."""
+    return bool(re.search(rf'\(property\s+"{re.escape(prop_name)}"', symbol_block))
+
+
+def _property_s_expr(
+    name: str, value: str, pos: dict[str, float] | None = None, hide: bool = False
+) -> str:
+    """Build a (property ...) s-expression string."""
+    x = pos.get("x", 0) if pos else 0
+    y = pos.get("y", 0) if pos else 0
+    parts = [f'(property "{name}" "{value}" (at {x} {y} 0)']
+    if hide:
+        parts.append("(hide yes)")
+    parts.append("(effects (font (size 1.27 1.27)))")
+    parts.append(")")
+    return "\n\t\t\t".join(parts)
+
+
+def add_library_symbol_property(params: dict[str, Any]) -> dict[str, Any]:
+    schematic_path = Path(params["schematicPath"])
+    library_name = params["libraryName"]
+    symbol_name = params["symbolName"]
+    prop_name = params["propertyName"]
+    prop_value = params["propertyValue"]
+    pos = params.get("position")
+    hide = bool(params.get("hide", False))
+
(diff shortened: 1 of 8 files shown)


mixelpixx (changes requested):
Fit: yes, happily. BOM-field property management (Manufacturer, MPN, LCSC) pairs directly with the JLCPCB integration, and batch-stamping library symbols is a real workflow. Three items before merge, one of them important:

1. **The tool names and function names are crosswired, and it reads as a bug.** In kicad_interface.py:

   ```python
   "add_library_symbol_property": add_schematic_symbol_property,
   "add_symbol_property": add_library_symbol_property,
   ```

   I checked the PR description and confirmed this is intentional -- your tool named `add_library_symbol_property` edits a schematic's lib_symbols section (backed by the function named `add_schematic_symbol_property`), and your tool named `add_symbol_property` edits `.kicad_sym` files (backed by the function named `add_library_symbol_property`). But a future reader will see those two lines as a copy-paste mistake and "fix" them, breaking both tools at once. Please align the names so tool name == function name == module name. My suggestion: keep the tool names as-is and rename the modules/functions to match them (`add_library_symbol_property.py` gets the schematic lib_symbols logic, the `.kicad_sym` logic moves to `add_symbol_property.py`), since renaming the public tool names is the more disruptive direction. While you are in there, give the two schema entries distinct titles -- both currently say "Add Property to Library Symbol".

2. **The `.kicad_sym` tool has no tests.** All 12 tests cover the schematic path, but the `.kicad_sym` module does its own separate text surgery with its own copy of the property-replacement regex (which is intricate enough that it needs a regression guard -- nested `(effects (font ...))`, `(hide yes)` ordering). Minimum set: add new property, replace existing property, symbol-not-found, and one guard that a multi-unit symbol's nested sub-symbols (`(symbol "R_0_1"` inside `(symbol "R"`) are not what gets matched when the caller passes the parent name.

3. Two smaller things, your call: `_has_property` and the property-builder are duplicated verbatim across both modules -- a small shared helper would keep the replacement regex in one place. And the `.kicad_sym` write skips the canonical formatter while the schematic write uses it; if that is deliberate (prettify was ported for schematics), a one-line comment saying so would prevent someone "fixing" it later.

One design note, not blocking: a property added via the schematic-side tool lives only in that schematic's lib_symbols cache, so a later `update_symbol_from_library` refresh (#291, just merged) will overwrite it. For durable BOM fields the right flow is your `.kicad_sym` tool first, then refresh. Worth a sentence in the tool descriptions so agent callers pick the right one.

Re-review will be quick -- the core text surgery looks right and the schematic-side test coverage is exactly what I would have asked for.

mixelpixx:
Merged. The restructure addressed everything: modules and functions now match the tool names, the .kicad_sym tool has its own test file including the nested sub-symbol guard, the schema titles are distinct, and the cross-referencing descriptions (cache-only vs. permanent, refresh ordering) are exactly the guidance agent callers need.

One thing I fixed during the merge, and it is worth flagging because it was invisible to your test run: the command_routes entry in kicad_interface.py still referenced the old function name --

    "add_library_symbol_property": add_schematic_symbol_property,

-- after the rename removed that name from the imports. Your unit tests import the tool modules directly, so they pass, but constructing KiCADInterface raises NameError, meaning the server would have failed at startup. I renamed the reference and added a construction check to my merge verification (the interface now builds with 184 routes, all three new tools present). If you want a durable guard for future tool PRs, a test that instantiates KiCADInterface with a mocked pcbnew and asserts command_routes builds would catch this whole class of mistake -- happy to take that as a follow-up if you don't.

Also handled in the merge: conflicts with #291 (merged an hour before you pushed) in schematic-batch.ts and tool_schemas.py -- both registrations kept -- plus black on four files and the CHANGELOG entry.

Full suite: 1350 passed. Both property tools plus update_symbol_from_library make a nice coherent BOM-field workflow -- good pair of contributions today.

mixelpixx:
Follow-up landed on main (693da5f): tests/test_interface_construction.py now constructs KiCADInterface against the stubbed pcbnew and asserts every command_routes entry is callable, every schema-listed tool has a route, and recent tools are present. Verified it fails with NameError on the exact bug this PR carried, so the renamed-handler class of mistake is now caught by CI instead of at server startup. No action needed on your side -- future tool PRs just get the guard for free.

PR #310, opened 2026-07-07
fix: cache LibraryManager across sync_schematic_to_board calls

boovaragan:
## Summary

Part of #248 (fix idea 1 of 2 — see scope note below; the issue stays open for idea 2). `sync_schematic_to_board` → `_add_missing_footprints_from_schematic`
built a brand-new `LibraryManager` on **every** call — re-parsing the global
`fp-lib-table` plus the project's `fp-lib-table` (recursively following any
`Table` references) each time. In the iterative rebuild flow the tool exists
for (call `sync_schematic_to_board`, tweak the schematic, call it again), that
parse cost was paid again on every single call even though the project hadn't
changed, on top of the per-footprint `pcbnew.FootprintLoad()` cost the issue
also flags.

## Fix (`python/commands/schematic_handlers.py`)

- New `_get_project_library_manager(project_dir)` caches the `LibraryManager`
  on the `KiCADInterface` instance, keyed on the resolved project directory,
  and only rebuilds it when that directory changes — the same
  cache-on-project-change shape `place_component` already uses a few hundred
  lines up in the same file (`kicad_interface.py`'s `_current_project_path`
  check), so the pattern isn't new to this codebase.
- `_add_missing_footprints_from_schematic` now calls that helper instead of
  constructing `LibraryManager(project_path=project_dir)` inline.

## Scope note

This targets fix idea #1 from the issue ("cache `LibraryManager` across
calls"). It removes the redundant fp-lib-table re-parse on repeat calls
against the same project. It does **not** address fix idea #2 (batching /
reducing the per-footprint `FootprintLoad` cost) — if a *single* call on a
~40-component board still approaches the 600 s timeout after this change,
that remaining cost is the more likely culprit and is a separate, larger
change (touching the KiCad-side footprint loading, not the Python-side
caching this PR covers).

## Tests

`tests/test_sync_library_manager_cache.py` (3 new tests) unit-test the cache
directly: same project dir across repeated calls reuses the identical
`LibraryManager` instance (construction count stays at 1), a different
project dir triggers a rebuild (and switching back rebuilds again — no stale
cross-project reuse), and two separate `KiCADInterface`-like instances don't
share cache state (it's on `self`, not module-global).

Ran the existing `tests/test_sync_schematic_to_board_footprints.py` suite
unchanged — those tests already patch `commands.library.LibraryManager` per
call and construct a fresh `KiCADInterface()` per test, so they weren't
exercising the cross-call caching gap this PR closes, but they confirm the
add/skip logic itself is untouched.

Full suite: no regressions vs. the established baseline (5 autoroute + 3
Freerouting Java-env + 1 #287 order-dependent flake — all pre-existing and
environment-specific, unrelated to this change).

## Verification note

I don't have a large real-world board to reproduce the reported 600 s
timeout end-to-end, so I can't confirm this alone resolves it — only that it
removes the specific redundant-reparse cost the issue's fix idea #1
describes. If you can retest the original repro (~40-component board) on
this branch, that would confirm how much of the 600 s it recovers.


mixelpixx:
Verified and merging, with one addition made on the branch before merge (dc7016b) — the cache needed a freshness key, not just a directory key.

What was verified:

- The caching itself is correct and the tests are well-shaped: per-instance state (no cross-session sharing), rebuild on project switch, construction-count assertions against a recording fake. The function-local `from commands.library import LibraryManager` is what makes the `patch("commands.library.LibraryManager")` in the tests actually take effect — worth keeping if anyone refactors the import to module level.
- The existing `tests/test_sync_schematic_to_board_footprints.py` suite passes unchanged, and the full suite on this branch merged with main shows only the 8 known environmental failures. black and mypy clean.

The addition: a cache keyed only on the project directory outlives edits to the fp-lib-table files the manager parsed. That is not hypothetical in this codebase — `register_footprint_library` is a live MCP tool that rewrites the project or global fp-lib-table mid-session, and the KiCad GUI can do the same (Preferences > Manage Footprint Libraries) while the server runs. The failure mode is the bad kind: register a new library, call `sync_schematic_to_board`, and footprints from the just-registered library get silently skipped because the cached manager predates the registration. The freshness key now includes the `mtime_ns` of the project fp-lib-table and of the re-probed global table (two `stat` calls per sync — noise next to the parse they let us skip), so table edits trigger a rebuild automatically while your intended reuse path (repeat syncs against an unchanged project) still hits the cache. A regression test that edits the table between calls is included, and your three original tests pass unmodified.

The scope note in the PR description is appreciated — fix idea #2 from #248 (the per-footprint `FootprintLoad` cost) remains open there, and this merge does not close it. #248 stays open with a note.

PR #311, opened 2026-07-07
feat: add PCB editing interface tools

OHNope:
# feat: add PCB editing interface tools

I’ve been experimenting with using kicad-mcp-server in a more closed-loop PCB workflow, especially trying to let the model iterate on placement and routing decisions. While watching the model work, I noticed it repeatedly had to inspect raw board files or redo the same lookup/edit steps because some small PCB editing/query APIs were missing.

This PR adds a set of helper tools that expose those repeated operations directly, so agents can inspect and update board state more cleanly without parsing raw KiCad files as often.

Added tools include:

- Board outline / graphics helpers:
  - `clear_board_outline`
  - `replace_board_outline`
  - `list_graphics`
  - `delete_graphic`
  - `update_graphic`

- PCB component / placement helpers:
  - `batch_move_components`
  - `get_pads`
  - `get_net_pads`
  - `get_component_geometry`
  - `get_ratsnest`
  - `estimate_airwire_lengths`
  - `check_placement_clearance`
  - `move_footprint_text`

- Board/session helpers:
  - `open_board`
  - `reload_board`
  - `save_board`
  - `save_as`
  - `is_dirty`
  - `discard_or_reload`

- Schematic-to-board helper:
  - `create_board_from_schematic`

The implementation was AI-assisted with GPT-5.5 XHIGH, then reviewed locally and checked with:

```bash
npm run build
pytest tests/test_board_editing_closure_apis.py tests/test_interface_construction.py -q
```


mixelpixx (changes requested):
Fit: yes. Board-lifecycle control (open/reload/save/dirty), graphics editing and the geometry queries are all squarely in this server's lane, and I appreciate that all 20 tools are both schema'd **and** routed — the schema↔route parity guard (`tests/test_interface_construction.py`) passes on your branch, which is more than some recent PRs managed.

It also merges clean against main and against every other open PR I tested. The blocker is not conflicts; it is backend session pinning. Three concrete gaps, all verified against main:

**1. `_handle_create_board_from_schematic` never re-pins the session.** This is the important one.

It does `self.board = board`, `_update_command_handlers()` and `_record_board_signature()` — but not `_pin_session_backend()`. Main only pins inside the `if command == "create_project" or command == "open_project"` branch (`kicad_interface.py:1021`, pin at `:1072`), so nothing else re-pins for you. Your own `_handle_open_board` gets this right, which is what makes the omission look accidental rather than deliberate.

Consequence: after this call `session_board_path` still points at the **previous** board while `self.board` is a brand-new file. In a session pinned `"ipc"`, every subsequent IPC-capable command (`place_component`, `route_trace`, `save_project`) routes to the GUI's old board while SWIG holds the new one. That is exactly the divergence class #223/#255 exist to prevent, and it fails silently — the user sees successful writes landing in a file they are no longer looking at.

Please call `_pin_session_backend(self._current_board_path())` after the board swap, or explicitly clear the pin if you would rather force a re-pin on next open.

**2. `save_board` / `save_as` bypass IPC dispatch.**

Both call `self._handle_save_project(...)` directly rather than going through `handle_command`. `save_project` **is** in `IPC_CAPABLE_COMMANDS` (mapped to `_ipc_save_project`), and the routing decision lives at `kicad_interface.py:918` — `if command in self.IPC_CAPABLE_COMMANDS and self._session_allows_ipc()`. Calling the handler directly skips that check, so in an IPC-pinned session your tools write the SWIG in-memory board instead of asking the GUI to save.

`_auto_save_board`'s sha256 divergence guard catches the case where the GUI already *saved* to disk, but not unsaved GUI edits — which is the common case when someone has the board open. Either route through `handle_command`, or gate on `_session_allows_ipc()` and delegate.

**3. `batch_move_components` has no IPC handler while `move_component` does.**

`move_component` maps to `_ipc_move_component`; the batch variant does not. So in an IPC-pinned session single moves go to the GUI and batch moves go to SWIG — same board, two different owners. Add `batch_move_components` to `IPC_CAPABLE_COMMANDS`, or have it refuse when `_session_allows_ipc()` is true and point the caller at the single-move tool.

Please add a case to `tests/test_backend_session_pinning.py` covering at least (1) — that file covers the existing paths but nothing this PR adds, and a pinning regression is invisible without a test.

**Two scope questions, your call, not blocking:**

- `reload_board` and `discard_or_reload` are functionally the same call, and `save_as` is `save_board` with a path. Three of your 20 tools are aliases. Given the tool surface is already ~190 registrations and every entry costs an agent context to choose between, I would rather ship 17 well-named tools than 20 with three synonyms. If the aliases exist because a specific client kept reaching for the wrong name, say so and I will keep them.
- `check_placement_clearance` would be the **third** AABB-gap checker in the tree, alongside the existing `check_courtyard_overlaps` (`component.py:1394`) and `check_clearance` from the open PR #315. Not your fault — #315 landed in the queue at the same time — but worth deciding which one is canonical before all three exist.

Everything else looked good on inspection: the five mutating commands are correctly added to `_BOARD_MUTATING_COMMANDS` so they inherit the auto-save and external-change guard, which is the detail most contributors miss.

One heads-up: CI now actually runs the Python suite as of #334 (it was a no-op — `pytest python/` collected 7 tests instead of the ~1477 in `tests/`, and every gate ended in `|| echo`). Push again and you will get real results for the first time.

mixelpixx:

OHNope:
Thanks a lot for the heads-up and for confirming that the PR still merges cleanly.

In the meantime, I ran some additional reviews using GPT-5.6 Sol and Opus 5.0, which surfaced several different potential logic issues. I reviewed them manually and confirmed that some of the issues are real.

I’ll make sure to run the formatter before my next push. I’m currently working on the fixes and adding more tests to prevent similar bugs from slipping through. I’ll share the full details once I’ve completely isolated and resolved the problems.

mixelpixx:
Merged. All three session-pinning gaps are fixed, and two of them more thoroughly than I asked — I verified each against the branch rather than taking the commit messages for it.

**Gap 1** — `create_board_from_schematic` now calls `_pin_session_backend` after the board swap. Straightforward and correct.

**Gap 2** — `save_board`/`save_as` route through `handle_command("save_project", ...)`, and your comment states the principle better than my review did: *"the dispatcher is the single owner of the session-pinning decision."* You then went past the ask and handled something I had not raised — Save-As changes board **identity**, so on an IPC save you reload the SWIG board from the new path and realign the pin, and when that reload fails you explicitly drop the stale board rather than retaining it under the new session path. That second branch is the one that would have bitten someone eventually, and it is the harder half to think of.

**Gap 3** — refusing `batch_move_components` in an IPC-owned session with a pointer to `move_component` is the right call given the IPC BoardAPI only exposes single moves, and the error message says why.

The test coverage is well past what I asked for: named cases for each gap, plus Save-As overwrite semantics and "current path is not a Save As" verified on **both** backends. 32 tests in `test_backend_session_pinning.py` now.

Two things I handled in the merge, neither your fault:

1. **The README tool count.** #338 landed a test an hour before this that pins the README's headline count to `getRegistryStats()`, and it caught this immediately — `expected 124 to be 143`. Your 19 registered tools moved the number and the prose had no way to know. Updated all four references to 143 total / 111 routed / 13 categories. **This was the guard's first real catch**, which is a nice validation of the idea.
2. `isort` normalized a stray blank line in two of your new test files — a side effect of #339 making formatting an enforced gate a couple of hours ago.

On the two scope questions from my review, I am leaving them as you have them: the `reload_board`/`discard_or_reload` and `save_board`/`save_as` pairs stay. Having now read the Save-As identity handling, `save_as` clearly earns its slot — it does materially more than `save_board` with a path. I would still like to consolidate the three AABB-gap checkers (`check_placement_clearance`, `check_courtyard_overlaps`, #315's `check_clearance`) at some point, but that is a separate cleanup and not yours to carry.

Gates: 1586 Python passed / 4 skipped, 64 TypeScript passed, pre-commit clean, schema/route parity and session-pinning guards green.

First contribution to this repo and it is a substantial one — thanks for turning it around this quickly.

PR #312, opened 2026-07-07
fix: restore sys.modules state leaked by test_rotate_schematic_mirror.py

### LINKED ISSUE #287: Order-dependent test pollution: move_with_wire_preservation + pin_world_xy_eeschema_truth fail when run in one pytest invocation

boovaragan:
## Summary

Fixes #287. Reproduced exactly as reported:

```
pytest tests/test_rotate_schematic_mirror.py tests/test_move_with_wire_preservation.py
```
→ `6 failed, 44 passed` (matches the issue's own count for this file pair).

## Root cause

`tests/test_rotate_schematic_mirror.py` installs a private, unshared copy of
`wire_dragger.py` under the module name `"wire_dragger"` at **collection
time** (before any test runs), and — to keep that private copy's lazy
`from commands.pin_locator import PinLocator` (inside `get_pin_defs`) from
touching the real module — does:

```python
_pin_locator_mock = MagicMock()
sys.modules.setdefault("commands.pin_locator", _pin_locator_mock)
```

`setdefault` never gets undone. Once installed, it stays for the rest of the
pytest process. Any *later-collected* file that imports the real
`commands.pin_locator` — e.g. `test_move_with_wire_preservation.py`, via
`commands.wire_dragger.WireDragger.get_pin_defs` — gets this throwaway mock
instead. `PinLocator.parse_symbol_definition(...)` becomes a `MagicMock()`
call, and iterating the result (`for pin_num, pin in pins.items():`) is a
silent no-op — `MagicMock.__iter__` defaults to `iter([])` — so
`compute_pin_positions` returns `{}` instead of raising, which is exactly the
failure signature reported (`assert "1" in positions` against an empty dict).

A second, related leak exists in the same file's `test_rotate_handler_no_crash`:
it stubs `sys.modules["pcbnew"]`/`["skip"]`/etc. with *fresh* throwaway mocks,
`exec_module`s a private copy of `kicad_interface.py` (which eagerly imports
~20 `commands.*` submodules — several do `from skip import Schematic` /
`import pcbnew` at their own module level), then only restores the
*top-level* `sys.modules` entries in its `finally`. Any submodule imported
for the first time during that window keeps its own already-bound reference
to the discarded mock — restoring `sys.modules["skip"]` afterward doesn't
retroactively fix a submodule that already did `from skip import Schematic`
against it. This is the same class of bug and is plausibly what causes the
14 `test_pin_world_xy_eeschema_truth.py` oracle failures the original report
also lists (`SchematicManager`/`ComponentManager` both do
`from skip import Schematic` at module level) — see Verification note below.

## Fix (`tests/test_rotate_schematic_mirror.py` only — no production code touched)

- Added `teardown_module`, which undoes the `commands.pin_locator`
  `setdefault` once this file's own tests are done (only if it's still our
  mock instance, and only if nothing was there before).
- In `test_rotate_handler_no_crash`, track `sys.modules` before the stubbed
  `exec_module` call; in `finally`, evict any **`commands.*`** submodule that
  got newly added during that window, so a later, unstubbed import
  re-executes it cleanly. Scoped specifically to `commands.*` — an earlier
  version of this fix evicted *everything* new and broke re-importing PIL's
  C extension (`cannot import name '_imaging' from 'PIL'`), so third-party/
  stdlib modules pulled in as a side effect are left alone.

## Tests

`tests/test_pin_locator_module_leak.py` (new): runs pytest as a subprocess
against `test_rotate_schematic_mirror.py` + the fast, non-`KiCADInterface`
subset of `test_move_with_wire_preservation.py`
(`TestComputePinPositions`), asserting a clean pass. This is a cross-file
collection-order bug, so it can't be reproduced by importing modules
directly in-process — confirmed this test fails (reproducing both
`TestComputePinPositions` failures) when the fix is reverted, and passes
with it restored.

Manually reproduced and fixed the full reported combination:
`tests/test_rotate_schematic_mirror.py tests/test_move_with_wire_preservation.py`
→ was `6 failed, 44 passed`, now `50 passed`.

## Verification note

The issue's other reported symptom — 14 failures in
`test_pin_world_xy_eeschema_truth.py` — could not be re-verified on this
machine: that file's own gating (`shutil.which("kicad-cli")` plus a
hardcoded `/usr/share/kicad/symbols/Device.kicad_sym` check) always skips
here (no such path exists on Windows), so its test bodies never execute
regardless of this fix. The second leak this PR also closes (the
`test_rotate_handler_no_crash` submodule eviction) targets the same
`from skip import Schematic`-at-module-level pattern that
`commands.schematic`/`commands.component_schematic` share — the modules
those 14 oracle tests depend on via `SchematicManager`/`ComponentManager` —
so the fix should cover that failure mode too, but I can't directly confirm
it on this machine. If you can retest the original 4-file reproduction where
`kicad-cli` + the `Device` library are available, that would close the loop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,20 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **Fixed a test-suite state leak that caused spurious pin-position failures
+  when test files ran in combination** (#287): `tests/test_rotate_schematic_mirror.py`
+  installed a throwaway `MagicMock` at `sys.modules["commands.pin_locator"]`
+  via `sys.modules.setdefault(...)` at module-collection time, with no
+  teardown. Any later-collected file relying on the real
+  `commands.pin_locator` (e.g. `WireDragger.get_pin_defs`, via
+  `commands.wire_dragger`) silently got empty pin data instead of an error —
+  iterating a bare `MagicMock()` is a no-op by default. A `teardown_module`
+  now undoes the stub, and `test_rotate_handler_no_crash`'s stubbed
+  `kicad_interface.py` exec additionally evicts any `commands.*` submodule it
+  imported for the first time while `pcbnew`/`skip` were mocked, so later
+  tests get a clean re-import instead of a module bound to a discarded mock.
+  Test-only change; no production code touched.
+
 - **`import_ses` no longer creates phantom slashless nets — routed tracks bind
   to the real board nets** (#246): KiCad global-label nets are named with a
   leading `/` (e.g. `/GND`), but a Specctra DSN round-trip through Freerouting
@@ -155,7 +169,7 @@ load on every KiCad 10.0.x build.
   dynamic loader (and the legacy fallback was removed in #288), so the seeds only
   leaked into user files. Both tools now copy a new blank KiCad 10 template
   (`python/templates/blank.kicad_sch`: `(version 20260101) (generator
-  "eeschema")`, empty `lib_symbols`, no placed symbols).
+"eeschema")`, empty `lib_symbols`, no placed symbols).
   `template_with_symbols.kicad_sch` is kept unchanged in-repo as a test fixture.
   A regression test asserts a created schematic contains no `_TEMPLATE_`
   references and no seeded `lib_symbols` entries.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,7 +187,7 @@ load on every KiCad 10.0.x build.
   dynamic loader (and the legacy fallback was removed in #288), so the seeds only
   leaked into user files. Both tools now copy a new blank KiCad 10 template
   (`python/templates/blank.kicad_sch`: `(version 20260101) (generator
-"eeschema")`, empty `lib_symbols`, no placed symbols).
+  "eeschema")`, empty `lib_symbols`, no placed symbols).
   `template_with_symbols.kicad_sch` is kept unchanged in-repo as a test fixture.
   A regression test asserts a created schematic contains no `_TEMPLATE_`
   references and no seeded `lib_symbols` entries.
--- a/tests/test_pin_locator_module_leak.py
+++ b/tests/test_pin_locator_module_leak.py
@@ -1,0 +1,52 @@
+"""Regression test for issue #287: cross-file sys.modules pollution.
+
+``tests/test_rotate_schematic_mirror.py`` used to install a throwaway
+``MagicMock`` at ``sys.modules["commands.pin_locator"]`` via
+``sys.modules.setdefault(...)`` at module-collection time, with no teardown.
+Any later-collected file relying on the real ``commands.pin_locator`` (e.g.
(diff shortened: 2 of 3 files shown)


mixelpixx:
Verified and merging. What was checked locally on this branch merged with main:

- The exact #287 repro pair (`test_rotate_schematic_mirror.py` + `test_move_with_wire_preservation.py`) goes from 6 failed on main to 50 passed here, and the new subprocess regression test passes.
- Full suite: 1365 passed with only the 8 known environmental failures (freerouting/autoroute, no Java on this machine). black clean.
- The teardown logic is right: `_had_real_pin_locator` correctly preserves a pre-existing real module (matching `setdefault` semantics), and the identity check prevents deleting a module someone else installed. The `commands.*`-scoped eviction with the `_modules_before` snapshot is the correct scope — evicting third-party modules imported as side effects would not be safely re-importable.

One adjustment made on the branch before merge (pushed as 671fd6b): the inner pytest invocation passed `--timeout=60 --timeout-method=thread`, but pytest-timeout is not a declared dependency of this repo — on a clean environment the subprocess exits with a usage error (code 4) and the regression test fails for the wrong reason. The flags are dropped; the outer `subprocess.run(timeout=90)` already provides the hang protection, so nothing is lost. If you rely on pytest-timeout locally, proposing it in requirements.txt as a dev dependency would be a separate, reasonable PR.

Also restored two spaces of indentation this branch accidentally dropped from an unrelated CHANGELOG continuation line (the `"eeschema")` line in the #221 entry) — worth checking what in your tooling keeps reformatting that line, since the same change shows up in your #310 as well.

PR #313, opened 2026-07-07
Run Freerouting headless so autoroute doesn't open a GUI window

danielsvane:
On a normal desktop session, running `autoroute` opens Freerouting's GUI window on every call. It still routes and imports the SES fine, but the window pops up and grabs focus each run, which gets old fast.

Freerouting only switches to headless on its own when it can't find a display. When there's no display it prints:

> Couldn't get screen resolution. If you are running in a headless environment, disable the GUI by setting gui.enabled to false.

So in the Docker path (no display forwarded) and in CI the window never appears, which is probably why this hasn't surfaced. On a desktop it shows up every time.

This passes `--gui.enabled=false` in `_build_freerouting_cmd` for both the direct and Docker invocations, as the warning suggests. It only controls whether the GUI is created — routing and the SES output are unchanged.

Tested with Freerouting 2.1.0 on Linux; autoroute now completes without opening a window.


mixelpixx:
Verified and merging. What was checked:

- `--gui.enabled=false` is the settings-override syntax Freerouting 2.x itself suggests in the exact warning you quoted, and this repo targets the 2.x line (module docstring requires Java 21+; the Docker path runs `eclipse-temurin:21-jre`), so the option exists in every supported configuration. The flag is added to both the direct and Docker invocations, which is the right symmetry — the Docker path only avoids the window today because no display is forwarded, not because the GUI is disabled.
- No test asserts the exact argv (the autoroute best-of-N tests locate `-mp` positionally, so inserting a flag ahead of it is inert), and the full suite on this branch merged with main shows only the 8 known environmental failures (freerouting/autoroute, no Java on this machine). A dedicated test was not required here: exercising the flag needs a live Java + Freerouting install, which the suite already treats as environment-gated.
- black and mypy clean.

One residual note, not blocking: anyone pointing `FREEROUTING_JAR` at a 1.x jar will now get an unknown-option failure — but 1.x was never supported by this integration (Java 21+ requirement), so that configuration was already broken, just later and less clearly.

Issue #302, opened 2026-07-05
export_dsn: Specctra DSN omits all .kicad_pro net classes — every net exported at Default width (silent hazard via one-call autoroute) — v2.3.0


brunomariani:
## Environment
- KiCad 10.0.3, **macOS 15** (cross-platform data point vs. the usual Windows reports)
- MCP server **v2.3.0**, **SWIG** backend (`get_backend_state` → `swig`)
- Consumer: Freerouting v2.0.1 (Java 21 Temurin), but the defect is entirely in `export_dsn`.

## Severity
**High — silent electrical hazard.** `export_dsn` returns success and emits a structurally valid DSN, but every net carries the Default class width. The one-call `autoroute` tool re-exports internally, so a power net is handed to Freerouting at signal width **with no error or warning** and comes back routed too thin.

## Symptom
With a non-default net class defined in `.kicad_pro` (`net_settings.classes`) — here a 2.0 mm power class over nets `PWR`/`PWR2` — `export_dsn` emits a **single** `(class kicad_default …)` listing *every* net at the 0.2 mm Default width. The power class never appears in the DSN.

Crucially, this is **not** the `create_netclass` persistence gap (#185 / #232): the class was written into `.kicad_pro` and the project reloaded via `open_project` **before** export, so it was genuinely on disk. The DSN exporter simply never reads it.

## Why it matters
`autoroute` re-exports the DSN internally, so it inherits this drop silently. On our test board the ≈4.8 A main path would be routed at 0.2 mm. There is no signal that class rules were lost — the DSN looks complete.

## Root cause (hypothesized)
The SWIG `pcbnew` `LoadBoard()` path loads only `.kicad_pcb`, so `BOARD_DESIGN_SETTINGS.m_NetSettings` carries only the stock Default class. Net-class *definitions* live in `.kicad_pro` (`net_settings.classes`) on KiCad 7+, which this path never reads — this is the **read-side analog** of the write-side gap acknowledged in #185 ("the SWIG board API doesn't serialize project settings"). A conformant `.kicad_pro` doesn't help: the classes exist in the project file; the DSN exporter just doesn't consult it.

## Expected
`export_dsn` should emit each defined net class as its own `(class <name> <member-nets…> (rule (width …) (clearance …)))` — with the class's members and its width/clearance/via rules — matching what KiCad's own **File → Export → Specctra DSN** writes.

## Workaround (validated)
[...]

PR #315, opened 2026-07-08
fix: implement assign_net_to_class and check_clearance, remove dead add_net_class duplicate

boovaragan:
## Summary

Found while auditing DRC/design-rule tool documentation for a docs contribution:
**4 of the 8 "Design Rules and DRC" tools were dead.** `add_net_class`,
`assign_net_to_class`, `set_layer_constraints`, and `check_clearance` each had
a full Zod schema in `design-rules.ts` and were listed in the router's `drc`
category — but none had an entry in `kicad_interface.py`'s command dispatch
table. Calling any of them always returned
`{"success": false, "message": "Unknown command: ..."}`. Not tracked in
`KNOWN_ISSUES.md`.

This PR fixes 2 of the 4, removes 1 dead duplicate, and explicitly flags the
4th as needing a different, larger fix.

### `assign_net_to_class`

Assigns an existing net to an existing net class. Mirrors `create_netclass`'s
established dual-write shape (`python/commands/routing.py`): a best-effort
in-memory `NETINFO_ITEM.SetClass()` for the live SWIG session, plus a durable
write to `<project>.kicad_pro`'s `net_settings.netclass_assignments` — net
class *membership* lives in the project file on KiCad 7+, not the board, the
same reasoning `create_netclass`'s persistence already established for class
*definitions* (#302). New `apply_net_assignment_to_project_settings` /
`persist_net_assignment_to_project` helpers in `routing.py`, alongside the
existing definition-persistence helpers they mirror.

### `check_clearance`

Resolves two items (by UUID, or by `reference` for components) and measures
the gap between their `GetBoundingBox()`s against the board's
`m_MinClearance`. This is the same AABB approximation
`check_courtyard_overlaps` already uses for footprint-to-footprint checks —
approximate for non-rectangular/angled items (a circular pad's bbox is its
bounding square), so it reads *closer* than the true clearance, never
*further*. It's a fast ad-hoc check, not a substitute for a full `run_drc` /
`get_drc_violations`, which uses KiCad's exact polygon-based resolver.
Position-based item lookup (vs. UUID/reference) is out of scope for this PR —
callers query the UUID first (`query_traces`, `get_component_pads`, etc.).

### `add_net_class` — removed (dead duplicate)

This one wasn't a "missing handler," it was a stale duplicate: a *different*,
working `create_netclass` tool (in the `routing` category) already covers net
class creation, and its Python implementation is actually **more capable**
than its own TS schema exposed — `uviaDiameter`, `uviaDrill`, `diffPairWidth`,
`diffPairGap`, and `nets` were all already read by
`RoutingCommands.create_netclass` but never reachable through the tool's
params. Deleted `add_net_class`'s registration and widened `create_netclass`'s
schema to expose what it already supported, instead of building a second
implementation of the same feature.

### `set_layer_constraints` — left alone, flagged

Checked the real pcbnew SWIG bindings (KiCad 10): there's no simple
per-layer minimum track/clearance/via property. Real per-layer constraints in
KiCad 9+ go through the DRC **custom rules** system (a `.kicad_dru` text file
with its own rules DSL) — a differently-shaped, larger feature than the other
three, not a quick missing-handler fix. Still registered and still
unimplemented; flagging it here rather than attempting a rushed fix.

## Tests

- `tests/test_assign_net_to_class.py` (18 tests): the pure JSON transform,
  atomic file round-trip (mirrors `test_netclass_kicad_pro_persistence.py`'s
  shape exactly), and `DesignRuleCommands.assign_net_to_class` wiring against
  a mocked board (success, missing net, missing class, missing params, no
  board, and persistence surviving a SWIG failure).
- `tests/test_check_clearance.py` (21 tests): the pure `_bbox_gap_nm` geometry
  (overlap, touching, X/Y separation, diagonal/Euclidean, symmetry),
  `_resolve_clearance_item` for every item type (component by
  reference/id/missing, track/via/pad/zone by id, unknown type), and
  `check_clearance` end-to-end against a mocked board (meets/violates
  requirement, missing items, unresolvable item, no board).
- `tests/test_mcp_error_wrapping_static.py`: updated its hardcoded
  design-rule tool list to drop `add_net_class`, matching its removal — this
  static test enforces the `formatKicadResult` wrapper convention across
  `design-rules.ts`'s tools by name, unrelated to backend-implementation
  status, so the list needs to track what's actually registered.

Full suite: no regressions vs. established baseline (5 `test_autoroute_best_of_n.py`
+ 3 `test_freerouting.py` Java-env + 1 `#287`-adjacent ERC-roundtrip flake, all
pre-existing and environment-specific).

## Verification note

`assign_net_to_class`'s in-memory `NETINFO_ITEM.SetClass()` call and
`check_clearance`'s `GetBoundingBox()` calls are exercised only against a
mocked `pcbnew.BOARD` here — I can't construct a live board in this
environment to confirm end-to-end against real KiCad. Both mirror method
calls and patterns already proven working elsewhere in this codebase
(`create_netclass`'s identical `SetClass` call, `check_courtyard_overlaps`'s
identical bounding-box approach), so the risk is low, but a real-board
retest would close the loop.


mixelpixx:
Ownership decision: this PR is the canonical implementation of `assign_net_to_class` and `check_clearance`, including the removal of the dead `add_net_class` registration. #314 carries a competing netclass slice — its `assign_net_to_class` binds to `routing_commands` rather than `design_rule_commands`, and it keeps `add_net_class` as a deprecated alias — and has been asked to drop that slice and re-propose its `create_netclass` pattern-persistence enhancement separately after this merges (details in the comment there). The reason a decision was needed rather than merging both: both PRs add the same `"assign_net_to_class"` key to kicad_interface.py's dispatch dict, and a duplicated dict key silently resolves to whichever entry is added last, so merge order would have picked the winning implementation with no error anywhere. Nothing to change here; full review to follow.

mixelpixx:
Status note from the automated queue pass: this branch now conflicts with `main` and needs a refresh before it can be reviewed or merged.

Your last push (`6900771`, merging `upstream/main`) came in after the ownership comment above, so the branch is up to date with `main` as of 2026-07-12. Since then #327 and #328 merged, and one of them collides with this branch.

There are exactly two conflicts, both easy:

1. `python/commands/eagle.py` — this is the interesting one, and it is a case where `main` has superseded a workaround in this branch. You added:

```python
Path(sch_out).write_text(
    prettify(text), encoding="utf-8", newline="\n"
)  # type: ignore[call-arg]  # newline= is valid on Path.write_text (3.10+); outdated typeshed stub
```

The `type: ignore` was correct reasoning about mypy, but the underlying problem was real rather than a stub bug: `Path.write_text(newline=)` genuinely does not exist before Python 3.10, so on a 3.9 interpreter that line raises `TypeError` at runtime. #328 fixed it properly by switching to an explicit `Path.open()` with `newline="\n"`, which works on every supported version and needs no ignore comment. So when resolving, **take mains side of this hunk in full** and drop your two-line change plus the `type: ignore`. If you keep your version, you re-introduce the 3.9 crash that #328 just removed.

2. `CHANGELOG.md` — the usual Unreleased-section adjacency from several PRs landing in the same wave. Resolve by keeping both entries.

Nothing else conflicts: `python/kicad_interface.py` and `python/schemas/tool_schemas.py` both auto-merge cleanly, so the substance of the PR (the `assign_net_to_class` / `check_clearance` implementations and the dead `add_net_class` removal) is unaffected.

To be clear on where this sits: the ownership decision above still stands — this PR is the canonical home for that work, and it is queued for an interactive review rather than waiting on anything else from you beyond the rebase. Once the branch is green against `main` again it goes into the review rotation.

mixelpixx:
Merged. Your dead-code finding was correct and I verified it independently — `add_net_class`, `assign_net_to_class` and `check_clearance` all had Zod schemas and router entries with no Python route, so every call returned `Unknown command`. Good audit.

I also confirmed the part I was most worried about, and it is good news: your storage choice is **consistent** with #302 rather than in conflict with it. You write `net_settings.netclass_assignments[net] = [class]`, and `utils/project_netclasses.py:78-92` reads exactly that key and shape back to rebuild NET_SETTINGS before `ExportSpecctraDSN`. So this fills the membership gap `create_netclass` explicitly leaves out of scope, and DSN export now picks assignments up. Nice fit.

Four things I resolved in the merge rather than bouncing back, one of which you should know about because it was blocking:

**1. `assign_net_to_class` did not work on any real project.** A freshly written `.kicad_pro` has `"netclass_assignments": null` — that is KiCad's own output, captured from real pcbnew in `utils/kicad_project.py:109`. `setdefault("netclass_assignments", {})` returns that existing `None` instead of replacing it, so `assignments[net_name] = [...]` raised `TypeError`. Because `persist_net_assignment_to_project` catches broadly, it degraded to `persisted: false` with a warning — a silent no-op rather than an error the caller would notice.

Your tests could not catch it: the fixture uses `"netclass_assignments": {}` and the other case tests the key being *absent*. Present-but-null is the shape that actually ships. Fixed by normalizing to a dict explicitly, and I added the null cases to `tests/test_assign_net_to_class.py`.

I also added `tests/test_netclass_assignment_roundtrip.py`, which pins the seam between this PR and #302 — write via your persistence path, read via the DSN loader. Neither side's tests crossed that boundary, which is precisely how the null bug survived two green suites.

**2. The `eagle.py` hunk was dropped, and it is worth knowing why.** You changed `Path.open("w", ..., newline="\n")` to `Path.write_text(..., newline="\n")` with a `type: ignore[call-arg]` and a comment saying the stub was outdated. The stub is right: `newline=` on `write_text` is **3.10+**. Main's form is #328's fix for exactly that — every `.kicad_sym` and schematic write was broken on the project's declared 3.9 floor. Taking your hunk would have reverted it. CI now runs the 3.9 matrix (#334), so this would have failed rather than shipped.

**3. CHANGELOG.** Your entry targeted what was the unreleased section when you branched; that shipped as 2.4.0 on 07-22. I moved it into a fresh `[Unreleased]` rather than backdating it into a released version, and dropped the "`set_layer_constraints` remains registered but unimplemented" sentence since your #323 implements it — merging that next.

**4. Stale doc references.** The `add_net_class` removal left mentions in `README.md:490`, `docs/TOOL_INVENTORY.md:110`, `docs/ROUTER_ARCHITECTURE.md:128` and `docs/ROUTER_QUICK_START.md:56`. Repointed them at `create_netclass`.

Full gates: 1519 passed / 4 skipped, black and mypy clean, `npm run build` and `test:ts` green, schema↔route parity guard passes.

On to #323 — I will rebase it on this and union-resolve the four shared files.

PR #316, opened 2026-07-08
feat: add replace_instance_lib_ids MCP tool for schematic migration

karu2003:
## Summary

Adds a generic MCP tool for replacing `lib_id` references in schematic instances — the mechanical layer of library migration (e.g., `eagle_import` → `FOG_components`).

## Changes

- **`python/commands/symbol_schematic.py`** — new `SymbolSchematicCommands` class with `replace_instance_lib_ids` method
- **`python/kicad_interface.py`** — register new command in dispatch
- **`tests/test_symbol_schematic.py`** — 6 tests covering replacement, mirror-variant angles, error handling, and lib_symbols preservation

## What it does

- Swaps `lib_id` references in schematic **instances** only (preserves `lib_symbols` section)
- Handles Eagle mirror-variant suffixes (`__m0`, `__m90`, `__m180`, `__m270`) with automatic angle correction
- Per-instance mapping dict: `{old_lib_id: new_lib_id}` — one eagle symbol can map to different FOG symbols depending on Value/footprint
- Returns `{success, replaced, remaining_eagle}`

## What it does NOT do

- No semantic matching logic (type, nominal, footprint) — that belongs in callers/skills
- Does not modify `lib_symbols` — use `update_sch_symbols` for that
- Does not validate the target symbol exists in the library

## Test plan

- [x] `test_basic_replacement` — plain lib_id swap
- [x] `test_mirror_variant_angle` — `__m90` suffix adds 90° to original angle
- [x] `test_missing_file` — error on non-existent file
- [x] `test_empty_mapping` — error on empty mapping
- [x] `test_no_remaining_eagle` — verifies 0 remaining after replacement
- [x] `test_lib_symbols_preserved` — lib_symbols section untouched


mixelpixx:
Sequencing note: #317 (next PR in this series) copies this PR's kicad_interface.py wiring — the symbol_schematic import, the instantiation, and the `replace_instance_lib_ids` handler entry — but does not include `symbol_schematic.py` itself, so #317 depends on this PR merging first. Review will treat this PR as the base of the series; nothing to change here. Details in the comment on #317.

mixelpixx:
Verified and merging — the tool fills a real gap (the mechanical layer of an eagle_import-to-curated-library migration, complementing the importer and update_symbol_from_library), the lib_symbols/instance boundary split is the right design, and the tests were a solid base. Four completions were made on the branch before merge (see the commit for full detail); flagging them here because two would have been invisible until someone hit them:

1. The tool was unreachable over MCP. Registration in kicad_interface.py's dispatch dict is one of three required layers — without a tool_schemas.py entry and a TS `server.tool` registration, no MCP client ever sees the tool. All three are now wired, and the interface-construction sentinel test includes the new name so a future rebase can't silently drop it. Worth noting for #317, which has the same gap.

2. Mapping values are now used verbatim. The original resolved each target as `targetLibrary` + the mapping value's symbol name, which silently rewrote every entry into a single library even when the mapping's own values said otherwise — and a real migration legitimately fans out (passives to Device, connectors to a project library). The redundant targetLibrary parameter is gone; the mapping is the whole contract. Regression test included.

3. The instance-header regex required a newline between `(symbol` and `(lib_id` — that matches KiCad's own writer and the Eagle importer, but this repo's dynamic loader writes single-line headers, so instances placed by add_schematic_component were silently unmigratable (replaced=0, no error). Separators widened to `\s+`, regression tests for both layouts.

4. Convention alignment: camelCase params (schematicPath, sourceLibrary) like every other tool in the dispatch, neutral defaults and docs (the FOG_components default and the Hermes reference were your environment leaking into a general-purpose tool — sourceLibrary keeps the eagle_import default since that is this repo's importer output), pathlib instead of os.path (main went through a pathlib sweep this week), the shared kicad_interface logger, error detail in the response, and the test fixture header moved 20260306 to 20260101 — KiCad 10.0.x refuses files claiming a newer format token than its writer, and fixtures get copy-pasted.

Gates: 8 tool tests pass (6 original semantics preserved, 2 new), interface construction green, full suite at the known environmental baseline, black/mypy clean, npm build + TS tests green.

PR #317, opened 2026-07-08
feat: add library management MCP tools (import, export, rename, delete)

karu2003:
## Summary

Adds MCP tools for managing symbols in `.kicad_sym` libraries — import, export, rename, and delete. This lets KiCad manage its own library through MCP, solving the problem of missing or broken components.

## New MCP Tools

| Tool | Description |
|------|-------------|
| `import_symbol` | Copy a symbol from one `.kicad_sym` to another, with optional rename and overwrite |
| `export_symbol` | Extract a single symbol to a standalone `.kicad_sym` file |
| `rename_symbol` | Rename a symbol and all subsymbol references (`name_0_1`, `name_1_1`) |
| `delete_symbol` | Remove a symbol from a library, preserving file structure |

## Implementation

- **`python/commands/library_management.py`** — `LibraryManagementCommands` class
- Pure S-expression text parsing with parenthesis-depth tracking
- Only operates on top-level symbols (nesting level 1 inside `kicad_symbol_lib`)
- Creates target library if missing (with KiCad 10 version header)
- All writes use `newline="\n"` to prevent CRLF drift

## Registration

Registered in `kicad_interface.py` dispatch:
```
"import_symbol": ...
"export_symbol": ...
"rename_symbol": ...
"delete_library_symbol": ...
```

## Test plan

- [x] `test_export_symbol` — export to standalone file
- [x] `test_export_symbol_not_found` — error on missing symbol
- [x] `test_import_symbol` — import into new library
- [x] `test_import_symbol_with_rename` — import + rename in one step
- [x] `test_import_symbol_duplicate` — fails without overwrite
- [x] `test_import_symbol_overwrite` — replaces existing
- [x] `test_rename_symbol` — renames symbol + subsymbols
- [x] `test_rename_symbol_collision` — fails if target exists
- [x] `test_delete_symbol` — removes symbol, keeps others
- [x] `test_delete_symbol_not_found` — error on missing
- [x] `test_delete_preserves_library_structure` — balance = 0 after delete

**11/11 tests pass.**


mixelpixx:
Coordination note ahead of the full review (about how this lands, not the feature itself): this branch currently cannot merge on its own, and will conflict with #316 if that merges first.

The kicad_interface.py hunks here include three lines that belong to #316: the `from commands.symbol_schematic import SymbolSchematicCommands` import, the `self.symbol_schematic_commands = SymbolSchematicCommands()` instantiation, and the `"replace_instance_lib_ids"` handler entry. But `python/commands/symbol_schematic.py` itself is only added by #316 — it is not in this PR's file list. Two consequences:

- If this PR merged first, kicad_interface.py would raise ModuleNotFoundError at import time — the import names a module that does not exist on main yet — and that file is the dispatch hub, so every MCP tool goes down with it.
- If #316 merges first (the sensible order), these duplicated lines conflict with the identical ones #316 adds at the same spots.

Suggested fix: rebase onto #316 (or onto main once #316 merges) and drop the three symbol_schematic lines from this PR's kicad_interface.py diff, so this PR only adds the LibraryManagementCommands wiring. That makes it independently reviewable and keeps each tool's wiring attributable to the PR that introduces it.

One more thing to address in the same pass, because it interacts with two other open PRs: the new import/export/rename/delete tools mutate library files on disk, while #320 adds module-level caches for library discovery and symbol extraction and #310 caches a LibraryManager instance. None of those caches would be invalidated by a mutation from these tools, so a rename or delete followed by a read could serve stale data. Worth stating in this PR how (or whether) it invalidates those caches, so the three PRs can be reviewed against each other.

mixelpixx:
Verified and merging. Since the coordination comment earlier today asked you to rebase onto #316 and drop the duplicated wiring: no action needed anymore — #316 merged first and the dedup was handled here during the merge, so don't spend time on it. The import/export/rename capability is a real gap filled (create/delete existed; moving symbols between libraries did not), and the parenthesis-depth parsing approach matches how the rest of this codebase treats .kicad_sym files.

Beyond the #316-wiring dedup, the completions made on the branch (one commit, full reasoning in its message):

1. Same reachability gap as #316: dispatch-dict registration alone makes a tool invisible to MCP clients. The three tools are now registered in symbol-creator.ts alongside create/delete_symbol and added to the interface-construction sentinel test.

2. The delete tool was dropped from this PR's surface — not because it was wrong, but because main already ships `delete_symbol` (SymbolCreator), and two deletes with different names and different param spellings is exactly the dispatch ambiguity the #314/#315 netclass decision resolved this week: one tool per capability. Your structure-preservation test was the best of the delete tests, so it lives on — pointed at the canonical tool.

3. These tools rewrite `.kicad_sym` files, and since #320 merged today, symbol resolution/extraction is cached process-wide — a library edit that doesn't invalidate those caches can serve stale symbol data to the next `add_schematic_component`. Every write path now calls the shared invalidation hook, with a regression test.

4. Two correctness items in rename: `(extends "old")` references from derived symbols elsewhere in the library are now repointed — renaming a parent used to orphan its children silently (the #282 failure class; the response reports `extends_updated`). And the sub-symbol rename is anchored on `(symbol "` so a property VALUE that happens to start with the symbol name cannot be rewritten.

5. Format hygiene: new-library and export headers now reuse SymbolCreator's `KICAD9_SYMBOL_LIB_VERSION` (20241209) instead of the unpinned `20251024` — this repo learned the hard way (v2.3.0, #221) that KiCad refuses files claiming a newer format token than the running build writes, so every `.kicad_sym` the server produces carries one shared, oldest-supported token from one constant. Also: the top-level matcher no longer accepts bare sub-symbol shards (`R_10K_0603_0_1`) as importable units, a delete-splice off-by-one that could eat a closing paren is fixed, params are camelCase like the rest of the dispatch, and the module docstring no longer claims a kicad-cli validation pass that was never implemented.

Gates: 12 library-management tests pass (your 8 surviving semantics preserved, 4 new), interface construction green, full suite at the known environmental baseline, black/mypy clean, npm build + TS tests green.

The import/rename pair combined with #316's replace_instance_lib_ids now forms a complete migration workflow: import or rename in the library, then repoint the placed instances. Good series.

PR #319, opened 2026-07-12
fix: snap component placement origin to 1.27 mm grid

bluebirdlboro:
## Summary

Components placed via `add_schematic_component` at raw millimetre coordinates land off the 1.27 mm (50 mil) schematic connection grid. Library pins sit at integer multiples of 1.27 mm from the origin, so off-grid origins leave every pin off-grid — wires and net labels cannot bind, ERC reports `endpoint_off_grid`, and the netlist comes up empty.

This PR snaps the symbol origin to the 1.27 mm grid inside `create_component_instance` so pins always land where KiCad expects them.

## Changes

- **`dynamic_symbol_loader.py`**: Snap `x`, `y` to the 1.27 mm grid before writing the symbol block. Log when snapping occurs.
- **`schematic_handlers.py`**: Read back the actual placed position from the file and include `placed_at` (and `requested_at` when they differ) in the handler response, so callers can see when coordinates were adjusted.
- **`test_add_schematic_component.py`**: Two new tests — off-grid coordinate is snapped, on-grid coordinate passes through unchanged.

## Design decision

**Snap + report**: The snap is silent (always on), but the handler response includes `placed_at` and, when different from the request, `snapped: true` + `requested_at`. A caller who asked for `x=100` can see it landed at `x=100.33`.

Closes the grid-snap portion of #299.


mixelpixx:
Verified and merging — this closes the grid-snap half of #299 (the cache half is under review in #320). Two adjustments were made on the branch before merge (ded08fd), both in the read-back/reporting layer; the snap itself was correct as submitted.

What was verified:

- The snap math and placement: off-grid coordinates land on 1.27 mm multiples, on-grid coordinates pass through bit-identical, and the mirror-canary and eeschema-oracle suites still pass, so the transform pipeline is unaffected.
- The design decision (silent snap + report) matches what we asked for in #299: electrical correctness is not optional, so refusing off-grid placements would only push the burden onto every caller, while `placed_at`/`snapped`/`requested_at` keeps the adjustment observable.

The two adjustments:

1. The read-back located the placed instance by the FIRST lib_id match in the file. With two instances of the same symbol — the most common case there is, two resistors — every placement after the first reported the first instance's coordinates as `placed_at`, plus a spurious `snapped: true`/`requested_at` pair. For the primary consumers of this metadata (AI callers deciding whether to re-place), wrong metadata is worse than none. The lookup is now `_find_placed_symbol_position`, which scans symbol headers and matches the block containing the just-placed `(property "Reference" ...)`; a regression test covers the two-resistor case.

2. `round(x / 1.27) * 1.27` carries float representation dust — `79 * 1.27` is `100.32999999999998` — and that full repr was written into the file verbatim, only to be reformatted by KiCad on its next save (a needless diff in version control). Rounding the product to two decimals is exact, not approximate: every multiple of 1.27 is `k * 127/100` and has at most two decimals. A test asserts the file contains the literal `100.33`.

Also added the CHANGELOG entry (house rule: every user-visible change gets one). Full suite on this branch merged with main: only the 8 known environmental failures. black and mypy clean.

PR #320, opened 2026-07-12
perf: add module-level caches for library discovery and symbol extraction

bluebirdlboro:
## Summary

`DynamicSymbolLoader` creates a fresh instance for every `add_component` call, so its instance-level caches never survive — each component add re-scans the sym-lib-table and re-reads multi-MB `.kicad_sym` files. With 200+ installed libraries this becomes super-linear in the test suite (300s+ timeout vs ~15s with caches).

This PR adds module-level caches shared across all loader instances so library discovery, file resolution, and symbol extraction each pay their cost once per process.

## Changes

- **`dynamic_symbol_loader.py`**:
  - `find_kicad_symbol_libraries()`: cache the directory list module-wide (`_LIB_DIRS_CACHE`)
  - `find_library_file()`: cache by `(project_path, library_name)` so repeated lookups within the same project don't re-parse the sym-lib-table
  - `extract_symbol_from_library()`: cache by `(resolved_lib_path, symbol)` so the multi-MB file is read and parsed once per process (`_SYMBOL_BLOCK_CACHE`)
  - `clear_library_caches()`: static method to reset all caches when libraries change on disk mid-session
- **`test_symbol_library.py`**: Two `__new__`-bypassing fixtures were missing `_cache_lock` (only surfaced when Simulation_SPICE is installed, i.e. kicad-symbols 9.0.9+)
- **`test_kicad10_paths.py`**: Isolate env-override tests from a pre-existing `KICAD9_*_DIR` in the process environment that could shadow the KICAD10 monkeypatch

Extracted from #299 (symbol-cache portion).


mixelpixx:
Verified and merging — this closes the cache half of #299 (the grid-snap half merged as #319). The caching design is right and the numbers justify it; one class of issue was completed on the branch before merge (cdf4704): the caches assumed libraries are immutable during a process run, and they aren't.

What was verified as submitted:

- The mechanism diagnosis is correct and worth having in the history: a fresh `DynamicSymbolLoader` per `add_component` call and a fresh `SymbolLibraryManager` (plus warm-up thread) per `KiCADInterface` meant instance caches never survived, and N interfaces spawned N concurrent full-library parse threads. Module scope is the right level for this data.
- `_WARM_STARTED` once-per-process and the `KICAD_SKIP_SYMBOL_WARMUP` test lever are both sound, and the conftest wiring makes interface-heavy tests dramatically cheaper.
- The two test repairs (missing `_cache_lock` on `__new__`-bypassing fixtures, `KICAD9_*_DIR` env leakage into the KICAD10 monkeypatch tests) are correct fixes to real latent problems.

What was added, and why it could not merge without it: four MCP tools mutate symbol libraries mid-session — `create_symbol`, `delete_symbol`, `add_symbol_property` (all rewrite `.kicad_sym` files) and `register_symbol_library` (rewrites the sym-lib-table). Against those, the unguarded caches had a particularly bad failure mode: `find_library_file` cached `None` for an unresolved name, so the natural flow "create library, register it, place a symbol from it" stayed at 'library not found' until process restart — and `clear_library_caches()` existed but nothing called it. The completion keeps your structure and adds the cheapest sufficient guard per cache:

- `_LIB_FILE_CACHE`: misses are never cached; hits are revalidated with `exists()` so deleted/moved libraries re-resolve.
- `_SYMBOL_BLOCK_CACHE` and `_SYMBOL_CACHE_BY_PATH`: entries carry the source file's `mtime_ns` and expire when it changes. A `stat` per hit is noise next to the multi-MB read/parse it saves.
- The mutating write paths now call `clear_library_caches()` — this covers the case a per-file stat cannot see, re-pointing an existing library name at a different path via re-registration.
- One threading nit: the shared-cache hit path in `list_symbols` wrote `self.symbol_cache` without taking `self._cache_lock` while the cold path below it did; both now take it.

A second commit (on the branch) was forced by the full suite itself: with the caches active, `tests/test_symdir_extends.py` started failing when run after any test that had resolved the same library name — each of those tests points an individual loader's `find_kicad_symbol_libraries` at its own temp dir, and that per-instance state is invisible to any module-level cache key, so the first test's resolution was served to the next. Two changes: loaders whose discovery is instance-patched now bypass the shared cache entirely (production never shadows the method), and the `KICAD*_SYMBOL_DIR` env values — which are inputs to discovery — are now part of the cache validity, so an env change invalidates instead of replaying the old environment's resolution. Worth internalizing for future caches in this codebase: the key must capture everything the computation depends on, or the cache turns passing tests into order-dependent ones — exactly the #287 class of failure this repo just finished cleaning up.

Known residual, judged acceptable and documented in the code: for a symdir symbol whose `(extends ...)` parent lives in a sibling shard, a parent-only edit is invisible to the child-file stat — the explicit clear from the mutating handlers covers the in-process tools, and external shard edits of that specific shape are far enough down the tail to not carry a per-extract directory scan.

Gates: the new invalidation regression tests plus your two repaired test files pass; full suite on this branch merged with main shows only the 8 known environmental failures; black and mypy clean. With this merged, both halves of the #299 split have landed — thanks for carrying the split through cleanly.

PR #322, opened 2026-07-12
fix: prevent CHANGELOG.md inline code spans from crossing line breaks

boovaragan:
## Summary

Root-cause fix for a formatting issue @mixelpixx flagged twice (once on #310,
once on #312): a CHANGELOG entry kept losing 2 spaces of indentation every
time it got reformatted, and had to be manually restored on both merges.

Traced it: the affected lines all had an **inline code span that crosses a
line break** — e.g. `` `(version 1) (generator\n"eeschema")` `` (opening
backtick on one line, closing backtick on the next). Prettier's markdown
list-continuation indentation logic mis-handles the line that starts mid-span,
dropping its leading indentation. Confirmed reproducible: take a known-good
CHANGELOG.md, add *any* unrelated new entry above the affected paragraph, run
`prettier --write`, and the affected line loses its indent — every time.

This isn't a one-off. A full scan of the file found **three** live instances
of the same pattern, two of which are already broken on `main` right now (not
just the one that got caught in review):

1. The `#221` blank-template entry (the one caught twice already)
2. The `#221, partial` fallback-header entry (`generator_version "10.0"`) —
   currently broken on `main`
3. The `add_gnd_stitching_vias` entry's `Returns \`{ placed: ... }\`` line —
   currently broken on `main`

## Fix

All three: keep the inline code span on one physical line, wrapping (if
needed) at a point *outside* the backticks instead of inside them. Verified
each is now stable — ran `prettier --write` twice in a row on the result,
zero further changes both times (true idempotency, not just "looks right
once").

## Regression test

`tests/test_changelog_no_multiline_code_spans.py`: scans CHANGELOG.md for any
inline code span (odd backtick count on a line, continuing into the next
non-fenced-code-block line) and fails with the exact line range if found.
Confirmed it fails on the original (broken) content and passes after the fix,
by temporarily reverting one instance and re-running.

## Verification

Full suite not needed — this only touches CHANGELOG.md and adds one new,
self-contained static test. Ran the new test standalone (pass), plus
`prettier --write` against the real file twice (no changes either time).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -228,8 +228,8 @@ load on every KiCad 10.0.x build.
   `add_schematic_component` tool synthesizes its own `lib_symbols` via the
   dynamic loader (and the legacy fallback was removed in #288), so the seeds only
   leaked into user files. Both tools now copy a new blank KiCad 10 template
-  (`python/templates/blank.kicad_sch`: `(version 20260101) (generator
-  "eeschema")`, empty `lib_symbols`, no placed symbols).
+  (`python/templates/blank.kicad_sch`: `(version 20260101) (generator "eeschema")`,
+  empty `lib_symbols`, no placed symbols).
   `template_with_symbols.kicad_sch` is kept unchanged in-repo as a test fixture.
   A regression test asserts a created schematic contains no `_TEMPLATE_`
   references and no seeded `lib_symbols` entries.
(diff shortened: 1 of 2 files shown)


mixelpixx:
Verified and merging — this is exactly the kind of contribution that saves maintainer time compounding forever: the same line got hand-restored during two separate merges before your trace, and the diagnosis explains why it kept coming back (the "correct" original formatting itself contained the multiline span, so every prettier pass re-broke it). Fixing all three live instances rather than just the flagged one, and proving idempotency by running prettier twice, is the right standard of done.

Verified here: clean merge over today's eight new CHANGELOG entries, the lint passes against the merged file, and the full suite is at the known environmental baseline.

One small hardening added on the branch before merge: the lint skipped fence delimiter lines but did not track fence state, so a lone backtick inside a fenced code block's content would have registered as an open inline span and failed the lint on an unrelated entry with a confusing message. It now toggles on the delimiters and exempts fenced content entirely — which is what your comment already said it intended.

PR #323, opened 2026-07-12
feat: implement set_layer_constraints via .kicad_dru custom rules

boovaragan:
## Summary

Fixes the fourth dead DRC tool flagged in the audit that produced #315:
`set_layer_constraints` was a registered MCP tool (full Zod schema in
`design-rules.ts`, listed in the router's `drc` category) with no entry in
`kicad_interface.py`'s command dispatch table, so every call silently
returned `{"success": false, "message": "Unknown command: set_layer_constraints"}`.
#315 deliberately deferred this one, since it's a differently-shaped problem
from `assign_net_to_class`/`check_clearance`: there is no pcbnew SWIG API for
per-layer DRC constraints at all.

## Research (done before writing any code)

Checked the real installed KiCad 10 SWIG bindings
(`pcbnew.py`) directly: no `DRC_RULE`/rules-parsing class exists anywhere in
the Python API. Real per-layer/conditional constraints in KiCad 9+ live
entirely in a **project-scoped `.kicad_dru` text file** — an S-expression
rules DSL, sibling to the `.kicad_pcb`/`.kicad_pro` with the same base
filename (confirmed against two real KiCad demo projects). `kicad-cli pcb
drc --help` has no `--rules`/`--dru` flag — it auto-discovers the sibling
file by convention, so `run_drc`/`get_drc_violations` need zero changes to
pick up whatever this tool writes.

Pulled the authoritative constraint-type vocabulary from KiCad's own
`DRC_RULES_PARSER` Doxygen docs, then **verified the exact syntax against
real KiCad 10** rather than guessing: wrote a rule with a deliberately
absurd 5mm minimum onto a real KiCad demo board (`vme-wren`) and ran
`kicad-cli pcb drc` directly. All four constraint types this tool needs —
`track_width`, `clearance`, `via_diameter`, `hole_size` — were recognized
and enforced, with violations explicitly citing the rule by name:

```
"Clearance violation (rule 'mcp_layer_constraint_F.Cu' clearance 5.0000 mm; actual 1.8013 mm)"
"... rule 'mcp_layer_constraint_F.Cu' min diameter 5.0000 mm"
"... rule 'mcp_layer_constraint_F.Cu' min hole 5.0000 mm"
"... rule 'mcp_layer_constraint_F.Cu' min width 5.0000 mm"
```

## Fix

New `python/utils/kicad_dru.py`: pure text-transform + persistence, no live
board interaction at all (matches the "no SWIG API" finding above).

- `build_layer_constraint_rule(layer, ...)` builds a
  `(rule "mcp_layer_constraint_<layer>" (layer "<layer>") (constraint ...))`
  block. The rule name is deterministic per layer, so a second call for the
  same layer **replaces** rather than duplicates.
- `apply_layer_constraint_rule` does a surgical find-and-replace-by-name (or
  append) on the existing file's text — paren-matched block boundaries,
  mirroring the same technique this codebase already uses for `.kicad_sch`
  edits (`schematic_text_utils.py`'s `_find_matching_paren`) rather than a
  full parse/reserialize. A hand-edited file's other rules, comments, and
  formatting outside our own named rule survive untouched.
- `persist_layer_constraint_rule` does the atomic read/modify/write
  (temp file + `os.replace`), creating the file with a `(version 1)` header
  if it doesn't exist yet.
- `DesignRuleCommands.set_layer_constraints` (`design_rules.py`) wires this
  up: resolves the sibling `.kicad_dru` path from the board's `.kicad_pcb`
  filename, requires `layer` plus at least one of
  `minTrackWidth`/`minClearance`/`minViaDiameter`/`minViaDrill`.

## Tests

- `tests/test_kicad_dru.py` (18 tests): the pure rule-text builder, the
  insert/replace/append transform (including idempotency, preserving
  unrelated rules before *and* after the target block, multiple layers
  coexisting), and the atomic file persistence — all headless, no pcbnew.
- `tests/test_set_layer_constraints.py` (8 tests): the `DesignRuleCommands`
  wiring against a mocked board (only `GetFileName()` is used, to derive the
  `.kicad_dru` path) — success, single-param sufficiency, replace-not-
  duplicate on a second call for the same layer, multiple layers coexisting,
  missing `layer`, no constraints provided, no board loaded, unresolvable
  path.

## Verification note

The pure text-transform and the tool wiring are fully unit-tested headless.
The rule *syntax itself* was verified for real against real KiCad 10 (see
above) — I did not add that as an automated test in this PR, since it needs
a real, valid `.kicad_pcb` fixture and this repo doesn't have one yet
(hand-crafting one without live pcbnew felt riskier than using KiCad's own
demo board, which isn't something to bundle into the repo). Happy to revisit
if a real board fixture lands here for other reasons.


mixelpixx:
Merged, on top of #315. Good call on the approach — I confirmed there genuinely is no pcbnew API for per-layer constraints, so the `.kicad_dru` text route is the only way to do this, and the surgical insert-or-replace-by-name (rather than parse/reserialize) is the right shape for a file users also hand-edit.

Three notes from the merge.

**1. The two PRs produced invalid Python when git merged them.** `design_rules.py` had one region where git found no conflict and merged both sides anyway — the result put an em-dash from your `set_layer_constraints` docstring outside any string literal, a hard `SyntaxError`. Both PRs add their new method at the same anchor (right after `get_design_rules`), which is what made a clean-looking auto-merge produce broken code.

Rather than hand-stitch it I rebuilt the file from main plus your method, so `assign_net_to_class`, `check_clearance` and `set_layer_constraints` are all intact. Worth knowing that `tests/test_interface_construction.py` caught this instantly — the construction smoke test exists for exactly this failure mode, where every module-level test still passes because nothing imports the broken module.

**2. Added a layer-name injection guard**, which was the nit I had flagged. The layer string lands inside two quoted `.kicad_dru` tokens — `(rule "mcp_layer_constraint_<layer>")` and `(layer "<layer>")` — so a `"` would corrupt the file *and* break `_find_rule_block`'s by-name lookup, meaning a later `set_layer_constraints` call on the same layer could no longer find its own rule to replace and would append a second one.

I deliberately did **not** validate against `board.GetLayerName()`, even though I initially planned to. KiCad 10 reports `F.Silkscreen` while `.kicad_dru` and `kicad-cli --layers` both accept `F.SilkS`, and user layers can be renamed — an allowlist would reject valid input. So the guard rejects only characters that cannot appear in a quoted S-expression token. Two tests cover both directions (metacharacters rejected; `F.Cu`/`F.SilkS`/`F.Silkscreen`/`In1.Cu`/`User.Comments` all accepted).

**3. CHANGELOG**: your entry targeted the pre-2.4.0 unreleased section, which shipped on 07-22. Moved to a fresh `[Unreleased]` rather than backdated. I also removed the "`set_layer_constraints` remains registered but unimplemented" sentence from #315's entry, since this PR makes it false.

Full gates: 1551 passed / 4 skipped, black and mypy clean, TS build and tests green, parity guard passes.

Both of your DRC-audit PRs are now in, and that audit was worth doing — three tools with schemas and no backend is the kind of thing nobody finds by using the software normally. If you are inclined to keep going in that direction, a systematic sweep for *any* other TS-registered tool lacking a Python route would be welcome; the parity guard covers schema-vs-route but not TS-vs-Python, so that gap can still open silently.

PR #324, opened 2026-07-13
fix(schematic): escape quotes in property values — corrupts .kicad_sch (likely the KiCad 10 corruption in the README)

PaulHubiss:
Placing any symbol whose library property value contains a double quote writes an unparseable schematic. `power:GND` is the common case — its `Description` is:

```
Power symbol creates a global label with name "GND"
```

`_property()` in `dynamic_symbol_loader.py` interpolates the value straight into the S-expression, so that inner quote closes the token early and swallows the rest of the file. KiCad and `kicad-cli` then refuse the schematic outright, with unbalanced parens.

**Repro:** `add_schematic_component` with `power:GND`, then `kicad-cli sch export svg` → *"Failed to load schematic file"*.

This may well be the root cause behind the schematic corruption on KiCad 10 that the README warns about — it needs no exotic setup, just a GND symbol, which nearly every real schematic has.

The fix escapes backslashes and quotes before emitting. The codebase already has the correct helper (`SchematicHandlers._escape_sexpr_string`); this path simply never called it. Verified on KiCad 10.0.4: a schematic with `power:GND` that previously failed to load now exports cleanly.

**Also included:** `get_board_2d_view` failed for every caller that omitted `layers`, because KiCad 10 requires `--layers` on `pcb export svg` ("at least one layer must be specified") but the flag was only passed when the list was non-empty. Defaults to `F.Cu, B.Cu, F.SilkS, Edge.Cuts` so the no-argument call renders a readable board.

Happy to split these into two PRs if you prefer.

### DIFF
--- a/python/commands/board/view.py
+++ b/python/commands/board/view.py
@@ -170,7 +170,15 @@ class BoardViewCommands:
                     "message": f"Unsupported format '{fmt}'. Use 'png', 'jpg', or 'svg'.",
                     "errorDetails": f"Got: {fmt}",
                 }
-            layers: List[str] = params.get("layers", [])
+            # KiCad 10 rejects `pcb export svg` without --layers ("at least one
+            # layer must be specified"), so an omitted layers list must not stay
+            # empty. Default to the layers that make a board readable at a glance.
+            layers: List[str] = params.get("layers") or [
+                "F.Cu",
+                "B.Cu",
+                "F.SilkS",
+                "Edge.Cuts",
+            ]
             response_mode = params.get("responseMode", "inline")
 
             kicad_cli = resolve_kicad_cli()
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -41,6 +41,16 @@ _SYMBOL_DIR_ENV_VARS = (
 )
 
 
+def _esc_sexpr(value: str) -> str:
+    """Escape a string for safe insertion into an S-expression double-quoted token.
+
+    Library property values may themselves contain quotes — power:GND's Description
+    is `Power symbol creates a global label with name "GND"`. Emitted raw, the inner
+    quote closes the token early and corrupts the rest of the .kicad_sch.
+    """
+    return value.replace("\\", "\\\\").replace('"', '\\"')
+
+
 def _symbol_dir_env_fingerprint() -> Tuple:
     """The env-var inputs library discovery depends on.
 
@@ -1009,7 +1019,7 @@ class DynamicSymbolLoader:
             """
             hide_line = "      (hide yes)\n" if hide else ""
             return (
-                f'    (property "{name}" "{value}"\n'
+                f'    (property "{_esc_sexpr(name)}" "{_esc_sexpr(value)}"\n'
                 f"      (at {_fmt(px)} {_fmt(py)} {_fmt(pa)})\n"
                 f"{hide_line}"
                 f"      (show_name no)\n"


mixelpixx:

- Full suite: **8 failed / 1420 passed**. All 8 failures are the known environmental baseline (Java not found — `test_autoroute_best_of_n` + `test_freerouting`); nothing this PR touches regresses.
- `black --check` clean on both changed files.
- The escaping logic is correct for the mechanism it targets: `_esc_sexpr` replaces backslashes *before* quotes, so a pre-existing `\` is not double-counted, and a value like `power:GND`'s Description (`... global label with name "GND"`) now serializes as `\"GND\"` instead of closing the token early and shredding the rest of the file. That is a genuine `.kicad_sch` corruption path, so the fix is worthwhile.

Two notes for the interactive review, neither a blocker on the escaping fix itself:

1. **This PR bundles a second, unrelated change** — the `get_board_2d_view` layer default in `python/commands/board/view.py`. That is a different concern from S-expression escaping (it addresses the `pcb export svg` "at least one layer must be specified" failure). It would be cleaner to split it into its own PR so each change stands or falls on its own merits; there is separate in-flight work on that same default, so the two may need sequencing.
2. **The escaping is applied only at the placement path** in `dynamic_symbol_loader.py`, which is the most likely source of the corruption noted in the README. Sibling writers still emit raw values — `add_symbol_property.py:38`, `add_library_symbol_property.py:49`, the sheet name/file in `schematic_hierarchy.py:65`, and the `Reference` fields in `eagle.py`. Those are pre-existing gaps, not introduced here, but the same embedded quote could corrupt a file through one of those tools, so a follow-up to route them through a shared escaper would close the class of bug.

mixelpixx:
Babysitter re-verification after a large main wave landed (#254, #316, #317, #319, #320, #322, #327, #328 all merged since the last check). Still not merging — flagging for the next interactive session.

Re-verified on top of current `main` (`2c40481`), branch still merges cleanly:

- Full suite: 8 failed / 1432 passed / 4 skipped. All 8 failures are the known environmental baseline (Java not found: `test_autoroute_best_of_n` + `test_freerouting`). No regression from this branch.
- `black --check` clean on both changed files.
- `mypy python/`: `Success: no issues found in 72 source files`.

The escaping fix itself is unchanged and still correct, so nothing has decayed there.

One update on the earlier note about the bundled `board/view.py` change, now that the other work is pinned down: it is not merely adjacent to #277, it is the **same edit on the same line**. Both PRs replace `params.get("layers", [])` with `params.get("layers") or [...]` at `view.py:170`. The only difference is the default list — this PR uses `F.Cu, B.Cu, F.SilkS, Edge.Cuts`, #277 uses `F.Cu, B.Cu, F.SilkS, B.SilkS, Edge.Cuts`. Confirmed by merging the two branches against each other: the sole conflict is that hunk.

That matters because the `--layers` half of #277 was already reviewed and approved, and it carries nine layer-flag regression tests that this hunk does not. Whichever lands second will conflict, and the loser of that race gets resolved by hand with no test coverage to catch a mistake.

The cleanest path is to drop the `python/commands/board/view.py` hunk from this PR entirely and let it be purely the S-expression escaping fix — which is a distinct, well-tested concern that can then merge in any order. #277 will bring the layers default with its tests. If you would rather keep `B.SilkS` out of the default, that is a comment worth making on #277 instead, where the tests live.

The second earlier note still stands as a follow-up, not a blocker: sibling writers (`add_symbol_property.py:38`, `add_library_symbol_property.py:49`, the sheet name/file in `schematic_hierarchy.py:65`, `Reference` fields in `eagle.py`) still emit raw values and can corrupt a file through the same embedded-quote path. Routing them through a shared escaper would close the class of bug.

mixelpixx (changes requested):
Thanks for this — the corruption is real, I reproduced it, and it is a **third** mechanism distinct from the two the README claims fixed (incomplete instance blocks, and minified single-line writes). Good find, especially for a first contribution.

Your escaper is correct as written and I want to be explicit about *why*, because it matters more than it looks: `\` before `"` is the only ordering that fixes the real case here, and three other escapers in this repo (`footprint.py:639`, `symbol_creator.py:79`, `eagle.py:74`) escape quotes only and would not. Yours matches the one correct precedent, `_escape_sexpr_string` at `schematic_handlers.py:389`.

Two changes before merge, plus one finding you should know about.

**1. The mechanism is not what the description says — and that points at a second bug.**

The PR says the inner quote of `power:GND`'s Description closes the token early. It can't, because the value never reaches `_property()` with a quote in it. `_extract_lib_property_value` at `dynamic_symbol_loader.py:906` matches with `"([^"]*)"`, which is not escape-aware, so on the source text `"...with name \"GND\""` it stops at the first `\"` and returns:

```
Power symbol creates a global label with name \      <- trailing lone backslash
```

Emitted raw, that trailing backslash escapes the *closing* quote, the token runs on, and the rest of the file is swallowed. I confirmed this end to end — `sexpdata.loads()` fails on the emitted fragment and passes once your escaper is applied.

So the outcome is the same and your fix genuinely stops the corruption. But it means that after this PR the file parses while the Description is silently **truncated** to a half sentence ending in a backslash, permanently mismatched against the same property in `lib_symbols`. The complete repair is escape-aware extraction — `"((?:[^"\]|\.)*)"` plus an unescape — at `:906` and at the sibling positional regex at `:676`, which has the same `"[^"]*"` and can therefore mis-associate `(at ...)` offsets to the wrong property. Happy for that to be a follow-up rather than growing this PR; I have logged it either way. Please just drop the "closes the token early" wording from the description so the next reader isn't sent down the wrong path.

**2. Please drop the `python/commands/board/view.py` hunk.**

It has nothing to do with quoting, and it is not merely adjacent to PR #277 — it is the *same edit on the same line*. #277 defaults to the same four layers plus `B.SilkS` and carries 9 regression tests for it. Whichever lands second will conflict, and #277's version is better covered. Keeping this PR to the escaping change also makes it obviously safe.

**3. Tests are required.**

The invariant worth locking in is not "the escaper escapes" but "a placed instance is parseable". Suggested:

- Unit: `_esc_sexpr('a\')` → `'a\\'`; `_esc_sexpr('say "hi"')` → `'say \\"hi\\"'`.
- Integration (the one that matters): build a temp `.kicad_sch` whose `lib_symbols` contains a symbol with `(property "Description" "Power symbol creates a global label with name \"GND\"")`, call `create_component_instance`, then assert the written file parses via `sexpdata.loads`. That fails without your change and passes with it — and it is the case a quote-only escaper would miss, so it also pins the ordering.

`tests/test_schematic_component_properties.py:104` is a good shape to copy.

Worth noting separately: `reference` is still emitted raw at `dynamic_symbol_loader.py:1057`, and it is caller-supplied, so `reference='R"1'` still corrupts through a path this PR leaves open. Not yours to fix here — I am tracking it with the extraction bug.

One heads-up on process: CI now actually runs the Python suite (it was a no-op until today, #334), so your PR will get real black/mypy/pytest results once you push. That is new as of a few hours ago and is not something you did wrong.

mixelpixx:

```bash
git fetch origin
```

mixelpixx:
@PaulHubiss — an update, and a correction to expectations rather than a request for more work.

**Your escaping fix is still needed. It has not been superseded.**

I landed #348 today for the same family of bug (#336), and I want to be precise about what it did and did not cover, because I initially assumed it overlapped yours more than it does.

**#348 fixed the READ side** — eight places that pulled property values out with `"([^"]*)"`, which stops at an escaped quote. It also fixed three writers that escaped quotes without escaping backslashes first.

**It did not fix the write site your PR targets.** `dynamic_symbol_loader.py` still emits instance properties raw:

```python
f'    (property "{name}" "{value}"\n'
```

And #348 arguably makes that more visible rather than less. Before, the reader truncated `power:GND`'s Description to a lone backslash; now it correctly returns the real value — **with a live quote in it** — which this writer then emits unescaped:

```
UNESCAPED (main today)
  (property "Description" "Power symbol creates a global label with name "GND"")
  parses into 5 elements:
     [0] Symbol('property')
     [1] 'Description'
     [2] 'Power symbol creates a global label with name '
     [3] Symbol('GND')          <-- the value has split into three tokens
     [4] ''

ESCAPED (your fix)
  (property "Description" "Power symbol creates a global label with name \"GND\"")
  parses into 3 elements:
     [2] 'Power symbol creates a global label with name "GND"'   <-- correct
```

So the property block is structurally malformed today. Your `_esc_sexpr` — backslash first, then quote — is exactly right, and it is the same ordering the shared helper uses.

**What changed for your PR:**

1. **Drop the `board/view.py` hunk.** That `--layers` default landed separately in #337 (cherry-picked from @stefangordon's #277, which had the same edit plus a regression test). Keeping it is what puts you in conflict.
2. **Use the shared helper instead of a local `_esc_sexpr`.** #348 added `escape_sexpr_string` to `utils.sexpr_format`, so `from utils.sexpr_format import escape_sexpr_string` replaces your function body. Same semantics, one implementation.

That leaves your PR as essentially a two-line change at the write site, which should also resolve the conflict.

The earlier review asked for tests. Now that the helper and its test module exist, the useful addition is narrower: one integration test that writes an instance from a library symbol whose Description contains an escaped quote and asserts the emitted block parses to 3 elements rather than 5. Happy to point at `tests/test_sexpr_escaping.py` as the pattern.

If you would rather not pick this back up, say so and I will land the write-site fix myself with attribution to you — you diagnosed it first and the diagnosis was right. No pressure either way.

mixelpixx:
@PaulHubiss — I have implemented this in #354 rather than leave it sitting, because it is live corruption on `main` right now. **The diagnosis and the fix are yours; the credit is on the commit and in the PR.**

To be straight about why I did not just wait: #348 landed earlier today and made this **worse**, not better. It fixed the reading side, so the reader now correctly returns `power:GND`'s Description **with a live quote in it** — which the write site you identified then emitted verbatim. Round-tripping a real stock symbol produced **11 s-expression atoms where there should be 3**. That was my gap to close, not yours to wait on.

Your `_esc_sexpr` had the ordering right — backslash first, then quote. That is precisely the part three other writers in this repo had backwards, and getting it right is the difference between a fix and a no-op.

What changed from your PR:

- uses the shared `escape_sexpr_string` from `utils.sexpr_format` rather than a local copy, so there is one implementation
- extends to the sibling raw writes #336 named: `(reference ...)`, `(lib_id ...)`, `(project ...)`, and the `Sheet name` / `Sheet file` properties
- drops the `board/view.py` hunk — that `--layers` default landed in #337
- fixes a consequence your PR would also have hit: `schematic_hierarchy`'s sheet lookup compares a plain name against file text that now holds the escaped form, so it had to match both

Closing this in favour of #354. Thank you for the report and the diagnosis — it was correct, and the delay in landing it was on our side, not yours. Review on #354 is welcome if you are interested.

PR #326, opened 2026-07-16
feat(jlcpcb): real-time part lookup via Open Platform API

ravishivt:
## Summary

`get_jlcpcb_part` now performs a **real-time** lookup against the JLCPCB **Open Platform**
(`getComponentDetailByCode`) — live stock and tiered pricing — when API credentials are
configured, and transparently falls back to the local snapshot database otherwise. Each
result reports which backend answered via a new `source` field (`"live-api"` vs `"local-db"`).

## Why

The local parts snapshot (populated by `download_jlcpcb_database`) mirrors a prebuilt catalog
that can lag JLCPCB by days-to-weeks, so a specific part's stock/price is often stale (observed:
a part reading 22.3M in-stock in the snapshot vs **0** live). The JLCPCB Open Platform exposes a
real `getComponentDetailByCode` endpoint for authoritative per-part data, but the repo never
used it — `get_part_by_lcsc` was a dead stub returning `None`, and the server never loaded the
`.env` where users put their credentials.

Note the Open Platform has **no keyword/parametric search endpoint** (only detail-by-code and a
thin full-catalog dump), so search intentionally stays local; only single-part lookup goes live.
The legacy `jlcpcb.com/external` bulk path (`fetch_parts_page`/`download_full_database`) is
deprecated and kept only for backwards compatibility.

## Changes

- **`python/commands/jlcpcb.py`** — `get_component_detail(codes)` (batched ≤1000, C-prefix
  normalization, tolerates both the bare-list and documented `componentDetailResponseVOList`
  shapes); a real `get_part_by_lcsc`; `normalize_detail_to_part` mapping the detail response to
  the `JLCPCBPartsManager` part shape — **`price_breaks` kept as `[{qty, price}]`** to match the
  local shape and the TS renderer, with the full tiers under `price_ranges`, and
  `libraryType` base/expand → Basic/Extended; `has_credentials`; `_post_signed` (reuses the
  existing JOP/HMAC-SHA256 auth, sends the signed body byte-identically as raw `data`); and
  `_load_env_file` (auto-loads a project-root `.env`, non-destructively — never overrides an
  already-set env var).
- **`python/kicad_interface.py`** — `_handle_get_jlcpcb_part` does live-first-then-local and
  backfills `manufacturer` (absent from the live detail) from the local snapshot row; adds
  `source` to the response.
- **`src/tools/jlcpcb-api.ts`** — `get_jlcpcb_part` description now documents the real-time
  behavior, surfaces `source`, and renders `manufacturer` gracefully when empty.
- **`tests/test_jlcpcb_live_api.py`** — new, fully mocked (no network): auth header
  construction, batching/normalization/dropped-codes, the fixed price shape and library-type
  mapping, `.env` non-override, `has_credentials`, and the handler's live-success /
  live-failure→local / no-creds→local paths.
- **`docs/JLCPCB_INTEGRATION.md`, `docs/JLCPCB_USAGE_GUIDE.md`, `.env.example`** — document the
  Open Platform endpoint, `.env` loading, the `source` field, and that search is local-only
  (replacing the stale JLCSearch offset-pagination narrative).

## Test plan

- `python -m pytest tests/test_jlcpcb_live_api.py -q -o addopts=""` → **28 passed**
- `python -m pytest tests/test_jlcpcb_downloader.py -q -o addopts=""` → **13 passed** (unchanged)
- `npx tsc --noEmit` → clean
- Manual (real creds in a gitignored `.env`): `get_jlcpcb_part {"lcsc_number":"C25804"}` returns
  live stock + tiered pricing with `source: "live-api"`; unplugging creds falls back to
  `source: "local-db"`.

Python 3.9-compatible; params are camelCase; type hints + docstrings. `.env` is gitignored and
never committed (a placeholder `.env.example` is provided instead).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,12 @@
+# JLCPCB Open Platform API credentials (for real-time get_jlcpcb_part lookup).
+#
+# Copy this file to `.env` in the project root and fill in your own values.
+# `.env` is gitignored — never commit real credentials.
+#
+# Obtain these from your JLCPCB account: Account -> API Management -> create key.
+# Without them, get_jlcpcb_part transparently uses the local snapshot database
+# (populated via download_jlcpcb_database); search is always local.
+
+JLCPCB_APP_ID=your_app_id
+JLCPCB_API_KEY=your_access_key
+JLCPCB_API_SECRET=your_secret_key
--- a/docs/JLCPCB_INTEGRATION.md
+++ b/docs/JLCPCB_INTEGRATION.md
@@ -4,16 +4,64 @@
 
 The KiCAD MCP Server integrates with JLCPCB's parts library to provide intelligent component selection, cost optimization, and automated part sourcing for PCB assembly.
 
-**Current Implementation**: Uses the **JLCSearch public API** (by tscircuit) for free, unauthenticated access to JLCPCB's ~100k parts catalog.
+**Current implementation — two backends, by design:**
+
+- **Search is local.** `search_jlcpcb_parts` queries a local SQLite snapshot that
+  you populate once with `download_jlcpcb_database` (the default `cdfer` source is a
+  ~600k in-stock subset). The JLCPCB Open Platform API has **no keyword/parametric
+  search endpoint**, so search must stay local.
+- **Single-part lookup is real-time.** `get_jlcpcb_part` prefers the **JLCPCB Open
+  Platform** `getComponentDetailByCode` endpoint (`https://open.jlcpcb.com`) for live
+  stock and tiered pricing when credentials are configured, and transparently falls
+  back to the local snapshot otherwise. The response's `source` field reports which
+  backend answered (`"live-api"` vs `"local-db"`).
+
+Why the split matters: the local snapshot mirror can lag JLCPCB by weeks, so its
+stock/price for a specific part may be stale — real-time lookup exists to close that
+gap for the part you actually care about.
 
 ## Features
 
-✅ **Parametric Search** - Find components by specifications (resistance, capacitance, package, etc.)
+✅ **Parametric Search** - Find components by specifications (resistance, capacitance, package, etc.) against the local snapshot
+✅ **Real-time Lookup** - Live stock + tiered pricing for a specific LCSC part via the Open Platform API (with local fallback)
 ✅ **Price Comparison** - Compare Basic vs Extended library pricing
 ✅ **Alternative Suggestions** - Find cheaper or higher-stock alternatives
 ✅ **Footprint Mapping** - Automatic JLCPCB package to KiCad footprint mapping
-✅ **Stock Availability** - Real-time stock levels from JLCPCB
-✅ **No Authentication Required** - Public API, no API keys needed
+
+## Real-time part lookup (Open Platform API)
+
+`get_jlcpcb_part` performs a **real-time** lookup when JLCPCB Open Platform
+credentials are configured, and falls back to the local snapshot otherwise.
+
+- **Endpoint:** `POST https://open.jlcpcb.com/overseas/openapi/component/getComponentDetailByCode`
+  with JSON body `{"componentCodes": ["C25804", ...]}` (batched up to 1000 codes per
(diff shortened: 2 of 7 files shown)


mixelpixx:
Babysitter verification pass (automated pre-review; a maintainer will do the design review separately — this is not an approval).

Verified locally on top of current `main` (`2c40481`), branch merges cleanly:

- Full suite: 8 failed / 1460 passed / 4 skipped. The 8 failures are the known environmental baseline (Java not found: `test_autoroute_best_of_n` + `test_freerouting`), unrelated to this PR. Your 28 new tests all pass.
- `mypy python/`: `Success: no issues found in 72 source files`.
- `requests` is already declared in `requirements.txt:26`, so the new dependency is covered.
- `.env` is already gitignored (`.gitignore:60`), so the credential flow in `.env.example` does not risk committing secrets.
- The tests genuinely do not touch the network — `requests.post` is patched in every path and `_load_env_file` is stubbed out, so a developer without JLCPCB credentials still gets a deterministic run. That is the right shape for this and it is appreciated; API-backed features usually arrive with tests that silently skip or hit the live endpoint.

One item to fix before this can go in:

`black --check` fails on `tests/test_jlcpcb_live_api.py`:

```
would reformat tests\test_jlcpcb_live_api.py
1 file would be reformatted, 2 files would be left unchanged.
```

The mechanism is a line-length mismatch, not a style opinion: this project configures black with `line-length = 100` in `pyproject.toml`, and the file looks like it was formatted against black default of 88. So black wants to *join* lines you split, for example:

```python
-    calls = _install_post_capture(
-        monkeypatch, lambda codes: [{"componentCode": c} for c in codes]
-    )
+    calls = _install_post_capture(monkeypatch, lambda codes: [{"componentCode": c} for c in codes])
```

and conversely wants to split the 110-char dict literal in `test_normalize_price_breaks_skip_null_prices`. Running `python -m black tests/test_jlcpcb_live_api.py` from the repo root (so it picks up `pyproject.toml`) resolves all of it — the other two changed files are already clean. Please avoid passing `--line-length` explicitly; letting it read the config is what keeps this consistent.

Everything else is green, so that formatting pass is the only gate outstanding from the automated side.

mixelpixx:
Merged. Clean work — this went in with no changes requested, which is rare for a PR this size.

I specifically verified the two things that worried me about a credentialed API path, and both were already handled properly:

1. **No secrets risk.** `.env.example` holds placeholders only, and `.env` is already gitignored (`.gitignore:60`).
2. **Tests do not touch the network**, despite `test_jlcpcb_live_api.py`'s name. `requests.post` is monkeypatched in every network-exercising test, and the `autouse` `_no_real_env` fixture both neuters `_load_env_file` and `delenv`s all three `JLCPCB_*` vars — so an ambient credential on a developer's machine cannot leak into a test run or silently turn it into a live call. That fixture is the detail I would have asked for if it weren't there.
3. **The offline path is preserved** as the unconditional fallback, gated on `has_credentials()` and also catching API exceptions, with the new `source: "live-api" | "local-db"` field making which one ran observable. 53 JLCPCB tests pass.

Full gates: 1477 passed / 4 skipped (+27 from this PR), `black --check python/` and `mypy python/` clean, `npm run build` and `npm run test:ts` green, and the schema↔route parity guard passes.

Two non-blocking notes:

- `_load_env_file()` now runs on every `JLCPCBClient()` construction, which happens unconditionally at server start (`kicad_interface.py:436`). So the server reads a repo-root `.env` into `os.environ` on every boot. It is non-destructive (`override=False`) and correct for the feature, but it is a new implicit global side effect — worth being deliberate about if it ever grows.
- Main now carries **four** JLCPCB code paths: `JLCPCBClient` (authenticated), `JLCSearchClient` (public/unauthenticated), `JLCPCBPartsManager` (local sqlite), and now the live-API detail lookup. Not a problem today and not yours to fix, but consolidation is worth considering before a fifth appears. If you have a view on which of those should be the long-term primary, I would value it — you know this surface better than anyone here.

Also FYI: CI now genuinely runs the Python suite as of #334 (it was a no-op — `pytest python/` collected 7 tests instead of the 1477 in `tests/`, and every gate ended in `|| echo`). Your future PRs will get real results.

PR #327, opened 2026-07-16
fix(jlcpcb): sanitize FTS query so hyphenated MPN search works

ravishivt:
## Summary

Searching JLCPCB parts by a real, hyphenated manufacturer part number (e.g.
`SHT41-AD1F-R2`) silently returned zero results. `search_parts()` now sanitizes its
FTS5 MATCH query so hyphens (and other FTS special characters) are treated as literal
text, while prefix matching is preserved.

## Why

`JLCPCBPartsManager.search_parts()` built its FTS5 `MATCH` expression by splitting the
query on whitespace and appending `*` to each term. For a hyphenated token like
`SHT41-AD1F-R2`, FTS5 parses `-` as an operator and reads the following bareword as a
column, raising:

```
sqlite3.OperationalError: no such column: AD1F
```

`search_parts()` catches that exception and returns `[]`, so the failure is invisible —
searching by an exact MPN (a very common thing to do) just finds nothing.

## Changes

- **`python/commands/jlcpcb_parts.py`** — new `_build_fts_match_query()` helper emits each
  term as a **quoted prefix phrase** (`"term"*`), making special characters literal;
  embedded double quotes are escaped by doubling, and a caller-supplied trailing `*` is
  normalized so the term stays a prefix. `search_parts()` uses it (and skips the FTS
  filter entirely if the sanitized query is empty).
- **`tests/test_jlcpcb_fts_search.py`** — regression tests on a tiny in-schema SQLite DB
  (no network, no 465MB snapshot): `SHT41-AD1F-R2` now returns `C7461862` (Sensirion),
  the pre-fix raw query still raises, prefix matching still works, multi-term/manufacturer
  filters, and the helper's output shapes/escaping.

## Test plan

- `python -m pytest tests/test_jlcpcb_fts_search.py -q -o addopts=""` → **12 passed**

Python 3.9-compatible; type hints + docstrings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,19 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **JLCPCB part search finds hyphenated MPNs** (#327): `search_parts` built
+  its FTS5 `MATCH` query by appending `*` to each whitespace term, so a real
+  manufacturer part number like `SHT41-AD1F-R2` became `SHT41-AD1F-R2*` — and
+  FTS5 reads `-` as a column/NOT operator, raising
+  `sqlite3.OperationalError: no such column: AD1F`. `search_parts` wraps the
+  query in a broad `except` that returns `[]`, so searching by an exact MPN
+  silently found nothing instead of erroring. Each term is now emitted as a
+  quoted prefix phrase (`"term"*`) with any embedded double quote doubled, so
+  hyphens and other FTS punctuation are matched as literal text; plain prefix
+  matching (`SHT41` still matches the full MPN) is unchanged. A regression test
+  builds a tiny in-schema FTS database and pins both the pre-fix crash and the
+  fixed lookup.
+
 - **`add_schematic_component` snaps the placement origin to the 1.27 mm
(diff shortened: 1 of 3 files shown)


mixelpixx (approved):

- Full suite: **8 failed / 1432 passed / 4 skipped**. The 8 failures are the known environmental baseline (Java not found — `test_autoroute_best_of_n` + `test_freerouting`); the +12 over the prior 1420 baseline are exactly this PR's new tests.
- `black --check` clean on both changed files.
- The mechanism is correct. FTS5 treats `-` as a set/column operator, so the old `SHT41-AD1F-R2*` string was parsed as a NOT/column expression and raised `sqlite3.OperationalError: no such column: AD1F`, which surfaced as an empty result — a real MPN search finding nothing. Quoting each term as a prefix phrase (`"SHT41-AD1F-R2"*`) forces the hyphen to be literal text, and doubling embedded quotes keeps a stray `"` from breaking the MATCH expression. The `test_raw_hyphenated_query_would_crash_fts` regression test pins the pre-fix crash, so the guard can't silently regress.

Ready to merge in the next interactive session. No changes requested.

PR #328, opened 2026-07-16
fix: use Path.open() instead of write_text(newline=) for Python 3.9 support

danliszka:
## Summary

`Path.write_text()`'s `newline` keyword argument was added in **Python 3.10**, but this project supports **Python 3.9+** — and on macOS that isn't optional: KiCad's macOS builds bundle Python 3.9.13, and `pcbnew` is version-locked to that interpreter, so the server *must* run on 3.9 there.

On Python 3.9 every code path below raised:

```
TypeError: write_text() got an unexpected keyword argument 'newline'
```

Affected commands:
- `import_symbol`, `export_symbol`, `rename_symbol` (`commands/library_management.py`)
- Eagle-import canonical-formatting step (`commands/eagle.py`)
- `replace_instance_lib_ids` (`commands/symbol_schematic.py`)

Two test helpers (`tests/test_library_management.py`, `tests/test_symbol_schematic.py`) used the same kwarg, so the pytest suite couldn't run on 3.9 either.

## Fix

Replace the five calls with `Path.open("w", encoding="utf-8", newline="\n")`. This is behaviourally identical — `write_text` simply forwards `newline` to the same underlying `TextIOWrapper` — and works on 3.9. The LF-forcing behaviour (`newline="\n"`) is preserved.

## Verification

- The previously-failing tests in `test_library_management.py` and `test_symbol_schematic.py` now pass on Python 3.9.13.
- `compileall` and `vermin` confirm a 3.9 minimum-version floor for `python/` and `tests/` after this change (this was the only 3.10+ API in the codebase).
- Full pytest suite: no regressions.

## Notes

The docs already state "Python 3.9 or Higher" and that macOS KiCad ships 3.9; this makes the code match that contract. CI covers 3.9 in its matrix, so this path is exercised there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,18 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **`.kicad_sym` and schematic writes work again on Python 3.9** (#328): the
+  library-management (`import_symbol`/`export_symbol`/`rename_symbol`), Eagle
+  prettify, and symbol-schematic writers introduced with the recent tooling
+  wrote files via `Path.write_text(content, newline="\n")`, but
+  `Path.write_text` did not accept the `newline` keyword until Python 3.10 —
+  so on the project's declared `>=3.9` floor every one of those calls raised
+  `TypeError: write_text() got an unexpected keyword argument 'newline'`. Each
+  site now opens the file explicitly (`path.open("w", encoding="utf-8",
+  newline="\n")`) and writes through the handle, preserving the forced LF line
+  ending (so the files stay byte-identical on Windows rather than emitting
+  CRLF) while running on 3.9.
+
 - **JLCPCB part search finds hyphenated MPNs** (#327): `search_parts` built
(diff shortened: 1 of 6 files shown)


mixelpixx (approved):

- Full suite: **8 failed / 1420 passed / 4 skipped**. All 8 failures are the known environmental baseline (Java not found — `test_autoroute_best_of_n` + `test_freerouting`); nothing this PR touches regresses.
- `black --check` clean on all five changed files.
- The mechanism is correct and the fix is a genuine bug on the declared `>=3.9` floor. `Path.write_text()` did not accept the `newline=` keyword until Python 3.10 (it was added there), so every `write_text(..., newline="\n")` call — introduced by the library-management work in the 07-12 session (#316/#317) — raises `TypeError` under 3.9. Rewriting each as `with path.open("w", ..., newline="\n") as f: f.write(...)` preserves the explicit `\n` line ending (important on Windows, where the default would emit CRLF) while working on 3.9.
- Coverage is complete: a tree-wide search for `write_text(...newline...)` returns no remaining occurrences, so this closes the class rather than one instance.

Ready to merge in the next interactive session. No changes requested.

PR #330, opened 2026-07-20
fix: update Eagle import headers for KiCad 10

### LINKED ISSUE #321: Eagle import still writes a KiCad-9 schematic header (20250114), plus stale-format leftovers after #221
#221/#295 removed the stale KiCad-9 header from the project/schematic creation path, but an audit of every remaining place that writes schematic headers turned up a few 20250114-era leftovers. Collecting them in one issue so they are not lost — and so nobody "fixes" the intentional ones.

**1. Eagle import writes a KiCad-9 schematic header (the one real bug here)**

`python/commands/eagle.py:1375` still emits:

```
(kicad_sch (version 20250114) (generator "KiCAD Schematic Editor")
```

This is now the only production code path that writes a KiCad-9 schematic header. The consequence is the same class of problem #221 documented for templates: schematics produced by Eagle import open with the legacy-format banner in KiCad 10 and are exposed to `lib_symbol_mismatch` ERC friction, so an imported project cannot pass a `kicad-cli sch erc --exit-code-violations` CI gate without a manual GUI resave. It should write the same header the blank template now uses: `(version 20260101) (generator "eeschema") (generator_version "10.0")`. Important: 20260101, not 20260306 — KiCad 10.0.x refuses to load files newer than its own writer version, which is why 374b6ea downgraded the template header after #295.

Related, worth checking while in there: `eagle.py:1107` writes `(kicad_symbol_lib (version 20250114) ...)` for the imported symbol library — the same dialect question applies.

**2. Dead constant inviting misuse**

`python/commands/footprint.py:20` defines `KICAD9_FORMAT_VERSION = "20250114"` with zero references anywhere in the codebase. A live-looking format constant with a stale value invites someone to helpfully reuse it. Delete it.

**3. Broken `__main__` demo in the legacy component path**

`python/commands/component_schematic.py:304` (`if __name__ == "__main__":`) calls `SchematicManager.create_schematic()` — which now correctly produces a blank schematic — and then the legacy `ComponentManager.add_component`, which post-#288 raises the intended "no placed `_TEMPLATE_*` donor" error on a blank file. Not a runtime bug (it is demo code), but the demo now demonstrates an exception. Either seed it from `python/templates/template_with_symbols.kicad_sch` or delete the block.

**4. Intentional — do not "fix": stale headers in the fixture templates**

`empty.kicad_sch`, `minimal.kicad_sch`, `template_with_symbols.kicad_sch`, and `template_with_symbols_expanded.kicad_sch` in `python/templates/` still carry `20250114` headers deliberately: they are test fixtures, not production templates. `empty.kicad_sch` alone seeds roughly 17 test files, and the `template_with_symbols*` pair anchors the pin-geometry and mirror-canary tests. Blanket-upgrading their headers would re-baseline a large slice of the suite for zero user-facing benefit. Anyone touching these files should expect and budget for test fallout; documenting that here is the point of this bullet.

giri256:
## Summary

- write Eagle-generated `.kicad_sch` files with KiCad 10's `20260101` / `eeschema` / `10.0` header
- reuse the repository's supported `20241209` symbol-library format token for Eagle-generated `.kicad_sym` files
- remove the unused stale schematic-format constant from the footprint module
- remove the broken standalone legacy `ComponentManager` demo
- leave the intentional `20250114` fixture-template headers unchanged

## Testing

- `.venv\Scripts\python -m pytest tests\test_eagle_import.py -q` � 4 passed, 2 skipped because KiCad CLI is not installed
- `.venv\Scripts\python -m pytest tests\test_component_manager_legacy.py -q` � 3 passed
- repository pre-commit hooks � passed (Black, isort, flake8, mypy, whitespace/EOF/line endings, large-file and merge-conflict checks)

Closes #321


mixelpixx (approved):
Verified locally against main at `1cadf89`. This is green from my side and I am queueing it to merge in the next interactive session. Details of what was checked, since a header bump is exactly the kind of change that passes unit tests and still produces files KiCad refuses to open:

**The header value is right, and right for the correct reason.** The new string is byte-identical to the canonical one this repo already writes in `python/commands/schematic.py:78` and `python/commands/project.py:113`:

```
(kicad_sch (version 20260101) (generator "eeschema") (generator_version "10.0")
```

Worth stating explicitly for future readers: `20260101` is deliberate and must not be "upgraded" to `20260306`. KiCad 10.0.x cannot load `20260306` — that was established in 374b6ea. You picked the value that matches the rest of the codebase rather than the newest-looking one, which is the correct call.

**The kicad-cli-gated tests you had to skip do pass.** Your run skipped 2 tests because KiCad CLI was not installed. I have kicad-cli 10.0 available, so I ran them with it on PATH:

```
tests/test_eagle_import.py ......                                        [ 66%]
tests/test_component_manager_legacy.py ...                               [100%]
9 passed in 3.15s
```

This matters more than the string assertions: `test_minimal_fixture_exports_pdf` feeds the file your changed code path generates to real `kicad-cli sch export pdf`, and `test_import_result_reports_ground_truth_erc` runs real ERC over it. Both succeed, so KiCad 10.0 genuinely parses and renders the `20260101` output — the header bump is not merely cosmetic-consistent, it is accepted by the tool.

**The symbol-library constant reuse is safe.** `KICAD9_SYMBOL_LIB_VERSION = "20241209"` from `symbol_creator.py` is already the shared source for `library_management.py:24` and `symbol_creator.py:165`, so Eagle output now tracks the same token as every other `.kicad_sym` writer instead of a third hardcoded literal. No circular-import risk: `symbol_creator` imports only stdlib, and `library_management.py:18` already establishes this exact import direction. isort is clean on the new import.

**Both deletions are genuinely dead.** `KICAD9_FORMAT_VERSION` in `footprint.py` has no remaining references anywhere in `python/`, `tests/`, or `src/` (only stale `.pyc` matches). The `__main__` demo in `component_schematic.py` did `from schematic import SchematicManager` — a bare top-level import that cannot resolve under the current package layout, so it was already broken rather than merely unused; no test touches it, since every test imports `ComponentManager` directly and `test_component_manager_legacy.py` still passes 3/3.

**Gates.** Full suite `8 failed / 1433 passed / 4 skipped`. The failure set is exactly the 8 known environmental failures on this machine (5 in `test_autoroute_best_of_n.py` + 3 in `test_freerouting.py`, all "Java executable not found"), i.e. no new failures — judge these by the failure set, not the pass count, which drifts with every merge. `black --check` clean on all four changed files, `isort --check-only` clean, `mypy python/` reports "Success: no issues found in 72 source files". Merges into current main with no conflict.

One scoping note, not a change request: `python/templates/*.kicad_sch` still carry `20250114`, which is correct and should stay — those are seed templates whose header is rewritten at write time by `project.py`/`schematic.py`, and `test_blank_template_no_seed.py` / `test_create_project_paths.py` already assert the stale token never reaches written output. Leaving them alone was the right decision.

PR #332, opened 2026-07-23
fix(schematic): scope replace-field restore to the component block; refuse ambiguous label deletes

### LINKED ISSUE #331: Schematic write-safety: replace_schematic_component corrupts lib_symbols; delete_schematic_net_label removes the wrong label on ambiguous names
Two related write-safety bugs in the schematic tools, both hit during a real board-review session (KiCad 10.0.4, Linux, server v2.3.1). A PR with fixes + regression tests follows.

## Bug 1: `replace_schematic_component` corrupts `lib_symbols` while restoring fields

`python/commands/schematic_batch.py::replace_schematic_component` restores the old component's fields after inserting the new symbol like this:

```python
prop_pat = re.compile(r'(\(property\s+"' + re.escape(fname) + r'"\s+)"[^"]*"')
content_after = prop_pat.sub(r'\1"' + fval + '"', content_after, count=1)
```

The substitution runs over the **whole file** with `count=1`, so it rewrites the *first* `(property "<name>" ...)` in the document. For field names that also occur in library definitions — `Description`, `Datasheet` — that first match is inside `lib_symbols`, not the replaced component.

**Repro** (real case): schematic whose `lib_symbols` contains `Connector_Generic:Conn_01x04` (with its `Description` property) serialized *before* the placed symbols; call `replace_schematic_component` on a `Device:D_TVS` diode (which carries `Description: "Bidirectional transient-voltage-suppression diode"`), replacing it with `Device:D_Zener`.

**Result**: the connector's library definition now reads `(property "Description" "Bidirectional transient-voltage-suppression diode")`, and every placed `Conn_01x04` instance fails ERC with *"Symbol 'Conn_01x04' doesn't match copy in library"*. The replaced diode's own `Description` field is left empty.

A secondary effect of the same code: fields that don't exist in the new symbol's property set (custom `MPN`, `LCSC`, …) are silently dropped, because the regex finds nothing to substitute inside the intended block.

## Bug 2: `delete_schematic_net_label` deletes the wrong label on ambiguous names

`WireManager.delete_label` treats `position` as optional and, when it is `None`, deletes the **first** label whose text matches. Net names legitimately appear many times (power rails like `3V3` — 11 labels in my schematic), so a bare-name delete silently removes an arbitrary one.

This is easy to trigger by accident: the tool schema strips unknown top-level keys, so a caller who passes `x`/`y` instead of `position: {x, y}` is *not* rejected — the handler receives `position=None` and deletes the first `3V3` label in the file, which in my session disconnected a supercap pin on the other side of the sheet.

**Expected**: a bare-name delete of a name that occurs more than once should refuse and report the candidate positions, so the caller can disambiguate.

## Environment

- KiCAD MCP Server v2.3.1 (current `main`, 2c40481)
- KiCad 10.0.4, Fedora Linux
- Server driven over stdio JSON-RPC

Euter2:
Fixes #331.

Two write-safety fixes for the schematic tools, each with regression tests. Both bugs were hit in a real board-review session; details and repro in the linked issue.

## 1. `replace_schematic_component`: scope field restore to the component block

The post-replace field restore ran a `count=1` regex substitution over the **whole file**, rewriting the first matching `(property "<name>" ...)` in the document — for names like `Description` that is a library definition inside `lib_symbols`, not the replaced component. In practice a `D_TVS → D_Zener` replace stamped the diode's old `Description` into an unrelated `Conn_01x04` library entry, and every placed instance of that connector started failing ERC with *"doesn't match copy in library"*.

Restoration now delegates to the existing block-scoped `_handle_edit_schematic_component` path, which:

- skips the `lib_symbols` section and edits only the referenced component's block, and
- **creates** fields the new symbol doesn't carry yet, so custom BOM fields (`MPN`, `LCSC`, …) survive a replace instead of being silently dropped.

Regression test: `tests/test_schematic_batch.py::TestReplaceFieldRestore` — replaces a component whose `lib_symbols` contains another symbol with a `Description` property, asserts the library text is untouched and the restore flows through the block-scoped editor with the old value.

## 2. `delete_schematic_net_label`: refuse ambiguous bare-name deletes

`WireManager.delete_label` silently removed the **first** label matching the requested name whenever `position` was omitted. Power-rail names appear many times per sheet, and the mistake is easy to make because unknown top-level `x`/`y` arguments are stripped by the tool schema and arrive as `position=None`.

A bare-name delete of a name that occurs more than once now raises `ValueError` listing every candidate position (the handler already surfaces exception messages to the caller):

```
Label '3V3' appears 11 times (at (105.41, 53.34), (144.78, 49.53), ...);
pass position to select which one to delete
```

Unique names and position-qualified deletes behave exactly as before.

Regression tests in `tests/test_schematic_tools.py`: ambiguous name without position refuses and deletes nothing; ambiguous name **with** position deletes only the matching label; unique name without position still works.

## Checks

- `pytest tests/test_schematic_batch.py tests/test_schematic_tools.py`: 45 passed, 1 failed — the failure is `TestHelpers::test_find_project_root_fallback`, which fails identically on a clean `main` checkout (pre-existing, unrelated).
- `black` and `flake8` (repo config) clean on all four touched files.
- End-to-end verified over stdio JSON-RPC against the original failing schematic: the replace no longer touches `lib_symbols`, and the ambiguous delete refuses with the candidate list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,28 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- **`replace_schematic_component` no longer corrupts `lib_symbols` while
+  restoring fields**: the post-replace field restore ran a `count=1` regex
+  substitution over the whole file, so the first matching
+  `(property "<name>" ...)` in the document was rewritten — for common field
+  names like `Description` that is a library symbol definition inside
+  `lib_symbols`, not the replaced component (observed corrupting an unrelated
+  `Conn_01x04` entry and tripping ERC library-mismatch warnings on every
+  instance). Restoration now delegates to the block-scoped
+  `edit_schematic_component` path, which also creates fields that the new
+  symbol does not carry yet instead of silently dropping them.
+
+- **`delete_schematic_net_label` refuses ambiguous bare-name deletes**:
+  without a `position`, `WireManager.delete_label` silently removed the first
+  label matching the name — dangerous when the same net name appears many
+  times (and easy to trigger, since misspelled top-level `x`/`y` arguments are
+  stripped by schema validation and arrive as `position=None`). A bare-name
+  delete of a name that occurs more than once now fails with the full list of
+  candidate positions so the caller can disambiguate; unique names and
+  position-qualified deletes behave as before.
+
 ### Performance
 
 - **Module-level caches for symbol library discovery, resolution, and
--- a/python/commands/schematic_batch.py
+++ b/python/commands/schematic_batch.py
@@ -385,17 +385,29 @@ class SchematicBatchCommands:
                 )
 
             if fields_to_restore:
-                content_after = sch_file.read_text(encoding="utf-8")
-                new_block, _, _ = _find_placed_symbol_block(content_after, reference)
-                if new_block:
-                    for fname, fval in fields_to_restore.items():
-                        prop_pat = re.compile(
-                            r'(\(property\s+"' + re.escape(fname) + r'"\s+)"[^"]*"'
-                        )
-                        content_after = prop_pat.sub(
-                            r'\1"' + fval.replace('"', '\\"') + '"', content_after, count=1
-                        )
-                    sch_file.write_text(content_after, encoding="utf-8")
+                # Restore the remaining fields through the block-scoped component
+                # editor. A plain regex substitution over the whole file (the
+                # previous implementation) replaced the FIRST matching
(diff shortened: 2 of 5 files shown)


Euter2:
Note for the maintainer: open PR #305 also touches `replace_schematic_component` (in `schematic_batch.py`), but addresses a different defect — transactional backup/rollback when the symbol loader fails mid-replace. This PR fixes the field-restore step corrupting `lib_symbols` (issue #331). The two changes are complementary; whichever lands second will need a trivial rebase of the `fields_to_restore` block.

PR #340, opened 2026-07-27
feat(pcb): import_pcb — wrap kicad-cli's native vendor-PCB importer

rossvonfange:
Part of splitting #314 into à la carte pieces, as offered in the thread. Rebased on v2.5.0; `pre-commit run --all-files` is green (black/isort/prettier/flake8/mypy/eslint) and CI-clean.

## What
`import_pcb` wraps KiCad 10's native `kicad-cli pcb import` to convert a vendor PCB (PADS, Altium, Eagle, CADSTAR, Fabmaster, P-CAD, SolidWorks, or a binary Cadence Allegro `.brd`) into `.kicad_pcb`. `format: "auto"` lets the CLI magic-detect the Allegro binary format (there is no `"allegro"` enum literal). Optional `reportFormat: "json"|"text"` captures the structured import report. **PCB/layout only** — kicad-cli has no schematic importer, so it never produces a schematic.

## Notes
- 2 commits: the tool + its tests (`tests/test_pcb_import.py`, 11 tests covering argv construction, output derivation, and error paths).
- Self-contained; touches only the pcb-import files plus the registration hooks (dispatch dict, schema, registry).

---

**Note on round-tripping imported boards (upstream, not this wrapper):** on KiCad 10.0.4, a board produced by `kicad-cli pcb import` is written in a newer file format (`20260206`) that the same 10.0.4 loader (`kicad-cli pcb drc`/`export`, `pcbnew.LoadBoard`) declines to reload. This reproduces with a **direct** `kicad-cli pcb import` call bypassing this tool entirely, so it's a KiCad importer/loader version gap rather than anything this wrapper does — `import_pcb` faithfully passes through and returns exactly what kicad-cli produces (verified: 22 footprints / 58 pads on a real Eagle board, byte-identical to the direct call). The imported board opens fine in the KiCad 10 GUI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,19 @@ both sides had green suites that never crossed it.
   resolves a different build per Python version (black >=25.12 requires
   > =3.10, so the 3.9 runner silently got 25.1.0 and disagreed on three files).
 
+### New Features
+
+- **`import_pcb`**: wraps KiCad 10's native `kicad-cli pcb import` so any MCP
+  consumer can convert a vendor PCB file (PADS, Altium, Eagle, CADSTAR,
+  Fabmaster, P-CAD, SolidWorks PCB, or a binary Cadence Allegro `.brd`) into
+  a `.kicad_pcb` file, with an optional structured import report
+  (`reportFormat: "json"|"text"`). Verified against a real 10 MB binary
+  Cadence Allegro `.brd` (581 footprints, thousands of net occurrences).
+  Binary Allegro `.brd` files must use `format: "auto"` — the CLI's
+  `--format` enum has no `"allegro"` literal, auto-detection by magic is the
+  only supported path. This tool imports PCB/layout data only; kicad-cli has
+  no Concept HDL / OrCAD schematic importer.
+
 ### Bug Fixes
 
 - **Saving to the board's own path was silently treated as a Save As, and
--- a/README.md
+++ b/README.md
@@ -40,6 +40,22 @@ The [Model Context Protocol](https://modelcontextprotocol.io/) is an open standa
 
 https://github.com/mixelpixx/arduino-ide
 
+## What's New (Unreleased)
+
+### Vendor PCB import
+
+- `import_pcb` wraps KiCad 10's native `kicad-cli pcb import` to convert a
+  vendor PCB file — PADS, Altium, Eagle, CADSTAR, Fabmaster, P-CAD,
+  SolidWorks PCB, or a binary Cadence Allegro `.brd` — into a `.kicad_pcb`
+  file. Use `format: "auto"` for binary Cadence Allegro `.brd` files: the CLI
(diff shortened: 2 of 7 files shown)


mixelpixx:
Merged as d4800a2 — thanks, this was the cleanest unit in the queue. Well-validated inputs, no `shell=True`, and the module docstring documents a genuinely useful gotcha (binary Allegro `.brd` needing `--format auto` because there is no `allegro` literal in the CLI enum). That is the kind of thing that saves the next person an hour.

One thing I added on merge rather than sending back, since it is a single line:

**`import_pcb` was not registered in `src/tools/registry.ts`.** It was registered on the MCP server via `server.tool()`, so it was callable — but only by a client that already knew the name. It did not appear in `search_tools`, and `getRegistryStats()` still reported 143 tools. Since the router exists specifically so AI clients discover tools without loading every schema, a tool missing from the registry is effectively invisible in normal use.

I added it under the `board` category, matching how #341, #342 and #343 each register theirs — so this looks like an oversight in this one PR rather than a difference of opinion. Tool count is now 144 and the README matches.

Worth knowing for future PRs: #338 added a guard test that fails CI when the README count and `getRegistryStats()` disagree. It did not catch this one, because omitting the registration kept both numbers consistently wrong at 143. I will look at whether that guard can also assert every `server.tool()` name appears in the registry — that would have caught it.

Verified before merge: 1588 passed / 5 skipped, 64 TypeScript tests, schema↔route parity, `pre-commit run --all-files` clean. The 8 autoroute/freerouting failures are the known missing-JRE ones — confirmed by stashing and re-running on clean main.

Also: CI had never actually run on this branch. All four of your splits were sitting at `action_required` behind the fork-PR approval gate, so the green you saw was your local run. I have approved all four and they passed on our runners.

mixelpixx:
Correction to my comment above — I overstated the impact and want to set it straight, because it affects how much this actually mattered.

I said `import_pcb` was "invisible" and "callable only by a client that already knew the name." **That was wrong.** `registerAll()` in `src/server.ts` registers every tool unconditionally via `server.tool()`, with no gating or `.disable()` anywhere, so `import_pcb` was always in `tools/list` and always callable. Your PR was not shipping a hidden tool.

What the registry entry actually buys is narrower: the registry backs three discovery tools — `list_tool_categories`, `get_category_tools`, `search_tools` — so without it, `search_tools` would not return `import_pcb`. That is a search-index gap, not a functional one. Still worth fixing, and the addition stands, but it was not the defect I described.

Checking this turned up something more useful, which I have since fixed on main: the README's claim that the router "reduces AI context by 70%" has not been true since 2026-03. `3d9497e` deliberately disabled the gating half — the commit note says it "causes Claude to hallucinate tool schemas via search_tools/execute_tool" — and `963a39c` deleted `execute_tool` outright. So direct registration is the intended design, and the docs were five months stale. Corrected in 16e2f00.

Sorry for the noise on your PR. The merge itself is unaffected.

PR #341, opened 2026-07-27
fix(project): preserve .kicad_pro net_settings + add board-origin tools

rossvonfange:
Part of splitting #314 à la carte (per the thread). Rebased on v2.5.0; `pre-commit run --all-files` green, CI-clean.

Grouped because both changes hinge on the same `.kicad_pro` write-safety helper (`utils/project_settings_guard.py`):

## fix: preserve `net_settings` across open/save
`open_project`/save rewrote `.kicad_pro` from a possibly-stale in-memory model and **silently dropped hand-edited `net_settings`** (custom netclasses/patterns). `preserve_project_settings()` snapshots + restores them around the save sites (project.py, freerouting, schematic export, ipc backend). Reproduced + verified on real pcbnew.

## feat: `set_board_origin` / `get_board_origin`
Aux (drill/place) + grid origins — the file-stored datums `export_drill`'s `drillOrigin:"plot"` and pick-and-place/plot exports consume. Real-pcbnew + drill-offset integration tested.

Note: the net_settings guard was originally also applied to `hierarchical_place.py`'s save site; that one hunk is deferred to the hierarchical_place PR (that tool isn't on main yet), so this PR is self-contained on current `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The [Model Context Protocol](https://modelcontextprotocol.io/) is an open standa
 
 **Key Capabilities:**
 
-- 143 tools across 13 categories with JSON Schema validation
+- 145 tools across 13 categories with JSON Schema validation
 - Smart tool discovery with router pattern (reduces AI context by 70%)
 - 8 dynamic resources exposing project state
 - Complete schematic workflow with 27 tools and dynamic symbol loading (~10,000 symbols)
--- a/docs/KNOWN_ISSUES.md
+++ b/docs/KNOWN_ISSUES.md
@@ -87,6 +87,24 @@ AttributeError: 'BOARD' object has no attribute 'LT_USER'
 
 ---
 
+### 6. `.kicad_pro` net_settings Must Be Edited via JSON Merge
+
+**Status:** BY DESIGN (guard added)
+
+Backend board saves are wrapped in `preserve_project_settings()`
+(`python/utils/project_settings_guard.py`): pcbnew serializes a possibly
+stale in-memory project model over `.kicad_pro` on every
+`SaveBoard`/`BOARD.Save`, so the guard restores the on-disk
+`net_settings` (and any dropped top-level keys) after each save.
+
+**Implication:** commands must persist net class / netclass_patterns
+changes via direct JSON read-modify-write of the `.kicad_pro` (the
+`persist_netclass_to_project` pattern in `python/commands/routing.py`),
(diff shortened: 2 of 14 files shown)


mixelpixx:
Merged as 2673813 — thanks, and specifically thanks for reworking this rather than defending it. Both objections from the #314 review were addressed at the mechanism level:

- `open_project`'s read path now restores **verbatim bytes** instead of round-tripping through JSON, so #220's byte-faithful `.kicad_pro` survives a restore.
- The save guard writes **only when something was actually lost** (`merge_preserved_keys`' `changed` flag) rather than on every save.

That is a better design than what I asked for. I also re-verified the guard's core premise against current main: every `net_settings` write path (`routing.py:78`, `routing.py:157`) is a direct JSON read-modify-write happening outside the guard's window, so nothing persists net classes *through* pcbnew's model and relies on `Save`. Restoring wholesale is safe.

Two things resolved on merge:

**1. The `sort_keys=True` nit.** Dropped it from `_write_json_atomic`. It only bites on the rare path where the guard actually writes, but when it does it alphabetises every top-level key — undoing #220 on a file the user never asked to have rewritten. `json.load` preserves insertion order and `merge_preserved_keys` builds from `after`, so KiCad's ordering now survives with restored keys appended. Confirmed against KiCad's real 11-key top-level order, and added two regression tests, since this is exactly the kind of property that regresses silently.

**2. README count conflict**, which was expected — your branch said 145, main said 141, and all four of your splits edit the same line. I did not arbitrate between the numbers; I recounted from `getRegistryStats()` after merging and took that: **143**.

Worth explaining why the baseline moved under you: while merging #342 I found `getRegistryStats()` was double-counting. It computed `routed + direct`, but seven schematic essentials are deliberately in *both* lists, so the headline was overstated by seven. Your 145 was arithmetically correct against the number main was publishing at the time — the number itself was wrong. Nothing to fix on your side.

Your branch also predated 16e2f00, so I took main's version of the adjacent line: the README's "reduces AI context by 70%" claim was removed after I found the router's gating half was rolled back in `3d9497e` back in March.

Verified before merge: 1633 passed / 12 skipped, 70 TypeScript tests, schema↔route parity, pre-commit clean.

That is three of your four splits merged (#340, #342, #341). #343 is next — it is the big one and I want to give the in-place `.kicad_sch` rewriters a proper look against real KiCad before it lands.

PR #342, opened 2026-07-27
Hierarchical schematic tools: remove_hierarchical_sheet, set_sheet_property, hierarchical_place

rossvonfange:
Part of the à la carte split of #314 (per maintainer request for smaller, independently-mergeable PRs).

**Adds three hierarchical-design tools** (all new files — no changes to existing command logic):
- `remove_hierarchical_sheet` — reverse of add_hierarchical_sheet; removes the (sheet …) block + its (sheet_instances) path entry (does not delete the sub-sheet file).
- `set_sheet_property` — edit a hierarchical sheet's name/file/other properties in place.
- `hierarchical_place` — PCB placement that mirrors the schematic sheet hierarchy (vendored HierPlace, MIT, credited in-file).

New files: `python/commands/schematic_hierarchy.py`, `python/commands/_hierplace.py` (vendored), `python/commands/hierarchical_place.py`, `src/tools/schematic-hierarchy.ts`, plus dispatch/registry/schema registration and tests.

Green on the enforced gate: `pre-commit run --all-files` (black/isort/prettier/flake8/mypy/eslint) passes; `npm run build` + `npm test` (64) pass; `pytest` for the hierarchy tools passes (22 passed, 2 skipped). README tool-count line synced to the registry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The [Model Context Protocol](https://modelcontextprotocol.io/) is an open standa
 
 **Key Capabilities:**
 
-- 143 tools across 13 categories with JSON Schema validation
+- 144 tools across 13 categories with JSON Schema validation
 - Smart tool discovery with router pattern (reduces AI context by 70%)
 - 8 dynamic resources exposing project state
 - Complete schematic workflow with 27 tools and dynamic symbol loading (~10,000 symbols)
--- a/docs/TOOL_INVENTORY.md
+++ b/docs/TOOL_INVENTORY.md
@@ -172,6 +172,8 @@ _Source: `src/tools/schematic.ts`_
 | ---------------------------------- | -------------------------------------------------- | ------------------ |
 | `add_schematic_hierarchical_label` | Add hierarchical label for multi-sheet connections | Routed (schematic) |
 | `add_sheet_pin`                    | Add a sheet pin to a hierarchical sheet symbol     | Routed (schematic) |
+| `set_sheet_property`               | Add/update a custom property on a hierarchical sheet (metadata, e.g. generator cell identity) | Routed (schematic) |
+| `get_sheet_properties`             | List hierarchical sheets with their full property maps | Routed (schematic) |
 
 ### Net Analysis (5)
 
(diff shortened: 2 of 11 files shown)


mixelpixx:
Merged as cd6b41b — thanks. The vendored HierPlace is self-contained, the sheet-property tools preserve file formatting rather than round-tripping through a parser, and 361 lines of tests is a good ratio for this much surface.

Two things resolved on merge rather than sent back:

**1. Three of the four tools were not in `src/tools/registry.ts`.** `remove_hierarchical_sheet`, `set_sheet_property` and `get_sheet_properties` were registered on the server but not indexed, so `search_tools` could not find them. `hierarchical_place` was registered correctly. Added the three to the `schematic_hierarchy` category and widened its description to match.

To be clear about what this did and did not affect: the tools were always in `tools/list` and always callable — the registry only backs keyword search. I said something stronger on #340 earlier today and have posted a correction there.

**2. Your PR surfaced a counting bug.** Reconciling the tool count after merging revealed `getRegistryStats().total_tools` was computing `routed.length + direct.length` — but seven schematic essentials are deliberately in **both** lists (always visible *and* searchable). So the headline count was overstated by exactly seven, and the README inherited it. The true registry count is 141; the README variously said 143 and 144. Fixed via a new `getDistinctToolNames()`, and `getRoutedToolNames()`'s docstring claimed it excluded direct tools when it never did.

Nothing you did caused this — it predates the PR by a long way. Worth knowing because the README count guard now reports a *smaller* number than before, and that is the correct one.

**New guard:** `tests-ts/registry-completeness.test.ts` now compares what is wired into the server against what the registry claims. Measuring found **75** pre-existing tools missing from the catalogue, so it lands as a ratchet — the known set is frozen, any new omission fails CI. Tracked in #345. Practically: if a future PR adds a `server.tool()` without a registry entry, CI will tell you, so this stops being something you have to remember.

Verified before merge: 1603 passed / 7 skipped, 70 TypeScript tests, pre-commit clean. The 8 autoroute failures are the known missing-JRE ones.

PR #343, opened 2026-07-27
Schematic authoring & lint tooling: diagnosed load errors, flat-symbol repair, correct SVG page, batch/label fixes, offgrid+cosmetic lint

rossvonfange:
Part of the à la carte split of #314 (per maintainer request for smaller, independently-mergeable PRs).

These commits form **one coherent unit** because they share `SchematicLoadError` infrastructure and all touch the same schematic export/handler paths — splitting them further would create PRs that don't build on their own.

**Included:**
- **Diagnosed load errors** — kicad-skip crashes on flat SnapEDA/SamacSys `.kicad_sym` symbols (no `_1_1` sub-unit), taking down every skip-based tool for the whole sheet. Load failures now raise `SchematicLoadError`; tools return a structured error naming the offending symbols instead of empty/partial results.
- **`repair_flat_symbols`** — wraps flat vendor symbols in a `(symbol "NAME_1_1" …)` sub-unit so skip-based tools work again.
- **Correct SVG page selection** — `export_schematic_svg` and schematic views exported *all* pages and picked one arbitrarily; now deterministically selects the root page (`_pick_root_svg`), exporting into a private temp dir so stale/hierarchical pages can't be picked up. (Fixes a latent `tmp_dir_p` NameError in the crop path surfaced by the merge.)
- **`delete_schematic_component`** — optional cleanup of labels attached to the deleted component.
- **`batch_connect`** — support global labels (`labelType`).
- **`lint_offgrid`** — detect and safely snap off-grid schematic geometry.
- **`lint_schematic_cosmetic`** — hide pin names, orient net labels.
- **`docs/HEADLESS_AUTHORING.md`** — guide for GUI-less schematic authoring.

Green on the enforced gate: `pre-commit run --all-files` (black/isort/prettier/flake8/mypy/eslint) passes; `npm run build` + `npm test` (64) pass; schematic pytest sweep passes (578 passed). README tool-count synced to the registry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,7 @@ connections = ConnectionManager.get_net_connections(sch, "VCC", sch_path)
 - Net connectivity: 100% accurate (VCC: 2 connections, GND: 4 connections)
 - Netlist generation: Working with accurate pin-level connections
 
-See [Schematic Tools Reference](docs/SCHEMATIC_TOOLS_REFERENCE.md) for the complete schematic tool documentation.
+See [Schematic Tools Reference](docs/SCHEMATIC_TOOLS_REFERENCE.md) for the complete schematic tool documentation, and the [Headless Authoring Guide](docs/HEADLESS_AUTHORING.md) for field-tested practice driving these tools without the KiCad GUI.
 
 ### IPC Backend (Experimental)
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The [Model Context Protocol](https://modelcontextprotocol.io/) is an open standa
 
 **Key Capabilities:**
 
-- 143 tools across 13 categories with JSON Schema validation
+- 144 tools across 13 categories with JSON Schema validation
 - Smart tool discovery with router pattern (reduces AI context by 70%)
 - 8 dynamic resources exposing project state
 - Complete schematic workflow with 27 tools and dynamic symbol loading (~10,000 symbols)
--- a/docs/HEADLESS_AUTHORING.md
+++ b/docs/HEADLESS_AUTHORING.md
@@ -1,0 +1,223 @@
+# Headless Schematic Authoring Guide
+
+Field-tested practice for driving the KiCAD MCP Server without the KiCad
+GUI: the per-sheet build recipe, diagnostics, ERC triage, and verification
+discipline. Distilled from building a complete multi-sheet automotive
+carrier design entirely through the server (observed on KiCad 9/10, Linux).
+
+See [SCHEMATIC_TOOLS_REFERENCE.md](SCHEMATIC_TOOLS_REFERENCE.md) for
+per-tool parameters and [ARCHITECTURE.md](ARCHITECTURE.md) for how the
+server is put together.
+
+---
+
+## 1. Two parsers, one truth
+
+Two different parsers are in play, and they disagree — this is the source
+of most headless pitfalls:
+
+| Parser         | Used by                                                                                        | Character                                  |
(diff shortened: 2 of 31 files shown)


mixelpixx:
Merged as 5ecb41a. This was the big one and I gave it a proper look — thanks for the split, reviewing it à la carte was much easier.

## The breaking change is accepted on the merits

`SchematicLoadError` flipping `find_orphaned_wires`, hierarchical net traversal and `sync_schematic_to_board` from graceful-degrade to hard-fail was my main reservation in the #314 review. I've changed my mind, and your framing is why: silently skipping an unparseable sheet produces an **incomplete pad→net map and a wrong-looking success**, which is worse than an error. You owned the change rather than softening it — documented in KNOWN_ISSUES and SCHEMATIC_TOOLS_REFERENCE naming the affected tools, paired with `repair_flat_symbols` so the error is actionable, and given a structured shape instead of a bare exception. That is the right way to ship a behaviour change.

## Verified against real KiCad, not just the unit tests

Both new tools rewrite the user's `.kicad_sch` in place, so I exercised them against a real KiCad 10.0 install:

**`lint_offgrid`** — built a schematic and knocked three coordinates off the 1.27 mm grid:

```
DRY RUN offenders=3 fixed=0 needsHuman=0
   (100.35, 50.8) -> (100.33, 50.8) off 0.02mm [wire_endpoint] line 10
   (120.66, 50.8) -> (120.65, 50.8) off 0.01mm [junction]      line 16
   (105.42, 49.53) -> (105.41, 49.53) off 0.01mm [label]       line 19
FIX     fixed=3
```

All three found and correctly classified. Dry run left the file byte-identical. After `fix=true`: real `kicad-cli sch erc` still parses it, `sexpdata` round-trips it, **line count is unchanged** — confirming the byte-exact splice claim rather than a reformat — and a second pass is a no-op.

**`repair_flat_symbols`** — `dryRun` genuinely defaults to true, and the default call left the file byte-identical. After `dryRun=false`: the `_1_1` sub-unit exists, both pins survive, real KiCad parses it, second pass is a no-op.

I also confirmed your premise rather than assuming it: real `kicad-cli` reports **0 violations** on a SnapEDA-style flat symbol. So this is genuinely a kicad-skip parser limitation, not an invalid file — which matters, because it means `repair_flat_symbols` is working around a tooling gap rather than fixing corruption.

The post-fix `sexpdata.loads()` sanity check before writing (`schematic_lint_grid.py:201`) is a good instinct. Cheap, and it means a bug in the splice logic can't reach the user's file.

## Resolved on merge

1. **`lint_offgrid` and `repair_flat_symbols` were missing from `src/tools/registry.ts`** (`lint_schematic_cosmetic` was correct). Caught automatically — the ratchet test I added while merging #342 failed the build with exactly these two names. So the guard earned its keep on its first real outing. `lint_offgrid` went to `schematic_layout`; `repair_flat_symbols` to `library` provisionally, with a comment pointing at #345, since there is no symbol-library category to hold it properly yet.

2. **README count** — recounted from `getRegistryStats()` rather than arbitrating: **146**. Note the baseline shifted under all four of your splits because I found `getRegistryStats()` was double-counting seven tools that sit in both the category and direct lists.

3. **KNOWN_ISSUES.md** — you and #341 both added a section numbered 6. Union-resolved, yours renumbered to 7.

Verified: 1713 passed / 12 skipped, 70 TypeScript tests, schema↔route parity, pre-commit clean.

That is all four splits merged. #314 gets closed as superseded next — I'll credit the decomposition there rather than closing it silently, since breaking it up was real work and made all four reviewable.

PR #353, opened 2026-07-31
fix(jlcpcb): decode yaqwsx source-db-v2 prices and Basic flag

### LINKED ISSUE #352: yaqwsx source: every part converts with price_json '[]' and basic_parts 0 — upstream changed the published cache.sqlite3 schema
`download_jlcpcb_database` with `source='yaqwsx'` now produces a database where **every** part has `price_json == '[]'`, and `basic_parts` is **0**. The download and extraction succeed, part count and description/category/stock look right — only prices and the Basic flag are wrong, so the failure is silent.

The default `cdfer` source is **not** affected.

## Cause: upstream re-shaped the published catalog

yaqwsx/jlcparts commit [`bb2bf6d7`](https://github.com/yaqwsx/jlcparts/commit/bb2bf6d7) *"Introduce compact source cache"* (2026-05-31) replaced the layout of the `cache.sqlite3` published at `yaqwsx.github.io/jlcparts/data`. The CI step that does it is `jlcparts migratecache cache.sqlite3` in `update_components.yaml`. The file now reports `meta.format = 'source-db-v2'`:

| | before | now |
|---|---|---|
| tables | `components` + `categories` + `manufacturers` | `jlc_components` + `lcsc_components` |
| `price` | JSON array `[{"qFrom","qTo","price"}]` | CSV `qFrom-qTo:unitPrice,...` |
| Basic flag | `basic` column (0/1) | `library_type` holding `'base'`/`'expand'` |
| manufacturer | `manufacturers` join | `jlc_components.manufacturer`, blank falls back to `lcsc_components` |

Confirmed against the live archive (13 volumes, ~625 MB → 5.3 GB `cache.sqlite3`, `Last-Modified` today):

```
meta: format=source-db-v2
jlc_components   7,157,071 rows
lcsc_components  4,099,872 rows
price sample: '1-199:0.0189,200-599:0.0163,600-3999:0.0152,...,20000-:0.0138'
library_type: base 351 | expand 7,156,720
```

## Where it breaks

1. `_ensure_components_view` (`python/commands/jlcpcb_downloader.py:131`) requires `components`+`categories`+`manufacturers`, none of which exist now — it returns without building a view.
2. `_pick_source_relation` (`:150`) therefore falls through to the "largest table" branch (`:163`) and picks `jlc_components` — which happens to work, masking the problem.
3. `normalize_price_json` (`:65`, called at `:262`) only understands a JSON array or a scalar. The CSV does not start with `[`, so it reaches `float(raw)` (`:93`), raises `ValueError`, and returns `'[]'` — **for every row**.
4. Same mismatch on the flag: there is no `basic` column any more, so `_library_type` sees only `preferred` and classifies every part as Extended/Preferred. `basic_parts` is always 0.

## Reproduction

```bash
python download_jlcpcb.py --source yaqwsx
```

Running the real converter over a 5,000-row slice of the live catalog on `main`:

```
STATS: {'total': 5000, 'basic': 0, 'extended': 4936}
empty price_json: 5000/5000
```

## Status

Fixed — PR to follow. `v_components` is rebuilt from `jlc_components LEFT JOIN lcsc_components`, mirroring upstream's own read query (`sourceDb.iterCategoryComponents`), and `normalize_price_json` learns the CSV encoding. Verified on the full catalog: 7,157,071 parts, 146 empty `price_json` — matching exactly the 146 source rows with `price = ''`; `basic` resolves to 351, matching the source's `library_type = 'base'` count.

stefanobaldo:
Fixes #352.

The yaqwsx source stopped producing prices: every converted part landed with `price_json == '[]'`, and `basic_parts` was always 0. Upstream re-shaped the published `cache.sqlite3` (`meta.format = 'source-db-v2'`) — full analysis in the issue.

## Changes

**`_ensure_components_view`** — builds `v_components` from `jlc_components LEFT JOIN lcsc_components` instead of the old `components`/`categories`/`manufacturers` join. Mirrors upstream's own read query (`sourceDb.iterCategoryComponents`): `basic` derived from `library_type = 'base'`, manufacturer falling back to `lcsc_components` when the JLC field is blank, and the `present = 1` filter that hides delisted parts.

**`normalize_price_json`** — new `_parse_price_csv` branch for the `qFrom-qTo:unitPrice,...` encoding. Placed after the JSON-array branch and before the scalar one, so no input that previously worked changes path. Malformed segments are skipped rather than discarding the row.

The old-layout branch is **removed** rather than kept alongside: upstream migrated, and the archive being served has only the new format.

## What is deliberately untouched

- **The target schema.** No change to the `components` table, its indexes or the FTS index, so `JLCPCBPartsManager` consumes the result unchanged. Verified on a freshly built database: same 13 columns in the same order, same 5 indexes, and `get_part_info("C1002")` returns populated `price_breaks`.
- **The CDFER path.** CDFER ships its own `v_components` view and JSON prices — verified by reading its `sqlite_master` directly — so it returns at the first guard and still needs the JSON branch. Removing that branch would have broken the default source.

## Verification

Full conversion of the real 5.3 GB catalog:

```
7,157,071 parts in 5.1 min
empty price_json: 146 / 7,157,071
basic: 351
```

Both residuals are exact: 146 matches the source rows with `price = ''`, and 351 matches the source's `library_type = 'base'` count. Nothing is lost by the converter.

Tests: 2 added (`source-db-v2` conversion end-to-end, and the CSV decoder incl. open-ended breaks, out-of-order breaks and malformed input). The obsolete legacy-layout test and its fixture are removed. 14/14 in the file; full suite 1776 passing — the 3 failures in `test_kicad_3rd_party_env_resolution.py` are pre-existing on `main`.

## Note for reviewers

The full yaqwsx catalog is now **7.1 M parts / ~5.1 GB**, against 616 k / 465 MB from CDFER — upstream growth, not this change. Only 685 k have stock > 0. If that size is unwanted, a `stock > 0` filter in the view would cut it, but that changes what `--source yaqwsx` has always delivered, so it is not part of this fix.

### DIFF
--- a/python/commands/jlcpcb_downloader.py
+++ b/python/commands/jlcpcb_downloader.py
@@ -62,16 +62,41 @@ def _progress(progress: ProgressFn, message: str) -> None:
 # ---------------------------------------------------------------------------
 
 
+def _parse_price_csv(raw: str) -> List[Dict[str, Any]]:
+    """Decode yaqwsx ``source-db-v2`` price CSV into ``[{"qty","price"}]``.
+
+    Format is ``qFrom-qTo:unitPrice`` segments joined by commas, where an empty
+    ``qTo`` means the break is open-ended, e.g.::
+
+        1-199:0.0189,200-599:0.0163,20000-:0.0138
+
+    Malformed segments are skipped rather than discarding the whole row.
+    """
+    out: List[Dict[str, Any]] = []
+    for segment in raw.split(","):
+        range_text, sep, price_text = segment.strip().partition(":")
+        if not sep:
+            continue
+        q_from = range_text.partition("-")[0]
+        try:
+            out.append({"qty": int(q_from), "price": float(price_text)})
+        except ValueError:
+            continue
+    out.sort(key=lambda brk: brk["qty"])
+    return out
+
+
 def normalize_price_json(raw: Any) -> str:
     """Normalize a source price value into the manager's ``[{"qty","price"}]`` shape.
 
-    Handles both CDFER/jlcparts JSON arrays (``[{"qFrom","qTo","price"}, ...]``)
-    and a bare scalar price. Returns a JSON string (``"[]"`` when no price).
+    Handles CDFER/legacy-jlcparts JSON arrays (``[{"qFrom","qTo","price"}, ...]``),
+    the yaqwsx ``source-db-v2`` price CSV (``qFrom-qTo:unitPrice,...``), and a bare
+    scalar price. Returns a JSON string (``"[]"`` when no price).
     """
     if raw is None or raw == "":
         return json.dumps([])
 
-    # JSON array string (CDFER / jlcparts store price as TEXT JSON)
+    # JSON array string (CDFER / legacy jlcparts store price as TEXT JSON)
     if isinstance(raw, str) and raw.strip().startswith("["):
         try:
             arr = json.loads(raw)
@@ -87,6 +112,10 @@ def normalize_price_json(raw: Any) -> str:
             qty = item.get("qFrom", item.get("qty", 1)) or 1
             out.append({"qty": qty, "price": price})
         return json.dumps(out)
+
+    # yaqwsx source-db-v2 CSV ("1-199:0.0189,200-:0.0138")
+    if isinstance(raw, str) and ":" in raw:
+        return json.dumps(_parse_price_csv(raw))
 
     # Scalar numeric price
     try:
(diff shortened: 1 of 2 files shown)


mixelpixx:
Merged — thank you for pairing the report (#352) with the fix and a faithful 25-column fixture of the new schema. Verified locally on top of current main before merging (64 tests green including the new decoder edge cases).

One non-blocking note for a possible follow-up: this removes the legacy components/categories/manufacturers branch entirely, and _ensure_components_view now early-returns unless jlc_components exists — so a user with a pre-migration cache.sqlite3 on disk silently gets a no-op view. Since upstream migrated the published archive that is probably academic, but a logger.warning naming the old layout ("cache predates source-db-v2; delete and re-download") would turn a silent empty result into a self-explaining one.

PR #355, opened 2026-08-01
feat(routing): add route_net — obstacle-aware A* net routing

fthrvi:
## The gap

`route_trace` and `route_pad_to_pad` draw a trace exactly where they are told. That leaves finding a legal path outside the server — which, for an LLM driving a board, is the hard part. It can read the netlist and the DRC report, but it cannot see the free space.

The only obstacle-aware option today is the Freerouting wrapper, and that is all-or-nothing: you hand it the whole board and take what comes back. There is nothing in between for "make this one connection, legally, and tell me if you can't."

(For what it's worth, Konnect has the same shape — `route_pad_to_pad` is L-bend geometry and `autoroute` wraps Freerouting — so this may be worth porting if the Rust rebuild wants it.)

## What this adds

`route_net` builds a per-layer occupancy model of the board — every other net's pads, tracks, zones and rule areas, grown by clearance — and runs A* over it, for one net, on demand. Pads are routed to whatever copper the net already has, so it repairs a partially routed net as well as building one.

```jsonc
{
  "net": "ESTOP_SENSE",
  "layers": ["F.Cu", "In2.Cu", "B.Cu"],
  "layerCosts": { "F.Cu": 3, "In2.Cu": 1, "B.Cu": 3 },
  "clearance": 0.2,
  "width": 0.26,
  "dryRun": true
}
```

`pads: ["U2.3"]` restricts it to one connection, which is the common repair case.

## Details that are wrong in the obvious implementation

Each of these is pinned by a test. They are all cheap to get wrong and expensive to notice on real copper:

- **Vias get their own occupancy map.** A via is fatter than a trace; checking a layer change against the trace map is how one lands a tenth of a millimetre from another net. A via also punches the whole stack, so it must clear *every* layer, not the two it nominally joins.
- **A diagonal step may not squeeze between two blocked cells.** On the grid it looks clear; in copper it clips the corner.
- **A pad terminal is every free cell inside the pad, not its centre.** On a 0.65 mm pitch package the neighbours' clearance covers the centre cell, so a centre-only terminal never enters the search — even though the pad is 1.5 mm long and its far end is wide open.
- **Per-layer costs**, so heavy outer copper can be made expensive and an empty inner signal layer cheap. Without it the router carves through power planes to save a via.
- **`relaxedArea` mirrors a `.kicad_dru` rule area** so the router agrees with DRC about local clearance. Without it, fine-pitch escapes that DRC would accept are refused.

## No new dependencies

Grids are `bytearray`; Pillow (already required) rasterises the polygons. numpy is deliberately not introduced — a 100×80 mm board at 0.1 mm is ~800k cells per layer, which indexes fast enough in CPython. Occupancy for a 4-layer board of that size builds in about 0.2 s.

The pathfinding core is `python/commands/net_router.py` and imports no pcbnew, so the new tests run under the existing conftest stub without KiCad installed.

## Validation

Exercised through `RoutingCommands` on a real 4-layer board (100×80 mm, 177 footprints, 49 A resonant power stage):

- routed **5 of 6 open pads** on a net Freerouting had left unrouted
- correctly reported `no legal path` for a pad whose only escape corridor is 0.3 mm tall — too narrow for a via. That is a placement problem, and the tool says so rather than drawing something illegal.

## Checks

- `npm run build` — clean
- `pytest tests/test_net_router.py` — 17 passed
- full `pytest tests/` — no new failures (14 pre-existing failures on `v2.6.0` are unchanged; verified by stashing this branch and re-running)
- `black`, `isort`, `flake8`, `mypy` clean on the new and edited files
- `eslint`, `prettier` clean

Registered in the `routing` category and dispatched as `route_net`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,47 @@
 
 All notable changes to the KiCAD MCP Server project are documented here.
 
+## [Unreleased]
+
+### Added
+
+- **`route_net` — obstacle-aware net routing** (new). `route_trace` and
+  `route_pad_to_pad` draw a trace exactly where they are told, so the caller
+  has to know a legal path before asking. For an LLM driving a board that is
+  the hard part: it can read the netlist and the DRC report, but it cannot see
+  the free space. `route_net` builds a per-layer occupancy model of the board —
+  every other net's pads, tracks, zones and rule areas, grown by clearance —
+  and runs an A\* search over it.
+
+  Use it when a connection has to be made but no legal path is known, when
+  `route_trace` produced a short or a clearance violation, or when the
+  autorouter left a net unrouted. Pads are routed to whatever copper the net
+  already has, so it repairs a partially routed net as well as building one.
+
+  The search encodes a few things that are wrong in the obvious implementation
+  and expensive to notice on real copper:
+  - **Vias get their own occupancy map.** A via is fatter than a trace;
+    checking a layer change against the trace map is how one ends up a tenth of
+    a millimetre from another net. A via also punches the whole stack, so it
+    must clear every layer, not the two it nominally joins.
+  - **Diagonal steps may not squeeze between two blocked cells.** On the grid
+    it looks clear; in copper it clips the corner.
+  - **A pad terminal is every free cell inside the pad, not its centre.** On a
+    0.65 mm pitch package the neighbours' clearance covers the centre cell, so
+    a centre-only terminal never enters the search even though the pad is
+    1.5 mm long and its far end is wide open.
+  - **Per-layer costs.** Make 2 oz outer copper dear and an empty inner signal
+    layer cheap and the router stops carving through power planes to save a
+    via.
+  - **`relaxedArea` mirrors a `.kicad_dru` rule area** so the router agrees
+    with DRC about local clearance. Without it, fine-pitch escapes that DRC
+    would accept are refused.
+
+  No new dependencies: the grids are `bytearray`, and Pillow (already required)
+  rasterises the polygons. The pathfinding core lives in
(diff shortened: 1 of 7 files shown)


mixelpixx:
Fit decision: declining, and it is a scope call, not a quality call — the geometry here is better than most contributed routers (the via occupancy grown by clearance plus via radius with every layer required clear, diagonal corner-cutting blocked, pad terminals as free-cells rather than centers, and no pcbnew import so the core tests under the stub conftest — all deliberate, all correct).

The reason is ownership: merging a homegrown router makes this project responsible for DRC-adjacent geometry forever. Every KiCad clearance-rule nuance — netclass-specific clearances, keepouts, rule areas, teardrops — becomes a bug report against us, and 'the router made an illegal board' is the most expensive class of bug a tool like this can ship. Freerouting remains the routing story precisely because that ownership stays upstream. (For transparency: a technical review also found an admissibility issue — _octile assumes unit step cost while astar multiplies by layer_costs, so any layer cost below 1.0 makes the heuristic inadmissible and paths silently non-optimal; clamp or scale by min cost if you take this work elsewhere.)

The part of this I would genuinely like to have: the occupancy model as a *diagnostics* tool. 'Why can't this net route — where is it blocked, at what clearance, on which layers' is a question Freerouting answers with silence, and your grid answers it directly with no legality obligations attached. A check_net_routability tool built on your occupancy core, reporting the blocking geometry, would be uncontroversially in scope and I would review it quickly.

Thank you for the work and for framing the gap honestly in the description — the single-net use case is real, which is why the diagnostic form of this is wanted even though the mutating form is not.

PR #357, opened 2026-08-03
feat(schematic): symbol-bbox, dry-run collision check, connectivity check, and auto-router tools

daniSoares08:
## Summary
- `get_schematic_symbol_bbox`: returns absolute-coordinate bounding boxes (body + pins) for requested symbol references.
- `add_schematic_wire` gains a `dryRun` mode: reports symbol crossings and foreign-net collisions before writing anything.
- `check_schematic_connectivity`: fast same-net check between two points (pin or x/y), without a full ERC run.
- `connect_schematic_pins`: new auto-router (`python/commands/schematic_router.py`) that A*-routes orthogonal wires between 2+ pins on a coordinate-compressed grid, avoiding symbol bodies and foreign-net wires, with own-net conflict detection and multi-pin sequential nearest-insertion.
- Also includes an unrelated, already-staged fix: native footprint placement with save/reload fallback for KiCad 10, plus command-timeout refactor.

Motivation: while manually converting a real schematic sheet from net-labels to point-to-point wires, the biggest bottlenecks were no visibility into symbol geometry, no way to detect a wire silently shorting two nets before writing it, and no fast same-net check. These tools close those gaps and add a first-pass auto-router.

## Test plan
- [x] `npm run build` (TS compiles, registry-completeness test passes)
- [x] `npm run lint:ts` (0 errors)
- [x] Python: `pytest` full suite — 1829 passed, 32 skipped
- [x] `black --check` on touched files
- [x] Live MCP smoke test against a real, densely-wired schematic (found and fixed an A* performance bug via this)


mixelpixx:
Reviewed — and this needs a split before it can be reviewed properly, because the title describes maybe half the diff.

The buried half is the problem. python/kicad_api/ipc_backend.py alone is ~740 lines of this diff — a wholesale kipy-0.7 rewrite — and its base blob is pre-#369, so it reimplements get_open_board_path differently from the version that merged with tests on Aug 9. Merging as-is would either conflict loudly or, resolved carelessly, silently replace a tested implementation with an untested near-duplicate. It also collides head-on with #378, which touches the same functions with a different approach; only one PR can own the kipy-0.7 story, and that decision should not be forced through a PR titled 'schematic tools'. The '1829 passed' claim also cannot be current — it was run against a base that is now several weeks stale.

The good news: the title half is genuinely useful, and its registry hygiene is correct (all three tools registered, first README count bumped).

Concrete ask — split into three:
1. **get_schematic_symbol_bbox + add_schematic_wire dryRun + check_schematic_connectivity** — small, well-scoped, wanted. Rebase on current main, fix the second README count occurrence (~line 489, which readme-counts.test.ts does not catch but readers do), and this part merges quickly.
2. **connect_schematic_pins / schematic_router.py** — a 549-line A* router needs its own scope discussion (the project just declined a PCB-side A* router on maintenance-ownership grounds in #355; schematic wire routing is a different question, but it deserves to be asked, not buried).
3. **The IPC/kipy-0.7 work** — rebase on post-#369 main and reconcile with #378 (see the review there: the fp-lib-table-based discovery is the direction). Between the two PRs, one factored implementation should win.

The freerouting.py and command-timeout.ts changes are unexplained in the body — justify or drop them in whichever split they belong to.

mixelpixx:
Closing for now. The split asked for on 2026-08-20 (the symbol-bbox / dry-run collision check / connectivity check trio separated from the kipy 0.7 rewrite of ipc_backend.py) did not get a reply, and the branch is now 88 commits behind main with conflicts in src/server.ts, python/kicad_api/ipc_backend.py and README.md, so it cannot be merged as it stands.

The three schematic tools are still wanted: a fresh PR with just them, rebased on main, gets a prompt review. The ipc_backend half should not come back in that PR; #369 already fixed get_open_board_path with tests, and the kipy compatibility work has its own thread in #378.

Issue #336, opened 2026-07-26
S-expression property values: extraction is not escape-aware, and several write sites still emit raw


mixelpixx:
Split out of the review of #324, which fixes the fatal half of this. Filing the remainder so it is not lost.

## 1. Extraction silently truncates values containing escaped quotes

`_extract_lib_property_value` (`python/commands/dynamic_symbol_loader.py:906`) matches with:

```python
re.search(r'\(property\s+"' + re.escape(prop_name) + r'"\s+"([^"]*)"', sym_block)
```

`[^"]*` is not escape-aware. On a source value containing `\"` — e.g. `power:GND`'s Description, `Power symbol creates a global label with name \"GND\"` — it stops at the first escaped quote and returns a value ending in a **lone backslash**:

```
Power symbol creates a global label with name \
```

Before #324 that trailing backslash escaped the closing quote on the way out and swallowed the rest of the `.kicad_sch` (verified: `sexpdata.loads()` fails on the emitted fragment). #324's escaper stops the file corruption, but the value is still **wrong** — a half sentence, permanently mismatched against the same property in `lib_symbols`.

Fix: match `"((?:[^"\]|\.)*)"` and unescape the result, so the round-trip is faithful rather than merely non-fatal.

The sibling positional regex at `dynamic_symbol_loader.py:676` has the same `"[^"]*"` for the value and can therefore fail to match a property whose value contains an escaped quote — mis-associating `(at ...)` offsets to the wrong property.

## 2. Write sites that still emit raw

`(reference "...")` at `dynamic_symbol_loader.py:1057` is **caller-supplied**, so `reference='R"1'` still corrupts the schematic through a path #324 does not cover. That one is worth doing first.

Others found by grepping for f-strings building `(property "`:

- `python/commands/schematic_hierarchy.py:65` and `:68` — Sheet name / Sheet file, both user-supplied
- `python/commands/add_symbol_property.py:38`, `add_library_symbol_property.py:49`
- `python/commands/eagle.py:853`, `:989`, `:1131`, `:1135` — Reference/Value emitted raw while neighbouring fields call `_escape`

## 3. Three escapers in the tree are wrong

These escape `"` but not `\`, which is a no-op for exactly the case above (the value arrives with a trailing backslash and no quote left to escape):

- `python/commands/footprint.py:639` `_esc`
- `python/commands/symbol_creator.py:79` `_esc`
- `python/commands/eagle.py:74` `_escape`

[...]

PR #358, opened 2026-08-03
Fix schematic corruption on quoted property values, and nets dropped during board sync

davidwesternall-oss:
I ran a trivial two-part board (LED + current-limiting resistor) through the whole
create → add component → sync → place → route → export Gerbers flow, and hit three
separate bugs on the schematic side. Fixes and tests for all three here.

## 1. Placing any stock power symbol writes a schematic KiCad can't open

`create_component_instance` copies `Datasheet` and `Description` from the library symbol
into the placed instance. It unescapes the value on the way in
(`_extract_lib_property_value` → `unescape_sexpr_string`) but interpolated it straight
back into a quoted token on the way out, without re-escaping.

Both `power:VCC` and `power:GND` ship with a quote inside their Description:

```
Power symbol creates a global label with name "VCC"
```

so the emitted instance came out as:

```
(property "Description" "Power symbol creates a global label with name "
    VCC ""
```

The token terminates at that inner quote and the rest of the form gets swallowed.
Eeschema then won't load the file at all, and `run_erc` reports a bare
"Failed to load schematic" with nothing pointing at the cause. Since most schematics
have a GND symbol on them somewhere, this is easy to walk into.

This looks like a straggler from the #336 work in #348 — that pass made the property
*readers* escape-aware, but this writer was missed. `escape_sexpr_string` already
existed, it just wasn't being called here.

## 2. `sync_schematic_to_board` silently drops nets that have no label

Two things going on in `_build_hierarchical_pad_net_map`:

**`PWR_FLAG` was being treated as a net name.** A `#FLG` symbol's Value is the literal
string `PWR_FLAG` — it's a marker for ERC, not a net. That put a bogus `PWR_FLAG` net on
the board, and since the name gets stamped at the flag's pin position it can also
propagate out over the real net through the BFS below it. A PWR_FLAG takes whatever net
it's wired to; it never names one.

**Unlabelled nets vanished entirely.** A net made only of a wire between two component
pins has no label and no power symbol, so nothing named it, and `nearby_net()` returned
None for both pads. Those pads reach the board with no net assigned at all — the
connection disappears from the layout and the pads can't be routed. Nothing errors;
`pads_assigned` just quietly comes back short (2 of 4, in my case) and you only notice
when the ratsnest is missing a connection.

KiCad synthesises a name for these during F8, so this now does the same: cluster the
unnamed wire points and name them after the first connected pin. On the LED case that
produces `Net-(D1-A)`, which is exactly what `kicad-cli`'s own netlist calls it, so a
later real F8 in the GUI won't churn the name.

## 3. `angle` and `mirrorY` were being thrown away

`create_component_instance` has always accepted both, and the TS tool layer has always
sent both, but `_handle_add_schematic_component` in between only read `unit`. So the
documented `angle` argument did nothing and every symbol got placed unrotated. Two lines.

## Tests

22 new tests across the three areas. 10 of them fail if you revert the matching fix — I
checked that individually for each rather than assuming, since a regression test that
passes either way isn't worth much. The rest are guards against over-correcting: an
explicit net label still wins over an auto-generated name, and a single-pin wire stub is
deliberately left unnamed.

Full suite passes otherwise. For transparency, there are 14 pre-existing failures on my
machine (Windows) in the autoroute/freerouting tests plus a few that shell out to
subprocesses, but they fail identically on a clean checkout of `main`, so they're
unrelated to these changes.

### DIFF
--- a/python/commands/dynamic_symbol_loader.py
+++ b/python/commands/dynamic_symbol_loader.py
@@ -14,7 +14,12 @@ import uuid
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
-from utils.sexpr_format import QUOTED_VALUE, QUOTED_VALUE_SKIP, unescape_sexpr_string
+from utils.sexpr_format import (
+    QUOTED_VALUE,
+    QUOTED_VALUE_SKIP,
+    escape_sexpr_string,
+    unescape_sexpr_string,
+)
 
 logger = logging.getLogger("kicad_interface")
 
@@ -1013,10 +1018,19 @@ class DynamicSymbolLoader:
             """Build a KiCad-10 (property ...) block.
 
             Field order: at, [hide yes], show_name no, do_not_autoplace no, effects.
+
+            ``value`` arrives unescaped -- _extract_lib_property_value() runs the
+            library text through unescape_sexpr_string(), and caller-supplied
+            values are plain Python strings. Re-escape it here, or a value
+            containing a quote terminates the token early and corrupts every
+            following s-expression in the file. The stock power:VCC / power:GND
+            Description fields ('...global label with name "VCC"') hit this, so
+            the damage lands on ordinary schematics, not just exotic input.
             """
             hide_line = "      (hide yes)\n" if hide else ""
+            safe_value = escape_sexpr_string("" if value is None else str(value))
             return (
-                f'    (property "{name}" "{value}"\n'
+                f'    (property "{name}" "{safe_value}"\n'
                 f"      (at {_fmt(px)} {_fmt(py)} {_fmt(pa)})\n"
                 f"{hide_line}"
                 f"      (show_name no)\n"
--- a/python/commands/schematic_handlers.py
+++ b/python/commands/schematic_handlers.py
@@ -246,6 +246,11 @@ class SchematicHandlersMixin:
             x = component.get("x", 0)
             y = component.get("y", 0)
             unit = component.get("unit", 1)
+            # The TS tool layer always sends these two (defaulting to 0/False);
+            # not reading them here silently ignored the documented `angle` and
+            # `mirrorY` arguments, so every symbol was placed unrotated.
+            angle = component.get("angle", 0)
+            mirror_y = component.get("mirrorY", False)
 
             # Derive project path from schematic path for project-local library resolution.
             # Walk up from the schematic file to find the directory that owns the project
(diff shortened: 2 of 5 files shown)


mixelpixx:
Reviewed in full. This bundles three independent fixes, and they need different dispositions — the middle one is the reason I want this PR.

**The nets-dropped-during-board-sync half is the most valuable content here and is not implemented anywhere else.** Both bugs check out against main: the #FLG branch stamps the literal 'PWR_FLAG' Value as a net name at the flag's pin, from where the BFS can propagate it over the real rail; and unlabeled pin-to-pin wires get no name, so their pads reach the board unassigned — silently, with pads_assigned just short. Your synthesized Net-(D1-A) names matching kicad-cli's netlist convention (so a later F8 does not churn names) is the right call, and the guard tests (explicit label still wins; single-pin stub not named) are the ones that prevent over-correction.

**The quoted-property half is superseded** — #354 merged today and escapes name and value plus reference/lib_id/project/path at the same site, so this hunk now conflicts. One piece of it is better than what we merged, though: your TestPlacedInstanceEscapesCopiedValues class tests end-to-end through create_component_instance against a real power:VCC fixture and asserts sexpdata parses the output — where #354's tests partially grep source text. I would like to cherry-pick that test class onto main with credit regardless of what happens with this PR — say if you object.

**The angle/mirrorY half duplicates #351** (same file, same lines). #351 is positioned to land first; theirs will own it unless you two prefer otherwise.

Ask: rebase onto post-#354 main and reduce this PR to the net-sync half plus its tests. As reduced, it merges quickly.

mixelpixx:
Thanks again for this. The nets half is on its way in #409 with you as co-author: the PWR_FLAG exclusion and the wire-cluster naming are yours as written. One correction made on the way, after checking with kicad-cli 10.0.0 on synthetic schematics: KiCad names such a net from the pad number even when the pin has a name (`Net-(D1-Pad2)`, not `Net-(D1-A)`), and picks the candidate that sorts lowest as a plain string, so R10 beats R2. The implementation and the tests now say that. The escaping half of this PR landed in #354 in August, and the angle/mirrorY read from your third hunk is in #410. Once #409 merges I will close this as superseded with credit; if you would rather rebase and carry it yourself, say so and I will hold.

mixelpixx:
Superseded by #409, now on main, with you as co-author on the merge commit. Thank you for the analysis and the fix: the PWR_FLAG exclusion and the wire-cluster naming are yours, corrected only where KiCad's own naming turned out to differ (pad number rather than pin name, lowest candidate as a plain string). Closing this one.

PR #362, opened 2026-08-06
fix: add_symbol_property corrupted .kicad_sym libraries

karu2003:
## Summary

`add_symbol_property` corrupted every `.kicad_sym` library it touched. The symbol block was sliced `content[start:end]` — excluding the symbol's own closing paren — and then spliced back with `content[end + 1:]`, which skipped over that same paren. One `)` vanished per call, and the library eventually stopped loading.

Three further defects lived in the same write path:

- **New properties landed inside a unit.** The insertion point was matched with `\n\t\t\t(symbol "` — three tabs. Units in a `.kicad_sym` file sit at *two* tabs (three is the schematic's `lib_symbols` nesting, where the sibling tool `add_library_symbol_property` operates). The match never fired, so every insert took the fallback branch and appended to the end of the block, i.e. inside the last unit.
- **Updating a property truncated it.** The match `\(property "Name".*?\)` is non-greedy, so it stops at the first `)`. In the multi-line layout eeschema writes that is the `(at ...)` line, leaving the old `(hide yes)` and `(effects ...)` lines behind as orphans directly under the symbol.
- **Parens inside quoted values were counted as structure.** A `Description` of `"Ceramic (X7R) 50V"` mis-located the end of the block.

This is the tool most often used to write procurement fields (Manufacturer, MPN, supplier PNs) across a whole library, so the damage accumulates silently over a batch run and only surfaces when eeschema refuses to open the library.

## What changed

- Block boundaries come from a string-aware paren matcher (`_match_paren`), so quoted text is never read as structure.
- Properties are located and inserted at the symbol's own nesting level via `_iter_children`. A unit's `Reference` can no longer be mistaken for the symbol's, and new properties are placed before the first unit — where KiCad expects them.
- The whole property block is replaced on update, so no orphan `(hide yes)` / `(effects ...)` fragments are left behind.
- Values are escaped with the existing `escape_sexpr_string` helper.
- Indentation is copied from the file being edited rather than hardcoded to tabs, so space-indented and Eagle-imported libraries keep their formatting and the diff stays limited to the inserted lines.
- Rewriting a property now inherits its current position and visibility unless `position` / `hide` are given. Without this, replacing the block (the fix for the truncation bug) would have introduced a new one: updating a value would un-hide the field and move it back to the origin. The schema and the zod descriptions document the omit-to-inherit behaviour.

## Test plan

- [x] 17 new tests in `tests/test_add_symbol_property.py`, all against the tab-indented multi-unit layout eeschema actually writes (the previous fixture used a compact 2-space format in which none of these bugs reproduce): paren balance on add and on update, insertion above the first unit, parent-vs-unit property resolution, indentation matching, parens and quotes in values, and a symbol with no units.
- [x] `pytest tests/` — 1756 passed. The 20 failures are pre-existing on `upstream/main` at `0dc3ee8` (`test_pin_locator_y_flip`, `test_wire_connectivity_multi_unit`, `test_case_sensitivity_warning`, and others), unchanged by this branch.
- [x] `npm run build`, `npx vitest run` (70 passed), `eslint`, `prettier`, `black`, `isort`, `flake8`, `mypy` all clean.
- [x] End-to-end against a real 487-symbol library (5 multi-unit) with KiCad 10.0.4: 20 adds plus 5 updates.
  - Before: paren balance `+12`, `kicad-cli sym upgrade` exits 2 with `Unable to load library`.
  - After: paren balance `0`, `kicad-cli sym upgrade` exits 0.

## Note

`add_library_symbol_property` — the sibling tool that writes into a schematic's `lib_symbols` — shares the truncating-update pattern. It is left alone here to keep this change reviewable; I can follow up separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to the KiCAD MCP Server project are documented here.
 ### Bug Fixes
 
 - **`add_symbol_property` corrupted `.kicad_sym` libraries**. Every call dropped
-  one closing paren, so the library stopped loading in eeschema. Four defects,
+  one closing paren, so the library stopped loading in eeschema. Eight defects,
   all in the same write path:
   - The rewritten symbol was spliced back as
     `content[:start] + block + content[end + 1:]`, but `block` was sliced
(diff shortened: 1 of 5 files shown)


mixelpixx:
Merged. Before merging I re-verified the bug class on current main: the dropped closing paren, the non-greedy update regex stopping at '(at 0 0 0)' (which in my reproduction promoted a unit symbol to a top-level sibling — silent structural corruption a paren count cannot see), and the three-tab insertion regex that never fires on real .kicad_sym layouts. All real, all fixed here. The balance guard comparing against the original rather than zero is the right call for Eagle-imported libraries, and _is_direct_child catching wrong-depth insertion is exactly the check the old code needed.

Two follow-ups filed rather than blocking: add_library_symbol_property has the same truncating-update pattern (you flagged this yourself), and once your batch landed there are two copies of the paren-walking helpers (your private ones here, the shared ones in sexpr_format.py from #365) — consolidation issue open.

PR #363, opened 2026-08-06
feat: validate_schematic and validate_symbol_library

karu2003:
## Summary

Two new tools, `validate_schematic` and `validate_symbol_library`, that locate structural damage in `.kicad_sch` and `.kicad_sym` files.

`kicad-cli` already answers whether a file loads. It does not answer *where* it broke: one misplaced paren in a 40 000-line library produces `Unable to load library` and nothing else, which leaves bisecting the file by hand as the only way forward. Since the server has a growing set of tools that edit these files as text, "did my last edit damage the file, and if so where" is a question worth being able to ask directly. `kicad-cli` is invoked internally in 14 places but never exposed as a tool.

## What the scan finds

A single string-aware pass reports unbalanced parens, unterminated strings and trailing content, each with a line and column. Parens inside quoted values (`"Ceramic (X7R) 50V"`) are treated as text, so the naive `count("(") - count(")")` check — which gives false positives on any real library — is not used.

Three checks then cover damage that is syntactically legal but still fatal:

| Check | Why it matters |
|---|---|
| `orphan_fragment` | A `(property ...)`, `(effects ...)` or `(at ...)` directly under `(kicad_sch ...)` is what a truncated property rewrite leaves behind. It is balanced, so nothing else notices, and eeschema refuses to open the file. |
| `unit_name_mismatch` | Renaming a symbol without renaming its `NAME_0_1` sub-symbols makes the whole library unloadable. |
| `duplicate_symbol` | KiCad silently keeps one of them. |

`unit_name_mismatch` is graded an **error** rather than a warning on evidence: the file is perfectly balanced, so only KiCad can say whether it loads, and kicad-cli 10.0 answers `Unable to load library`. There is a test pinning that (`test_cli_agrees_a_stale_unit_name_is_fatal`).

## Two decisions worth calling out

**Semantic checks stop when the structure is broken.** Past a paren fault every subsequent node nests one level too deep, so the name checks turn into hundreds of reports of the same root cause. On a real 487-symbol library, deleting a single `)` produced 489 errors before this guard and 2 after. The response carries `semanticChecksRan: false` so the omission is visible rather than implied.

**An indentation heuristic locates the break.** `unclosed_form` can only ever name the outermost form that stayed open, which for a paren dropped mid-file is always line 1 — true and useless. KiCad writes one tab per nesting level, so the first line whose tab count disagrees with its actual depth is where the structure broke. On the corrupted 487-symbol library that points at line 75, the line right after the deleted paren. The heuristic is skipped for files that are not tab-indented, where it would say nothing.

**kicad-cli runs on a copy.** It is the authoritative answer, so a file that passes the scan but not KiCad is still reported (`kicad_cli_rejected`). The copy is not optional — `upgrade` rewrites in place, and a validator must not modify the file it is validating. There is a test for that too. A hierarchical schematic only loads if its sub-sheets are reachable, so sibling `.kicad_sch` files are copied alongside it, capped at 64 MB. `runKicadCli: false` gives a fast structure-only check, or lets the tool work where KiCad is not installed.

## Test plan

- [x] 30 new tests in `tests/test_validate_kicad_files.py`. All but the three `@needs_cli` ones are hermetic (`runKicadCli: false`), so the suite does not depend on a KiCad install; the cli-backed ones skip when `resolve_kicad_cli()` returns `None`.
- [x] `pytest tests/` — 1769 passed. The 20 failures are pre-existing on `upstream/main` at `0dc3ee8` and unchanged by this branch.
- [x] `npm run build`, `npx vitest run` (70 passed), `eslint`, `prettier`, `black`, `isort`, `flake8`, `mypy` clean. The README headline count guard is updated to 148 tools / 14 categories.
- [x] Against real files with KiCad 10.0.4: a 487-symbol library and three schematics from a hierarchical project all validate clean; a deliberately corrupted copy of the library reports `unclosed_form` plus an `indent_depth_mismatch` at line 75, and the original file is byte-identical afterwards.

Registered in all five places (Python module, `kicad_interface` import and route, `tool_schemas`, TS tool file, and `registry.ts`), so `search_tools` can find them — the new `validation` category rather than an addition to `KNOWN_UNREGISTERED`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,21 +18,36 @@ All notable changes to the KiCAD MCP Server project are documented here.
   `"Ceramic (X7R) 50V"` are text, so the naive `count("(") - count(")")` check
   that gives false positives on real libraries is not used. On top of that:
   - **Orphaned fragments.** A `(property ...)`, `(effects ...)` or `(at ...)`
-    sitting directly under `(kicad_sch ...)` is what a truncated property
-    rewrite leaves behind. It is balanced, so nothing else notices, and
-    eeschema refuses to open the file.
+    sitting directly under `(kicad_sch ...)` — or, in a `.kicad_sym`, an
+    `(effects ...)`, `(at ...)`, `(hide ...)`, `(font ...)` or `(justify ...)`
+    directly inside a `(symbol ...)` — is what a truncated property rewrite
+    leaves behind. It is balanced, so nothing that counts parens notices, and
+    KiCad refuses to open the file — kicad-cli 10.0.4 answers
+    `Unable to load library` for a perfectly balanced library with either
+    fragment misplaced. The `.kicad_sym` rule is deliberately not applied to
+    `.kicad_sch`, where `(at ...)` is a legal child of a placed `(symbol ...)`.
   - **Stale unit names.** Renaming a symbol without renaming its `NAME_0_1`
     sub-symbols makes the _whole library_ unloadable — confirmed against
     kicad-cli 10.0, which is why this is graded an error rather than a warning.
+  - **Escaped units.** A dropped paren promotes a unit to the top level, where
+    its name still says which symbol it came from. Graded a _warning_, because
+    the library does still load (kicad-cli returns 0) — it just loads wrong:
+    KiCad reads the unit as a symbol of its own, so the parent keeps none of
+    those graphics or pins and `symbolCount` comes out too high.
   - **Duplicate symbol names**, where KiCad silently keeps one.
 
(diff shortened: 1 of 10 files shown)


mixelpixx:
Merged, immediately after #362 deliberately: this validator detects precisely the corruption class that fix repairs, so users get both the fix and a way to find already-damaged libraries in the same release. Grounding _ORPHAN_FRAGMENTS against all 222 shipped KiCad libraries is the standard of evidence this repo aims for. Running kicad-cli on a throwaway copy (because upgrade rewrites in place) is a detail most implementations get wrong.

PR #359, opened 2026-08-04
fix(registry): give the symbol-library tools a search_tools home

AmirF194:
## Root cause

`searchTools()`, `getAllCategories()`, and `getRegistryStats()` in `src/tools/registry.ts` only read `toolCategories` and `directToolNames`. Any tool wired up with `server.tool()` in `src/tools/*.ts` but never added to either structure is callable, but invisible to `search_tools`/`get_category_tools` and uncounted by `getRegistryStats()`. `tests-ts/registry-completeness.test.ts` freezes the current gap as a 75-entry `KNOWN_UNREGISTERED` ratchet (added after #340/#342 shipped unregistered tools the same day).

15 of those 75 are the symbol-library tools, spread across `library-symbol.ts`, `symbol-creator.ts`, and `schematic-batch.ts`: browse/search a `sym-lib-table` library, create/edit/rename/import/export a `.kicad_sym` file, and sync a placed schematic instance back to its library definition. The `library` category's own comment already named this exact fix as pending: it parks `repair_flat_symbols` there "for now because there is no symbol-library category yet -- see #345" (added when #343 merged).

## Fix

Adds a `symbol_library` category holding those 15 tools plus `repair_flat_symbols`, and removes the now-stale placeholder comment. Shrinks `KNOWN_UNREGISTERED` from 75 to 60. Updates the two README lines `readme-counts.test.ts` pins to `getRegistryStats()` (146 -> 161 tools, 13 -> 14 categories) to match.

**Scope: this covers only the 15 symbol-library entries, not all 75 named in #345.** The remaining 60 span board/component/schematic/routing/footprint-library categorization, each with its own judgment call on category placement; the ratchet is built to make these reviewable in separate batches, so leaving `Refs #345` open rather than closing it with this partial fix.

## Verification

- `npm run test:ts` (Vitest, clean `node:20-bookworm` container): 70/70 pass, including `registry-completeness.test.ts`'s five completeness checks and `readme-counts.test.ts`'s README/registry parity check.
- `npm run build` (tsc) and `npm run lint:ts` (eslint) both clean; `pre-commit run --files` on the three changed files (prettier, eslint, trailing-whitespace, etc., the repo's own Code Quality CI job) all pass.
- Competing-PR sweep (14/14 open PRs, full pagination): 2 candidates touch `registry.ts` (#355 adds a `routing` category tool, #357 adds three `schematic` category tools), read both patches directly and neither overlaps the `library`/`symbol_library` hunks this PR touches.
[...]

Issue #345, opened 2026-07-28
search_tools misses 75 of the server's 216 tools


mixelpixx:
## What

The server registers **216** distinct tools via `server.tool()`. The registry in `src/tools/registry.ts` indexes **141** of them. The other **75** are absent from every category and from `directToolNames`.

## Impact — real but narrow

To be precise about severity, since it is easy to overstate:

- **Not** a functional break. `registerAll()` (`src/server.ts`) registers all 216 unconditionally, nothing is gated, so every tool is in `tools/list` and directly callable.
- What breaks is **search**. The registry backs `list_tool_categories`, `get_category_tools` and `search_tools`. A tool missing from it cannot be found by keyword — an assistant that asks "what routing tools are there?" gets an incomplete answer and may conclude a capability does not exist.

Several of the 75 are things a user would plausibly search for: `run_erc`, `refill_zones`, `delete_trace`, `route_pad_to_pad`, `route_differential_pair`, `create_netclass`, `import_eagle_project`, `search_jlcpcb_parts`, `get_component_list`.

## Guard is already in place

`tests-ts/registry-completeness.test.ts` (landed in cd6b41b) freezes the 75 as `KNOWN_UNREGISTERED` and fails CI on any **new** omission. It also checks the reverse direction — a registry entry with no `server.tool()` would be advertised then fail on call — currently zero.

So this is not bleeding. The task is working the baseline down.

## Why it accumulated

Nothing enforced it. Adding a tool means editing two files, and the second is easy to forget with no feedback — #340 and #342 both missed it on the same day, which is what prompted the measurement. The pre-existing 75 are the same oversight going back further.

The README count guard could not catch it either: omitting the registration leaves the README and the registry consistently wrong at the same number.

## Work

For each of the 75, decide routed (add to a `toolCategories` entry) or direct (add to `directToolNames`, reserved for always-visible essentials), then remove it from `KNOWN_UNREGISTERED`. Mostly mechanical; the judgement is category choice. Some may deserve new categories — there is no category today for datasheet, JLCPCB, footprint-creation or symbol-creation tools, which is part of why they were skipped.

Good first issue for someone wanting to learn the tool surface. Can be done in batches; the ratchet means partial progress is safe.

## Related

[...]

PR #364, opened 2026-08-06
feat: read, remove and repoint library table entries

karu2003:
## Summary

Three tools that complete the library-table story: `list_library_table`, `remove_library_table_entry` and `set_library_table_uri`.

`register_symbol_library` and `register_footprint_library` can add a row to `sym-lib-table` / `fp-lib-table`. Nothing could read one back, drop one, or repoint one. So the tail end of any library migration — dropping the `eagle_import` row, moving a project onto `${KIPRJMOD}`-relative paths, finding the stale row behind a wall of `footprint_link_issues` — falls back to hand-editing the table.

That is also the part most likely to go wrong when scripted. The obvious approach, `content.replace(")", ...)` to splice before the closing paren, corrupts a table whose last entry ends on the same line as the table's own `)` — which is exactly how KiCad writes the global table.

## How these avoid that

Entries are parsed as `(lib ...)` spans with a string-aware paren matcher, edits slice exactly that span, and **the result is re-parsed before anything is written**. If an edit would leave the table unbalanced, it is refused and the file is left untouched. Concretely:

- Multi-line rows (newer KiCad writes one field per line) are handled — a line-based edit silently mangles them, and there is a test for it.
- Batch removal cuts from the back, so offsets computed before the first deletion are still valid for the rest.
- `set_library_table_uri` replaces only the `(uri ...)` field inside the matched span, leaving `type` and `descr` intact and neighbouring rows untouched. Values go through `escape_sexpr_string`.

## URI resolution

`list_library_table` reports each entry with its URI resolved through `${KIPRJMOD}`, KiCad's configured path variables (read via the existing `PlatformHelper.load_kicad_env_vars`, which already exists for this purpose) and the environment, plus whether the file is actually there.

A variable that resolves to nothing is reported as **missing** rather than being treated as a literal path — `${KICAD9_SYMBOL_DIR}/Device.kicad_sym` on a machine where that variable is unset points nowhere, and saying "exists" there would defeat the purpose of the check.

Path variables are stored in the table verbatim; expansion happens only for reporting. Writing a resolved absolute path back into the table would break the project for everyone else who opens it.

## Test plan

- [x] 31 new tests in `tests/test_library_tables.py`, fully hermetic (`tmp_path`, no KiCad needed): field parsing, exact span boundaries, multi-line rows, `${KIPRJMOD}` and `${KIPRJMOD}/..` resolution, unresolved variables, missing files, removing the last row before `)`, batch removal from a multi-line table, unknown nicknames leaving the file byte-identical, and quote escaping in a new URI.
- [x] `pytest tests/` — 1770 passed. The 20 failures are pre-existing on `upstream/main` at `0dc3ee8` and unchanged by this branch.
- [x] `npm run build`, `npx vitest run` (70 passed), `eslint`, `prettier`, `black`, `isort`, `flake8`, `mypy` clean. README headline count updated to 149.
- [x] Against real data: 10 project tables from a hierarchical KiCad 10 project read correctly with their `${KIPRJMOD}/../` URIs resolved, and the global table at `AppData/Roaming/kicad/10.0/sym-lib-table` is found via `scope: "global"`. A round-trip on a copy (remove one entry, repoint another) leaves the table parseable.

Registered in all five places, including `registry.ts`, so `search_tools` finds them rather than adding to `KNOWN_UNREGISTERED`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,18 +14,42 @@ All notable changes to the KiCAD MCP Server project are documented here.
   a script — `content.replace(")", ...)` on the closing paren — corrupts a
   table whose last entry ends on the same line.
 
-  These work on parsed `(lib ...)` spans: entries are found by nickname, the
-  edit slices exactly that span, and **the result is re-parsed before it is
-  written**. A rewrite that would leave the table unbalanced is refused and the
-  file is left alone. Multi-line rows (which newer KiCad writes) are handled,
-  and batch removal cuts from the back so earlier offsets stay valid.
+  These work on parsed `(lib ...)` spans: rows are located as the direct
+  children of the table — never by scanning raw text for `(lib`, so an ordinary
+  Eagle-import description like `(descr "Caps (X7R), see (lib old)")` cannot
+  invent a phantom row — the edit slices exactly that span, and **the result is
+  re-parsed before it is written**. A rewrite that would leave the table
+  unbalanced is refused and the file is left alone. Multi-line rows (which newer
+  KiCad writes) are handled, and batch removal cuts from the back so earlier
+  offsets stay valid.
+
+  `scope="global"` resolves to the machine-wide table in KiCad's user config,
+  which every project on the machine loads, so writes are made through a
+  temporary file and `os.replace` rather than truncating in place, keep the
+  file's existing newline style instead of rewriting all 200-odd lines to CRLF,
+  and copy the previous contents into a sibling `.mcp-backups/` directory first.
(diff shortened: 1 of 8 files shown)


mixelpixx:
Merged. One resolution note: your branch predated #359's registry split, so the three table tools were placed in the then-combined library category. During merge I kept them in the library category (they span both sym-lib-table and fp-lib-table) and preserved the symbol_library category #359 introduced — the comment referencing #345 was dropped since that home now exists.

PR #365, opened 2026-08-06
feat: add backannotate_footprints (PCB footprints -> schematic)

karu2003:
## The gap

The server can push a schematic into a board (`sync_schematic_to_board`, the F8 equivalent) but has nothing going the other way.

That direction matters because after a layout pass the **board** is the side that is right. A footprint gets swapped in pcbnew — a 0402 that had to become an 0603, a connector changed for one that fits — and the schematic keeps the old value. Nothing complains, because nothing reads it, until the next ECO run pushes the stale schematic value back onto the board and quietly undoes the layout decision.

Imported projects start out in that state. Anything that came through `import_eagle_project` has schematic-side `Footprint` fields that never matched the placed parts, and today the only fix is to parse the `.kicad_pcb` by hand and patch each field.

## What this adds

`backannotate_footprints` copies the placed footprint into the schematic's `Footprint` field, matching by reference designator.

- Scans every `.kicad_sch` beside the board, so a hierarchical design is handled in one call; `schematicPath` narrows it to one sheet.
- Updates **all units** of a multi-unit symbol, not just the first one found.
- Skips virtual references starting with `#` — power and ground symbols carry no footprint, and writing one to them is wrong.
- Adds the field to instances that lack it (`addMissing`, default true); `references` limits the run to a list; `dryRun` reports the diff without writing.

Two behaviours worth calling out, because both are choices rather than accidents:

**Position and `hide` are inherited from the existing field.** Back-annotating a footprint must not move a field the user placed, and must not reveal one they hid. Only the value changes.

**The `lib_symbols` cache is left alone.** Instance fields override the cache when KiCad loads a sheet, so updating the instance is sufficient and correct. Rewriting the cache is a different operation with different risks, and silently doing both would make this tool much harder to reason about.

## Parsing

Structural, not regex-based, driven by two things found in real files.

Footprint names contain parentheses — `HRS_U.FL-R-SMT-1(10)` is a stock Hirose connector — so counting parens without tracking string state truncates the name mid-value and unbalances the file. And KiCad writes board files whose indentation does not track nesting, so indentation cannot be used as a shortcut for depth.

The s-expression walkers this needed, `match_paren` and `iter_child_offsets`, move into `utils/sexpr_format.py`. They already existed as private copies in `add_symbol_property` and `library_tables`; this replaces both with the shared version rather than adding a third.

## Test plan

24 new tests in `tests/test_backannotate_footprints.py`, plus 8 in `tests/test_sexpr_format.py` for the promoted helpers. They cover the cases where a simpler implementation breaks:

- footprint name containing parens, read from the board and written back intact
- board indentation that does not match nesting
- legacy `(fp_text reference ...)` boards alongside the modern `(property "Reference" ...)` form
- every unit of a multi-unit symbol updated
- `lib_symbols` cache left untouched
- `hide` and `(at ...)` preserved on update
- added field lands inside its own instance, not a neighbour's
- paren balance holds after every write
- second run is a no-op

Verified end to end on a real 71-footprint board outside the test suite: it found two genuine mismatches (`FOG_components:0402S` in the schematic where the board had `FOG_components:0402HF`), `kicad-cli sch upgrade` accepted the result without changes, and a second pass reported nothing to do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,25 +14,51 @@ All notable changes to the KiCAD MCP Server project are documented here.
   projects have the same problem from day one: the schematic-side fields never
   matched the placed parts.
 
-  Matches by reference designator across every `.kicad_sch` beside the board (or
-  one named sheet), updates all units of a multi-unit symbol, and skips virtual
-  references starting with `#` (power and ground symbols carry no footprint).
-  Instances with no `Footprint` field get one added unless `addMissing` is
-  false; `dryRun` reports the diff without writing. Position and `hide` are
-  carried over from the existing field, so back-annotating does not move or
-  reveal a field the user had placed. The `lib_symbols` cache is left alone —
-  instance fields override it, and rewriting the cache is a different tool.
+  Matches by reference designator, updates all units of a multi-unit symbol, and
+  skips virtual references starting with `#` (power and ground symbols carry no
+  footprint). Instances with no `Footprint` field get one added unless
+  `addMissing` is false; `dryRun` reports the diff without writing. The
+  `lib_symbols` cache is left alone — instance fields override it, and rewriting
+  the cache is a different tool.
+
+  The designators come from each symbol's `(instances ...)` block, which is
+  where KiCad 7+ keeps one per sheet instance; the `(property "Reference" ...)`
+  holds only one of them. A sub-sheet placed twice therefore contributes both
+  designators, and because those instances share a single `Footprint` field,
+  nothing is written when the board disagrees between them — the references and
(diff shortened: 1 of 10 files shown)


mixelpixx:
Merged. This also landed the shared match_paren/iter_child_offsets helpers in sexpr_format.py — your other PRs' identical copies merged as clean no-ops on top, which is how coordinated batches should be built. The non-recursive sheet discovery that filters backup paths is a quietly important safety property: a backannotation that also rewrites the backup destroys the fallback at the exact moment it might be needed.

PR #366, opened 2026-08-06
feat: add set_symbol_pin_type (bulk pin electrical type edits)

karu2003:
## The gap

The server can read pins — `list_symbol_pins`, `batch_list_symbol_pins` — but has no way to write them. Changing pin electrical types in bulk means `sed` over the `.kicad_sym` file.

That is unsafe in three ways, all of which show up on real libraries:

**It rewrites things that are not pins.** `sed 's/bidirectional line/passive line/'` also hits a symbol named `XCVR_bidirectional line driver`, a `Description` reading `"pin bidirectional line buffer"`, and — the one that actually bites — `(alternate "SPI_CLK" output line)`, which is a *function of* a pin rather than a pin. The alternate silently changes meaning and ERC starts reporting something new.

**It cannot see which symbol it is standing on.** "Make only the shield pins passive" is not expressible; you get the whole file or nothing.

**Nothing checks the replacement token.** KiCad refuses to load a library containing a pin type it does not recognise, and the error it reports names the file, not the pin. A typo costs a bisect.

## What this adds

`set_symbol_pin_type` sets the electrical type and/or graphic style of pins in a `.kicad_sym`, filtered by `symbols`, `pinNumbers`, `pinNames`, or `fromType`. It validates the new token against KiCad's sets **before** touching the file, and rewrites only the pin head.

The usual reason to need it: symbols converted from Eagle or pulled from SnapEDA arrive with every pin `unspecified` or `bidirectional`. ERC then reports conflicts on nets that are electrically fine, and the noise hides the real errors. `fromType: "unspecified", type: "passive"` is the whole fix.

Two details aimed at the same failure mode — a call that quietly does nothing:

- A symbol name that is not in the library comes back in `missingSymbols`, and the message says so, rather than reporting a cheerful zero.
- Pins that already have the requested type are counted in `alreadyCorrect` and distinguished from pins that never matched the filters.

`dryRun` returns the change list without writing.

## Implementation notes

Pins are located by walking the file with a symbol stack rather than by regex. Two reasons:

- A pin lives inside a unit sub-symbol (`OPAMP_DUAL_2_1`) whose name differs from the symbol the caller asked for (`OPAMP_DUAL`). The stack keeps the pin attributed to the top-level symbol while still reporting which unit it came from.
- `(pin_names (offset 0.254))` and `(pin_numbers hide)` share a prefix with `(pin ...)` and are not pins.

The walk skips quoted text, so a pin named `OUT(A)` does not unbalance the scan.

The token lists live in a new dependency-free `utils/pin_types.py` so the JSON schema's `enum` and the runtime validator cannot drift apart.

This branch also brings in `match_paren` / `iter_child_offsets` in `utils/sexpr_format.py`. That hunk is byte-identical to the one in #365, #367 and #368, so it merges in whichever order those land.

The rest of the branch does need a rebase once any sibling PR lands — an earlier note here said otherwise, which was only true of the `sexpr_format.py` hunk. Merging all seven locally, the overlaps are `CHANGELOG.md`, the hand-maintained tool counters in `README.md`, the dispatch table in `python/kicad_interface.py`, and `src/tools/library-symbol.ts` (shared with #367). All of them resolve additively: keep both entries and sum the counts. `tests-ts/readme-counts.test.ts` then checks the README figure against `getRegistryStats()`, so an arithmetic slip fails CI rather than shipping. With all seven applied the total is 157 tools across 15 categories.

## Test plan

28 new tests in `tests/test_set_symbol_pin_type.py`, weighted toward the cases where the `sed` approach breaks:

- a type word inside a symbol name and inside a `Description` is left alone
- `(alternate "SPI_CLK" output line)` is left alone while its parent pin changes
- every unit of a multi-unit symbol is covered, and reported per unit
- `(pin_names ...)` / `(pin_numbers hide)` are not counted as pins
- the `(name ...)` of a nested font block is not mistaken for the pin name
- a pin name containing parens
- the single-line pin form that script-generated libraries use
- unknown type and unknown style are refused with the file untouched
- a `.kicad_sch` passed by mistake is refused by root form
- misspelled symbol name reported; already-correct pins counted, not rewritten
- `dryRun` writes nothing

Verified outside the suite on a real 2.2 MB library: 2279 pins across 472 symbols walked correctly, 32 `bidirectional` pins on one symbol changed to `passive`, paren count unchanged, `kicad-cli sym upgrade` accepted the result without modifying it, and a second pass reported nothing to do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,34 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ## [Unreleased]
 
+### New Tools
+
+- **`set_symbol_pin_type`** — set the electrical type, and optionally the
+  graphic style, of pins in a `.kicad_sym` library, filtered by symbol, pin
+  number, pin name, or current type. The server could read pins
+  (`list_symbol_pins`, `batch_list_symbol_pins`) but not write them, so bulk
+  fixes were done with `sed` over the library file.
+
+  That is unsafe in ways that are easy to miss. A blind substitution also
+  rewrites the words it matches inside symbol names, `Description` properties
+  and `(alternate "SPI_CLK" output line)` pin functions; it cannot tell which
+  symbol it is standing on, so "only the shield pins" is not expressible; and
+  nothing validates the replacement token, which matters because KiCad refuses
+  to load a library containing a pin type it does not recognise and reports the
+  file rather than the pin. This tool checks the token against KiCad's sets
+  before touching the file, and rewrites only the pin head.
+
+  The usual reason to need it: symbols converted from Eagle or pulled from
+  SnapEDA arrive with every pin `unspecified` or `bidirectional`, and ERC then
+  reports conflicts on nets that are electrically fine, burying the real errors.
+  `fromType` makes that a one-liner. A misspelled symbol name is reported in
+  `missingSymbols` rather than silently matching nothing, and `dryRun` shows
+  the list first.
+
+  Verified on a 2.2 MB library: 2279 pins across 472 symbols walked correctly,
+  32 pins changed, paren balance unmoved, `kicad-cli sym upgrade` accepted the
+  result, and a second pass was a no-op.
+
 ### Bug Fixes
 
 - **`add_layer` could not add inner copper layers, and corrupted an unrelated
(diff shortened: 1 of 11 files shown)


mixelpixx:
Merged. Validating pin types pre-write because KiCad's load error names the file rather than the pin is exactly the right reasoning — the failure would otherwise surface far from its cause. The shared pin_types.py module keeping the schema and the command in agreement is the pattern the rest of the repo should adopt.

PR #367, opened 2026-08-06
feat: add find_duplicate_symbols (the same part stored twice)

karu2003:
## The gap

A library collects duplicates the moment more than one source feeds it. An Eagle import lands a part the curated library already had. A SnapEDA download arrives under the vendor's naming. Someone re-adds a part because search did not find the existing name.

KiCad reports none of this, because the names differ — which is also why grepping for the name does not find it. The library grows, the BOM picks whichever symbol the schematic happened to reference, and nobody notices until two boards ship with different fields for the same chip.

## What this adds

`find_duplicate_symbols` groups symbols that look like the same part, and says on what evidence.

| strategy | groups on |
|---|---|
| `mpn` | manufacturer part number |
| `supplier` | distributor part number, for parts with no MPN field |
| `value_footprint` | same Value on the same Footprint |
| `graphics` | byte-identical drawn body |
| `name` | name ignoring case and separators |

Two design points came out of running it on a real 489-symbol library rather than from reasoning about the format.

**MPN matching has to normalise property names first.** That library holds the manufacturer part number as `MPN`, `MP`, `MANUFACTURER PART NUMBER` *and* `PART NUMBER`, depending on which importer wrote each symbol. A plain group-by on any one of those finds nothing. The tool compares under a normalised key and reports which property each symbol's value came from, so the result stays auditable.

**`graphics` is off by default.** Matching on the drawn body alone put 78 different resistor values into one group, because every resistor in a library shares one drawing. It is genuinely useful for finding a custom IC copied under a new name, so it stays available — documented as evidence rather than as a verdict. Symbols that `extends` another are excluded from it entirely: sharing a body is what `extends` is for.

## Making the report actionable

`schematicPaths` (files or directories, searched recursively) counts instances per symbol across the project. That is what turns a list of suspicious pairs into a decision — the duplicate nothing places is the one to retire. Each group carries `usageCount` and `usedIn` per member, an `unusedMembers` list, and a `suggestedKeep` with the reason it was chosen (most used; on a tie, the one with the fullest fields; deterministic after that).

A group found by several strategies is reported once with all the evidence attached, rather than once per strategy.

On the library above, the defaults reported 86 groups covering 116 redundant symbols, 69 of which no schematic places. One group holds five symbols for the same TI transceiver — `CC1200EP`, `CC1200EP-1`, `IC_CC1200RHBT`, `IC_CC1200RHBT_QFN`, `IC_FOG-CC1200-PA` — of which exactly one is in use.

## Test plan

32 tests in `tests/test_find_duplicate_symbols.py`:

- MPN match across two differently-named properties holding the same value, case-insensitively, with `ignoreCase: false` proving the comparison is the reason
- the property that supplied each key is reported, per group and per member
- `value_footprint` skips symbols missing either field
- `graphics` finds a body copied under a new name whose fields share nothing
- `graphics` is off by default, with the resistor-family finding recorded as the reason
- a symbol using `extends` is never a graphics duplicate
- three strategies agreeing produce one group, not three
- usage counts ignore the library nickname, since the same library is registered under different nicknames per project
- a directory of sheets is expanded; a missing sheet path does not abort the scan
- the suggested symbol is the used one, and the message says when no usage data was supplied
- unknown strategy, `.kicad_sch` passed by mistake, missing file, empty library, library with no duplicates

This branch also brings in `match_paren` / `iter_child_offsets` in `utils/sexpr_format.py`; that hunk is byte-identical to the one in #365, #366 and #368, so it merges in whichever order those land.

The rest of the branch does need a rebase once any sibling PR lands — an earlier note here said otherwise, which was only true of the `sexpr_format.py` hunk. Merging all seven locally, the overlaps are `CHANGELOG.md`, the hand-maintained tool counters in `README.md`, the dispatch table in `python/kicad_interface.py`, `python/schemas/tool_schemas.py`, and `src/tools/library-symbol.ts` (shared with #366). All of them resolve additively: keep both entries and sum the counts. `tests-ts/readme-counts.test.ts` then checks the README figure against `getRegistryStats()`, so an arithmetic slip fails CI rather than shipping. With all seven applied the total is 157 tools across 15 categories.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,35 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ## [Unreleased]
 
+### New Tools
+
+- **`find_duplicate_symbols`** — group symbols in a `.kicad_sym` that are the
+  same part stored twice under different names. Libraries collect these the
+  moment more than one source feeds them: an Eagle import lands a part the
+  curated library already had, a SnapEDA download arrives under the vendor's
+  naming, someone re-adds a part because search did not find the existing name.
+  KiCad reports nothing, because the names differ — which is also why grepping
+  for the name does not find it.
+
+  Matches on manufacturer part number, distributor part number, Value +
+  Footprint, an identical drawn body, or near-identical names. The MPN match
+  normalises property names first, because a real library holds that field as
+  `MPN`, `MP`, `MANUFACTURER PART NUMBER` and `PART NUMBER` depending on which
+  importer wrote it, and a plain group-by therefore finds nothing. Each group
+  reports which strategy and which property produced it.
+
+  `schematicPaths` counts instances per symbol across the project's sheets,
+  which turns the report into a decision: the duplicate nothing places is the
+  one to retire. `suggestedKeep` names it and says why.
+
+  `graphics` is off by default. Verified against a 489-symbol library: matching
+  on the drawn body alone put 78 different resistor values in one group, because
+  every resistor shares one drawing. It stays available for hunting a custom
+  part copied under a new name, and is documented as evidence rather than
+  verdict. With the defaults the same library reported 86 real groups covering
+  116 redundant symbols, 69 of which no schematic places — including five
+  symbols for one TI transceiver, of which one is in use.
+
 ### Bug Fixes
 
 - **`add_layer` could not add inner copper layers, and corrupted an unrelated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,23 +15,41 @@ All notable changes to the KiCAD MCP Server project are documented here.
   for the name does not find it.
 
   Matches on manufacturer part number, distributor part number, Value +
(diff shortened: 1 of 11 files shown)


mixelpixx:
Merged, last of your batch as the conflict hub — the registry.ts and library-symbol.ts overlaps with #364/#366 were resolved by hand (one splice needed repair, caught by the TypeScript build before push). Excluding graphics and name matching from the defaults with stated reasons, and keeping blank markers (N/A, TBD, ~) out of grouping keys so importer gaps don't collapse into one giant group, both show real-library testing.

Whole-batch note: all seven PRs computed README tool counts from the same pre-merge baseline, so I re-derived them at each merge — final state is 169 tools across 15 categories, enforced by the readme-counts test. Thank you for an unusually well-built batch: every PR used the escape-aware helpers, none reinvented the unsafe quoted-value regex, and the per-PR test suites reproduced the actual failure shapes.

PR #368, opened 2026-08-06
feat: add Digi-Key Product Information V4 integration (search, library availability sweep)

karu2003:
## The gap

The server has six JLCPCB tools and zero for Digi-Key. Stock checks, lifecycle checks and replacement hunts all happen in one-off scripts outside it — which means the answers never make it back into the library, and the next person starts over.

## What this adds

Three tools, in `src/tools/digikey-api.ts` and `python/commands/digikey.py`.

`digikey_search_parts` — keyword search by part number, MPN, or a parametric phrase (`0402 100nF 50V X7R`). Returns stock, price, lifecycle status, every packaging variation with its own pricing breaks, and the parametric table.

`digikey_check_library_availability` — sweeps a `.kicad_sym` and reports which parts are obsolete, out of stock, unfindable, or carry no part number at all. This is the check to run before a board goes to purchasing, and after inheriting a library from an import. Each symbol is searched by its distributor part number first and its MPN second, because retired Digi-Key numbers are the common failure and the MPN usually still resolves; `searchByMpnFirst` reverses it. Part numbers are read under any of the spellings real libraries use — `MPN`, `MP`, `MANUFACTURER PART NUMBER`, `PART NUMBER`, `SUPPLIER PART NUMBER 1` — because a library that has seen more than one importer holds the same field under several names.

`digikey_test_connection` — separates *missing* credentials from *rejected* credentials from a locale misconfiguration, which are three very different fixes that all surface as "it doesn't work".

## Credentials never enter the source tree

This was the hard requirement, so it is worth being explicit about how it is enforced rather than asserted.

- Read from `DIGIKEY_CLIENT_ID` and `DIGIKEY_CLIENT_SECRET` in the server environment, optionally via a `.env` at the repo root. `.env` was already gitignored; a test now asserts that, so the guarantee cannot be quietly removed.
- **Absent from every tool schema.** A key passed as a tool argument ends up in the conversation transcript, in the MCP log, and in anything that replays the call. A test passes `clientId` / `clientSecret` / `apiKey` to a tool and asserts the call is refused rather than honoured.
- Every string leaving the module — error messages included — goes through one redaction step. This is not paranoia: Digi-Key echoes the client id in some error bodies. Tests cover redaction on an HTTP error body and on a transport exception.
- `.env.example` gains the variable names with placeholder values, and a test asserts the placeholders stay placeholders.
- The `.env` loader never overrides an already-exported variable, so a stale file cannot shadow the operator's credentials.

## Three things the API does not tell you

Each of these costs an afternoon to rediscover, and all three are the difference between "works" and "works on my account".

**The locale headers are mandatory.** Without `X-DIGIKEY-Locale-Site` and its two companions, every search returns **404** — which reads like a wrong URL, so the natural next move is to go hunting for the right endpoint. The client always sends them; the values are configurable via `DIGIKEY_LOCALE_*` or per call, defaulting to US/en/USD. A 404 is reported with the locale actually sent, so the next person does not repeat the hunt.

**The Digi-Key part number is not on the product.** It lives on each entry of `ProductVariations`, one per packaging option, so the reel and the cut tape have different numbers. `preferPackaging` chooses which is quoted as the headline, and matching handles localized packaging names — a German account calls cut tape `Gurtabschnitt`, so matching only `"Cut Tape"` silently hands back the 10 000-piece reel to someone ordering prototypes. All variations are returned regardless.

**`ProductStatus.Status` is localized.** A German account returns `Aktiv` and `Obsolet`. Code that compares the string against `"Active"` reports every part in the library as a problem. Lifecycle is read from `ProductStatus.Id`; the string is passed through unchanged for whoever reads the report.

## Token and rate handling

Token cached with a one-minute safety margin so a request cannot start on one that expires mid-flight; one forced refresh and retry on a 401 mid-flight, and no retry beyond that so bad credentials fail fast rather than looping; `Retry-After` honoured on 429; a minimum interval between requests, since the library sweep is one request per symbol.

## Test plan

37 tests in `tests/test_digikey.py`. **No test performs a real request** — `requests.post` is monkeypatched, matching the pattern in `tests/test_jlcpcb_live_api.py`, and an autouse fixture clears `DIGIKEY_*` from the environment and stubs the `.env` loader so a developer's real credentials can neither be used nor leak into a failure message.

Coverage: the credential guarantees listed above; locale headers always sent and configurable; 404 explained as a locale problem; part number taken from a variation; cut tape selected by its localized name; lifecycle from the id rather than the string; token reuse, expiry refresh, 401-refresh-and-retry, persistent 401, 429 backoff; search normalisation, empty results, limit clamping; the library sweep classifying available / inactive / no-part-number, falling back to the MPN, respecting `maxSymbols`, and returning partial results when a request fails partway.

**Not verified against the live API.** No working Digi-Key credentials were available while writing this, so the response handling is written against a captured V4 response and the documented behaviour rather than a live call. Saying so seems better than implying a test run that did not happen. `digikey_test_connection` exists partly so a reviewer with credentials can check it in one call.

Everything else green: `pytest` (full suite: 1783 passed; the 20 failures are pre-existing and identical on a clean `main`), `black`, `isort`, `flake8`, `mypy`, `tsc --noEmit`, `eslint`, `prettier`, `vitest` (70 passed, including the registry-completeness and README-count guards).

This branch also brings in `match_paren` / `iter_child_offsets` in `utils/sexpr_format.py`; that hunk is byte-identical to the one in #365, #366 and #367, so it merges in whichever order those land.

The rest of the branch does need a rebase once any sibling PR lands — an earlier note here said otherwise, which was only true of the `sexpr_format.py` hunk. Merging all seven locally, the overlaps are `CHANGELOG.md`, the hand-maintained tool counters in `README.md`, the dispatch table in `python/kicad_interface.py`, and `src/tools/index.ts` (shared with #363). All of them resolve additively: keep both entries and sum the counts. `tests-ts/readme-counts.test.ts` then checks the README figure against `getRegistryStats()`, so an arithmetic slip fails CI rather than shipping. With all seven applied the total is 157 tools across 15 categories.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,3 +10,26 @@
 JLCPCB_APP_ID=your_app_id
 JLCPCB_API_KEY=your_access_key
 JLCPCB_API_SECRET=your_secret_key
+
+# Digi-Key Product Information V4 credentials.
+#
+# Create an app at developer.digikey.com. It must be a Production app with
+# Product Information V4 enabled — a Sandbox app fails authentication against
+# these endpoints.
+#
+# The Digi-Key tools read these two variables and nothing else: credentials are
+# never accepted as tool arguments, so they cannot end up in a chat transcript,
+# a saved workflow, or a log line.
+
+DIGIKEY_CLIENT_ID=your_client_id_here
+DIGIKEY_CLIENT_SECRET=your_client_secret_here
+
+# Optional. Digi-Key requires locale headers on every request — without them
+# a search returns 404 — but the values are yours to pick. Defaults below.
+# Note that this also localizes the response: the lifecycle status and the
+# packaging names come back in the language requested, which is why the tools
+# compare ids rather than those strings.
+
+# DIGIKEY_LOCALE_SITE=US
+# DIGIKEY_LOCALE_LANGUAGE=en
+# DIGIKEY_LOCALE_CURRENCY=USD
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,63 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ## [Unreleased]
 
+### New Tools
+
+- **Digi-Key Product Information V4 integration** — `digikey_search_parts`,
+  `digikey_check_library_availability` and `digikey_test_connection`. The server
+  had six JLCPCB tools and nothing for Digi-Key, so every stock check, lifecycle
+  check and replacement hunt happened in one-off scripts outside it.
+
+  `digikey_search_parts` searches by part number, MPN, or a parametric phrase and
+  returns stock, price, lifecycle status, every packaging variation and the
+  parametric table. `digikey_check_library_availability` sweeps a `.kicad_sym`
+  and reports which parts are obsolete, out of stock, unfindable, not orderable,
+  or carry no part number at all — searching by distributor number first and MPN
+  second, because retired Digi-Key numbers are the common failure and the MPN
+  usually still resolves. Part numbers are read under any of the property
+  spellings real libraries use. A symbol whose lookup fails is marked
+  `state: "error"` and the sweep continues, so one transient 500 does not throw
+  away the lookups already paid for; five consecutive failures stop it, because a
+  revoked key fails identically for every remaining symbol. The sweep costs up to
+  two rate-limited requests per symbol, so `maxSymbols` defaults to 25 and the
+  command is granted the extended Node-side timeout in `src/command-timeout.ts` —
+  on the 30 s default the timer fires mid-sweep and the caller loses every lookup
+  the sweep had already paid for.
+
+  Three things the API makes easy to get wrong:
+  - The locale headers are mandatory. Without `X-DIGIKEY-Locale-Site` and its two
+    companions a search returns **404**, which reads like a wrong URL. The client
+    always sends them; the values are configurable.
+  - The Digi-Key part number is not on the product. It lives on each entry of
+    `ProductVariations`, so the reel and the cut tape have different numbers.
+    `preferPackaging` picks which one is quoted and matches localized packaging
+    names. A product with no variations at all is reported as `not_orderable`
+    rather than available, since there is no number to put on an order.
+  - `ProductStatus.Status` is localized, so comparing it against `"Active"`
+    reports every part as a problem once the account is not set to English.
+    Lifecycle is read from `ProductStatus.Id`.
+
+  **Credentials never enter the source tree.** They are read from
+  `DIGIKEY_CLIENT_ID` and `DIGIKEY_CLIENT_SECRET` in the environment, optionally
+  via the already-gitignored `.env`, and are deliberately absent from every tool
+  schema — a key passed as a tool argument would be recorded in the conversation,
+  in the MCP log, and in anything replaying the call. Absent from the schema means
+  such an argument is _ignored_, not rejected: the MCP layer strips unknown keys,
+  so the tools now return a `warnings` entry naming the argument and telling the
+  caller to rotate the value, which is the honest description of what happened.
+  Everything leaving the module passes through a redaction step, and a non-JSON
+  response body or a hostile `Retry-After` is converted into a redacted error
+  rather than escaping as an unfiltered traceback. Tests assert all of it,
+  including that the schemas contain no credential-shaped field, that `.env` is
+  gitignored, and that `.env.example` carries names without values.
+
+  Nothing here has been exercised against the live Digi-Key service: there were no
+  working credentials available (the token endpoint answers
+  `401 invalid_client`), so every test drives a mocked transport and the
+  German-language strings in the fixtures are constructed illustrations of the
+  localization hazard rather than captured responses. `ACTIVE_STATUS_ID = 0` is
+  documented in the source as an assumption to confirm on first real use.
+
 ## [2.7.0] - 2026-08-20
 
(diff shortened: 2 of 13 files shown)


mixelpixx:
Fit decision: accepted. Deciding factors — JLCPCB set the precedent that supplier data belongs in this server, requirements.txt has declared the Digi-Key intent since before this PR ('for JLCPCB/Digikey APIs - future', line 26), no new dependencies, and the credential handling here is the best I have reviewed on this repo: env-only, the _CREDENTIAL_ARG guard actively rejecting credential-shaped tool arguments with the correct reasoning that an argument is already in the transcript, a single redaction chokepoint with a min-length guard, and override=False on .env loading. The command-timeout analysis (a timed-out sweep desyncing the single-threaded worker's subsequent responses) was independently confirmed and is now fixed at the protocol level in #382.

Two changes before merge:

1. Add python/schemas/tool_schemas.py entries for the three tools. The existing JLCPCB tools lack them too, which is a fault to fix, not a precedent to extend — supplier tools should be discoverable in the Python-side tools/list like everything else.
2. Rebase and re-derive the README tool counts from getRegistryStats() — main has moved a lot this week (currently 169 tools / 15 categories) and this PR's 149 was computed against a stale base. The readme-counts test pins the first occurrence; update both.

There is also a small src/tools/index.ts conflict with the now-merged #363 — trivial rebase. With those done, this merges.

karu2003:
Both changes are in, rebased onto `aa53d52` (v2.7.0). The `src/tools/index.ts` conflict is resolved by keeping both exports.

**1. `python/schemas/tool_schemas.py`** — added `DIGIKEY_TOOLS` with all three entries and included it in the combine loop, so `tools/list` on the Python side returns real schemas instead of the annotation-derived best effort. Four tests came with it:

- each tool is registered and carries `title` / `description` / an object `inputSchema`;
- the property names match the zod shape in `src/tools/digikey-api.ts` exactly, parsed out of the source per `server.tool()` block. This one is not decorative — it failed on first run because my regex required `z.` and prettier had wrapped the longer builders onto the next line, so it was reading `keywords` and `preferPackaging` as absent;
- `required` is `["keywords"]` / `["libraryPath"]` / absent, and `maxSymbols.default` is pinned to `digikey.DEFAULT_MAX_SYMBOLS` rather than a literal;
- no property on any of the three is credential-shaped, which is the same assertion `tests-ts/digikey-schemas.test.ts` makes about the zod side. Adding a second schema surface is exactly where that guarantee could have been quietly dropped.

**2. README counts** — re-derived from `getRegistryStats()` on the merged branch, not by arithmetic: **172 tools, 16 categories**, 147 routed, 32 direct. The `readme-counts` test passes against that.

One thing to flag, since it goes past what you asked for. The README has five other count statements besides the pinned one, and several were already stale on `main` before this PR: `22 direct tools` (actual 32), `113 routed tools` organized into `14 categories` (actual 147 in 16, and the parenthesised list was missing `symbol_library`), `all 122 tool schemas`, `Capabilities: Tools (122)`, and `Working Features (148 tools)`. I updated all of them to the derived values, so the file is internally consistent — but that is a doc fix that is not about Digi-Key, and I am happy to strip it back to just the pinned occurrence if you would rather keep this PR narrow.

**On #382** — thanks for confirming the timeout analysis, and it changes what the code should claim. The comment in `src/command-timeout.ts` and the docstring on `test_the_node_side_grants_the_sweep_an_extended_timeout` both described the late response being delivered to the caller's *next* tool call. That is no longer true: correlation discards it by id. Both now say the sweep still needs the long allowance — up to two rate-limited requests per symbol cannot fit in 30 s — and that what a timeout costs is this command's whole result set, every lookup in it already charged against the caller's Digi-Key rate limit. The CHANGELOG entry lost the same stale claim. The entry moved to `[Unreleased]`, since 2.7.0 has shipped.

Unchanged: there is still no live verification. There were no working Digi-Key credentials on this machine — the token endpoint answers `401 invalid_client` — so every test drives a mocked transport, the German-language fixture strings are constructed illustrations rather than captured responses, and `ACTIVE_STATUS_ID = 0` remains documented as an assumption to confirm on first real use. `digikey_test_connection` exists so that confirmation is one call.

I squashed the branch to a single commit while rebasing, because the five original commits each conflicted on `CHANGELOG.md` and `README.md` and resolving the same hunk five times is how a wrong resolution gets in.

Locally: `pre-commit` clean on the full hook set (`black` did reformat the new schema block; that is committed), `tsc` builds, 80/80 vitest, and the Python suite is at 2208 passed / 21 failed — the same 21 failures `main` produces on this machine, all environmental (no `kicad-skip`, no `kicad-cli` on PATH). `tests/test_digikey.py` is 88/88.

mixelpixx:
Merged, thank you. The `DIGIKEY_TOOLS` block, the 70 Python and 7 TypeScript tests, and the credential posture (environment only, one redaction chokepoint, credential-shaped arguments ignored with a warning naming them, the minimum-length guard) all match what was reviewed in August, and CI was green on all 22 checks. It rebased onto v2.7.0 cleanly; the only conflicts were CHANGELOG ordering and the README counts, resolved to 173 tools / 16 categories because #376 landed in the same batch. The two items you flagged stay as they are: the `sort_keys` churn is cosmetic, and `ACTIVE_STATUS_ID = 0` is documented in the source as an assumption for whoever first runs this against a live Production app. If you get working credentials, a one-line follow-up confirming or correcting that constant would be welcome.

PR #378, opened 2026-08-17
fix: kipy 0.7.x / KiCad 10 IPC backend compatibility (IPC 后端兼容修复)

AlloyPlane:
## 修复内容 / What this fixes

让 KiCAD-MCP-Server 的 IPC 后端在 **kicad-python 0.7.x + KiCad 10** 环境下可用（实时 UI 同步）。
Fixes the IPC backend so it works with **kicad-python 0.7.x + KiCad 10** (real-time UI sync).

### 问题 / Problem
在 KiCad 10.0.5 + kicad-python 0.7.1 下，IPC 后端连接失败并静默回退到 SWIG（`_realtime: false`），
因为服务器代码针对的是旧版 kipy API。另外 KiCad 10 移除了 `pcbnew.GetGlobalFootprintLib`。
Under KiCad 10.0.5 + kicad-python 0.7.1, the IPC backend fails to connect and silently
falls back to SWIG (`_realtime: false`), because the server code targets an older kipy API.
Also, KiCad 10 removed `pcbnew.GetGlobalFootprintLib`.

### 改动 / Changes
1. **ipc_backend.py – connect()**: `KiCad()` 自动拨号 + 用 `get_version()` 验证（新 kipy 的 `ping()` 会超时）
   `KiCad()` auto-dial + verify via `get_version()` (new kipy's `ping()` times out)
2. **ipc_backend.py – board path**: `get_open_documents(DocumentType.DOCTYPE_PCB)`（新 kipy 需要参数），
   并用 `board_filename` + `project.path` 组装板文件路径（`DocumentSpecifier` 已无 `path` 字段）
   `get_open_documents(DocumentType.DOCTYPE_PCB)` (new kipy requires the arg) and assemble the
   board path from `board_filename` + `project.path` (no more `path` field on DocumentSpecifier)
3. **ipc_backend.py – footprint loading**: 改为按 `.pretty` 目录 `pcbnew.FootprintLoad`
   （KiCad 10 移除了 `GetGlobalFootprintLib`）
   load footprints by `.pretty` dir path (KiCad 10 removed `GetGlobalFootprintLib`)
4. **kicad_interface.py – _ipc_place_component**: 当 `footprint` 参数缺失时回退到 `componentId`
   fall back to `componentId` when the `footprint` param is missing
5. **command-timeout.ts**: 将 `open_project` / `create_project` / `open_board` / `place_component`
   等加入长时命令列表（SWIG 初始化可能超过 30s）
   treat `open_project` / `create_project` / `open_board` / `place_component` as long-running
   (SWIG init can exceed 30s)

### 测试 / Testing
- 环境: Windows 11 · KiCad 10.0.5（便携版）· kicad-python 0.7.1 · Node 24 / TS 2.6.0
- Environment: Windows 11 · KiCad 10.0.5 (portable) · kicad-python 0.7.1 · Node 24 / TS 2.6.0
- ✅ `open_project` → `_backend: "ipc"`, `_realtime: true`（实时会话）
- ✅ IPC `place_component` → 成功，KiCad GUI 实时显示（含自定义库封装）
- ✅ `npm run build` 通过；冒烟测试通过

### 相关 / Related
- 会话固定机制（session pinning, #223）在 KiCad 10 下正常：板子需在 GUI 中打开才能 pin IPC
- session pinning (#223) works under KiCad 10: the board must be open in the GUI to pin IPC

---
中文说明：本 PR 修复 KiCad 10 + kicad-python 0.7.x 环境下 IPC 实时控制不可用的问题，共 5 处改动，
已在 Windows 上实测（实时放件成功）。如有任何格式或命名规范问题，请告知，我会调整。谢谢！

### DIFF
--- a/python/kicad_api/ipc_backend.py
+++ b/python/kicad_api/ipc_backend.py
@@ -145,8 +145,10 @@ class IPCBackend(KiCADBackend):
 
         def worker() -> None:
             try:
-                kicad = KiCad(socket_path=path) if path else KiCad()
-                kicad.ping()  # returns None on success, raises on failure
+                # kipy 0.7.1: KiCad() auto-dials the running KiCad; ping() can
+                # time out even when connected, so verify with get_version().
+                kicad = KiCad() if not path else KiCad(socket_path=path)
+                kicad.get_version()
                 result["kicad"] = kicad
             except Exception as e:  # noqa: BLE001 — surfaced to caller below
                 result["error"] = e
@@ -679,29 +681,33 @@ class IPCBoardAPI(BoardAPI):
                 lib_name = None
                 fp_name = footprint_path
 
-            # Get the footprint library table
-            fp_lib_table = pcbnew.GetGlobalFootprintLib()
+            # KiCad 10 removed GetGlobalFootprintLib; load by .pretty dir path.
+            def _pretty_dirs() -> list:
+                roots = [
+                    os.environ.get("KICAD_FOOTPRINTS_DIR", r"D:\kicad\share\kicad\footprints"),
+                    r"D:\kicad\custom-lib",
+                    os.path.expanduser(r"~\Documents\KiCad\10.0\libraries"),
+                ]
+                dirs: list = []
+                for root in roots:
+                    if lib_name:
+                        dirs.append(os.path.join(root, lib_name + ".pretty"))
+                    elif os.path.isdir(root):
+                        dirs.extend(
+                            os.path.join(root, d)
+                            for d in os.listdir(root)
+                            if d.endswith(".pretty")
+                        )
+                return dirs
 
-            if lib_name:
-                # Load from specific library
+            for pretty_dir in _pretty_dirs():
                 try:
-                    loaded_fp = pcbnew.FootprintLoad(fp_lib_table, lib_name, fp_name)
+                    loaded_fp = pcbnew.FootprintLoad(pretty_dir, fp_name)
                     if loaded_fp:
-                        logger.info(f"Loaded footprint '{fp_name}' from library '{lib_name}'")
+                        logger.info(f"Loaded footprint '{fp_name}' from {pretty_dir}")
                         return loaded_fp
                 except Exception as e:
-                    logger.warning(f"Could not load from {lib_name}: {e}")
-            else:
-                # Search all libraries for the footprint
-                lib_names = fp_lib_table.GetLogicalLibs()
-                for lib in lib_names:
-                    try:
-                        loaded_fp = pcbnew.FootprintLoad(fp_lib_table, lib, fp_name)
-                        if loaded_fp:
-                            logger.info(f"Found footprint '{fp_name}' in library '{lib}'")
-                            return loaded_fp
-                    except:
-                        continue
+                    logger.debug(f"FootprintLoad failed for {pretty_dir}: {e}")
 
             logger.warning(f"Footprint '{footprint_path}' not found in any library")
             return None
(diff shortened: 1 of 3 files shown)


mixelpixx:
Reviewed. The kipy 0.7 / KiCad 10 claims are plausible and consistent with what #369 independently found, and the two-arg pcbnew.FootprintLoad(libPath, fpName) call is the correct KiCad 10 signature. But there is one blocker and two structural asks before this can merge.

**Blocker — hardcoded developer-machine paths.** _pretty_dirs() ships two absolute D:\ paths from your own portable install (D:\kicad\share\kicad\footprints as the *default* for the env-var escape hatch, and D:\kicad\custom-lib unconditionally), plus Windows-only separators that mean nothing on macOS/Linux. On every other user's machine this silently probes the wrong disk. The correct replacement already exists in this repo: python/commands/library.py has _get_global_fp_lib_table() (line 60) and _parse_fp_lib_table() (line 84), with platform roots handled by python/utils/platform_helper.py. Route footprint discovery through fp-lib-table and hand the resolved .pretty directory to the two-arg FootprintLoad — only the path discovery needs replacing, the load call itself is right.

**Second:** this adds a third copy of the DocumentSpecifier decoding (#369's in get_open_board_path, yours in _place_loaded_footprint, plus the legacy doc.path fallback). Factor one helper and use it at both sites — three copies is exactly how the next kipy break gets fixed in one place and missed in two. Relatedly, the legacy hasattr(doc, 'path') check now sits after an unconditional break and reads as dead code; restructure so the fallback is explicit.

**Third:** src/command-timeout.ts promotes eight commands to the long-running bucket, including the read-only get_component_list and get_board_info. That changes a hang from a 30 s failure into a 10-minute wait on queries. Justify those two per-command or drop them.

And tests: #369 shipped a 138-line test file for a 12-line change to this same subsystem; hold this to that bar (the DocumentSpecifier helper is easily unit-testable with fakes). A CHANGELOG entry (English, under Unreleased) is also missing.

The IPC-compat work itself is wanted — #357 collides with this same territory on a stale base, and a corrected version of this PR is the better vehicle.

mixelpixx:
Checking in one more time. The blocker from the 2026-08-20 review is unchanged in the diff: `_pretty_dirs()` still ships `D:\kicad\share\kicad\footprints` as the default for KICAD_FOOTPRINTS_DIR and `D:\kicad\custom-lib` unconditionally, both from your own portable install and both meaningless on any other machine. The kipy 0.7 fixes themselves are still wanted; they agree with what #369 found independently.

If the hardcoded paths are replaced by the existing library-table resolution (python/commands/library.py already expands the KICAD*_FOOTPRINT_DIR variables and reads the fp-lib-table in `_resolve_uri`) by 2026-09-23, I will review and merge. Otherwise I will close this and carry the kipy signature fixes over myself with credit.

mixelpixx:
Closing on the date given in the 2026-09-10 reminder, since the hard-coded `D:\kicad` paths were still in the diff. The KiCad 10 fixes themselves were right and are needed, so they continue in #426 with you credited as co-author. That PR resolves footprints through the existing fp-lib-table parsing instead of fixed search paths, and reads the board file from the kipy Board the API is bound to. I verified both breakages you found: `GetGlobalFootprintLib` is gone from the KiCad 10.0.5 pcbnew module, and kicad-python 0.7.1 requires a document type in `get_open_documents`. Not carried over: the extra entries in the long-running command list (a timeout policy change that needs its own case), the `ping()` to `get_version()` swap (not reproduced here), and the componentId-as-footprint fallback. If you can reproduce the ping timeout, an issue with the kipy version and the log would be welcome. Thank you for finding this.

PR #390, opened 2026-08-21
fix: restart stalled Python worker

siddolo:
Restarts the Python worker after timeouts or unexpected exits, then resumes queued requests after READY.

Tests: `npm run test:ts`, `npm run build`.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -264,6 +264,7 @@ export class KiCADMcpServer {
 
   private resetReadyState(): void {
     this.startupBuffer = "";
+    this.responseBuffer = "";
     this.readyDetected = false;
     this.readyPromise = new Promise((resolve, reject) => {
       this.resolveReady = resolve;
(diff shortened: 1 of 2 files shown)


mixelpixx:
Thanks for this; it is the right shape, and the timeout test is the part that made review fast. Rather than ask you to rebase, I carried your two commits to main unchanged in #415 and added on top what the review asked for: a guard so a worker that dies before its first READY is not respawned by the exit handler (on the initial start that spawned a second worker after start() had already failed), a budget of three automatic restarts in five minutes so a broken install does not respawn forever, exit-path tests next to your timeout test, and the CHANGELOG entry. When #415 merges, GitHub marks this PR merged and you keep authorship of both commits. If you would rather land it under this PR yourself, say so and I will hold off.

siddolo:
Thanks for carrying this forward and adding the guards and tests! Happy to have it land via #415.

mixelpixx:
This landed today as part of #415, with your two commits unchanged, and ships in v2.8.0. The additions on top were the startup-race guard, a limit of three restarts in five minutes, and exit-path tests next to your timeout test. Thank you.

PR #391, opened 2026-08-23
fix(schematic): make placement, labelling and dragging match eeschema

komar3456:
Five defects found while drawing a two-sheet nRF52/FPGA design through the schematic
tools, on KiCad 10.0.4 with the SWIG backend. Each one produces a file KiCad loads
but a person then has to repair by hand, and none of them announces itself — the
tool reports success every time.

## What changes

**Placed symbols lost the library's field visibility.** `Reference` and `Value` were
written visible unconditionally, ignoring the `(hide yes)` the library symbol carries.
Power symbols hide `Reference` by convention — a `#PWR101` designator tells a reader
nothing — so every ground and rail symbol placed through the API printed one. A sheet
with 26 grounds came out with 26 stray designators over the wiring. Visibility now
comes from the injected `lib_symbols` definition, alongside the position and effects
that were already inherited. The marker is matched as a token, so a field whose
*value* contains the word "hide" is not mistaken for a hidden field.

**Net labels snapped to a pin faced the wrong way.** KiCad pairs label angle 0/90 with
`justify left` and 180/270 with `justify right`, so a label anchored at a pin endpoint
has to run along the pin's outward direction or its text lies across the symbol body.
`add_schematic_net_label` defaulted to angle 0 regardless of the pin, and
`batch_connect` inverted the pin bearing outright:

```python
orientation = {0: 180, 90: 270, 180: 0, 270: 90}.get(cardinal, 0)
```

which put the net name over the body on every right-hand pin of every part. Since
`justify` is not separately settable through the API, callers could not repair it
without editing the file. Both entry points now orient the label along the pin's
outward bearing, snapped to the four orientations KiCad allows; an explicit
`orientation` still wins. The angle/justify pairing was checked against the
schematics shipped in `share/kicad/demos` (0→left, 180→right, 90→left, 270→right,
no counter-examples).

**Dragging a component broke its routing and left its no-connect flags behind.**
Moving or rotating a symbol updated only the wire endpoint touching each pin. A trace
routed pin → corner → corner → pin is a chain of separate two-point segments, so the
first segment came out diagonal and its bend landed off-grid — reported afterwards as
`endpoint_off_grid` on a trace nobody touched. A no-connect flag on a moved pin stayed
at the old coordinate, becoming a dangling flag plus an unconnected pin.

`move_schematic_component` and `rotate_schematic_component` now carry the neighbouring
bend along with the pin, cascading through the chain until the routing settles, and
move no-connect flags with the pins they mark. A corner is left alone — and the
segment left diagonal — when an explicit junction, a fork of three or more segments,
or another component's pin holds it: moving those would drag unrelated wiring or tear
a connection apart. The response counts both outcomes (`wiresStraightened`,
`wiresLeftDiagonal`, `noConnectsMoved`), and `straightenWires: false` opts out.

**A single unit of a multi-unit part could not be addressed.** Every unit is placed as
its own `(symbol ...)` block under one shared reference, so `move_schematic_component`,
`rotate_schematic_component` and `delete_schematic_component` acted on whichever block
came first in the file — not predictable from the outside — and `delete` took the whole
part. All three now accept an optional `unit`. Naming a unit that is not placed reports
which ones are, rather than failing as "not found". Wire dragging follows: pins are
filtered to the moved unit, and the part's other units count as stationary, so their
wiring stays put.

**`batch_add_components` accepts `unit`.** `add_schematic_component` had it, the batch
did not, so a five-unit FPGA had to be placed one call per unit — the round-trips the
batch exists to avoid.

## Testing

- Full suite: 2190 passed, 36 skipped. The one failure,
  `test_ipc_open_board_path::test_returns_full_path_composed_from_project_and_filename`,
  is a pre-existing Windows path-separator assertion (`/` vs `\`), unrelated to these
  changes and failing on `main` as well.
- 28 new tests in four files, plus two added to `test_schematic_batch.py`. One existing
  assertion changed: `test_places_label` encoded the inverted orientation, and now
  encodes the pin-following one, with the reasoning in a comment.
- `vitest` 73 passed · `tsc --noEmit`, `eslint`, `prettier`, `black`, `isort`,
  `flake8`, `mypy` all clean.
- End-to-end against the built server over stdio (`node dist/index.js`, `KICAD_PYTHON`
  pointed at KiCad 10.0.4's Python): 15 checks covering every fix above — schema
  parameters present, `(hide yes)` written for a placed `power:GND`, label written at
  angle 90 with `justify left` from a pin whose outward bearing is 90°, both FPGA
  units placed by one batch call, move/delete touching only the named unit, and a
  dragged L-route left with no diagonal segment and its no-connect flag on the pin.
- `kicad-cli sch erc` on the dragged schematic reports no `endpoint_off_grid`. The
  remaining messages come from the test fixture itself (a wire deliberately drawn to
  nowhere) and are identical on an equivalent pre-move file.

## Notes

`get_schematic_pin_locations` already reports `unit` and `sexpr_format` already
escapes quoted values, so the two worst problems from the same session — a `power:*`
`Description` containing quotes splitting the S-expression, and multi-unit pin
coordinates computed for one placement — are fixed on `main` and are not part of this
PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,64 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- **Placed symbols lost the library's field visibility.** `Reference` and
+  `Value` were written visible unconditionally, ignoring the `(hide yes)` the
+  library symbol carries. Power symbols hide `Reference` by convention — a
+  `#PWR101` designator tells a reader nothing — so every ground and rail symbol
+  placed through the API printed one. A sheet with 26 grounds came out with 26
+  stray designators over the wiring, and the only way back was hand-editing the
+  file. Visibility is now read from the injected `lib_symbols` definition
+  alongside the position and effects that were already inherited. The marker is
+  matched as a token, so a field whose *value* contains the word "hide" is not
+  mistaken for a hidden field.
+
+- **Net labels snapped to a pin faced the wrong way.** KiCad pairs label angle
+  0/90 with `justify left` and 180/270 with `justify right`; a label anchored at
+  a pin endpoint has to run along the pin's outward direction or its text lies
+  across the symbol body. `add_schematic_net_label` defaulted to angle 0
+  regardless of the pin, and `batch_connect` went further and turned the label
+  around (a right-facing pin got 180), which put the net name over the body on
+  every right-hand pin of every part. Since `justify` is not separately
+  settable through the API, callers could not repair it without editing the
+  file. Both now orient the label along the pin's outward bearing, snapped to
+  the four orientations KiCad allows. An explicit `orientation` still wins.
+
+- **Dragging a component broke its routing and left its no-connect flags
+  behind.** Moving or rotating a symbol updated only the wire endpoint touching
+  each pin. A trace routed pin → corner → corner → pin is a chain of separate
+  two-point segments, so the first segment came out diagonal and its bend landed
+  off-grid — reported afterwards as `endpoint_off_grid` on a trace nobody
+  touched. A no-connect flag on a moved pin stayed at the old coordinate,
+  turning into a dangling flag plus an unconnected pin.
+
+  `move_schematic_component` and `rotate_schematic_component` now carry the
+  neighbouring bend along with the pin, cascading through the chain until the
+  routing settles, and move no-connect flags with the pins they mark. A corner
+  is left alone — and the segment left diagonal — when an explicit junction, a
+  fork of three or more segments, or another component's pin holds it: moving
+  those would drag unrelated wiring or tear a connection apart. The response
+  counts both outcomes (`wiresStraightened`, `wiresLeftDiagonal`,
+  `noConnectsMoved`); `straightenWires: false` opts out.
+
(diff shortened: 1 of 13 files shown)


mixelpixx:
Merged, thank you. The five fixes check out against KiCad's behaviour: `pin_locator.get_pin_angle` already returns the outward direction, so the {0:180, 90:270, 180:0, 270:90} remap in batch_connect was the only thing turning labels into the body, and removing it is the whole fix. The merged dynamic_symbol_loader.py keeps the angle/mirrorY handling and the KiCad-10 token gate that landed in #410 after your branch point, with no double handling, and `_straighten_bent_wires` is bounded (eight passes over a finite list) and leaves anchored corners alone, which is the right conservative choice.

Two things changed on the way in: the CHANGELOG conflict (both sides kept, your entries under Bug Fixes and New Features, since the project uses New Features rather than Improvements) and one prettier complaint, `*value*` to `_value_` in your first entry, which was the only reason the Code Quality job was red. Verified on the merged tree: vitest 84/84, the full Python suite with the same four environmental failures as main, pre-commit clean; CI on the merge commit is the final word.

One follow-up you may want, not blocking: `_straighten_bent_wires` does not treat bus, bus_entry or sheet_pin endpoints as anchors, so a corner touching one of those could be moved. A test for that case would settle it either way.

PR #392, opened 2026-08-24
Name five schema parameters the way the handlers read them

karpovantonme:
Five tools declare a parameter under a name the handler never reads, so a client that follows the schema gets an error back.

`set_active_layer` is the clearest one. The schema requires `layerName`; `commands/board/layers.py:147` does:

```python
layer = params.get("layer")
if not layer:
    return {"success": False, "message": "No layer specified",
            "errorDetails": "layer parameter is required"}
```

So the caller sends exactly what the schema asks for and is told that `layer` is required.

| tool | schema required | handler reads |
|---|---|---|
| `set_active_layer` | `layerName` | `layer` (`board/layers.py:147`) |
| `add_net` | `netName`, `netClass` | `name`, `class` (`routing.py:221`) |
| `add_copper_pour` | `netName` | `net` (`routing.py:1615`) |
| `export_gerber` | `outputPath` | `outputDir` (`export.py:25`) |
| `delete_trace` | `uuid` | `traceUuid` (`routing.py:731`) |

This renames the schema side rather than the handler side, because the rest of the server already agrees with the handlers: across the 86 tools whose schema and handler do match, the names in use are `layer` (7 tools), `position` (7), `net` (5), `reference` (9). `layerName` and `netName` appear only in the tools that are broken.

`export_gerber` is worth a second look on its own: it is the tool that produces the fabrication output, and it currently cannot run.

There are eleven more tools with the same class of mismatch, but those need the argument shape changed rather than a rename (the handler takes `position` as an object where the schema promises `x` and `y`). I left them out of this PR since that is a design call. Say the word and I will send them the same way.

### DIFF
--- a/python/schemas/tool_schemas.py
+++ b/python/schemas/tool_schemas.py
@@ -457,12 +457,12 @@ BOARD_TOOLS = [
         "inputSchema": {
             "type": "object",
             "properties": {
-                "layerName": {
+                "layer": {
                     "type": "string",
                     "description": "Name of the layer to make active (e.g., F.Cu, B.Cu, Edge.Cuts)",
                 }
             },
-            "required": ["layerName"],
+            "required": ["layer"],
         },
     },
     {
@@ -1373,17 +1373,17 @@ ROUTING_TOOLS = [
         "inputSchema": {
             "type": "object",
             "properties": {
-                "netName": {
+                "name": {
                     "type": "string",
                     "description": "Name of the net (e.g., VCC, GND, SDA)",
                     "minLength": 1,
                 },
-                "netClass": {
+                "class": {
                     "type": "string",
                     "description": "Optional net class to assign (must exist first)",
                 },
             },
-            "required": ["netName"],
+            "required": ["name"],
         },
     },
     {
@@ -1453,7 +1453,7 @@ ROUTING_TOOLS = [
         "inputSchema": {
             "type": "object",
             "properties": {
-                "uuid": {
+                "traceUuid": {
                     "type": "string",
                     "description": "UUID of a specific trace to delete",
                 },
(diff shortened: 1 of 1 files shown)


mixelpixx:
Thanks for this; the five mismatches are real, and worth a note on which layer they live in, because it changes one of the renames.

There are two schema layers. The zod schemas in `src/tools/*.ts` are what a Node-fronted MCP client sees; `python/schemas/tool_schemas.py` (the file this PR edits) is what the standalone JSON-RPC path's `tools/list` returns. For four of the five tools the zod layer already agrees with the handler (`layer`, `traceUuid`, `net`, `outputDir`), so your renames bring the JSON-RPC layer into line with both, and those are right as they are.

`add_net` is the exception. Its zod schema declares `netClass`, and the handler reading `class` was the live bug on the Node path (the class was silently dropped). #408 fixes that by making the handler read `netClass` and keep `class` as a fallback, so the zod name is now the canonical one. Please drop the `netClass -> class` hunk and keep `netName -> name`; with that change this is mergeable, and I will merge it in the next batch. It would also be worth one sentence in the description saying the file is the JSON-RPC schema, so a reader does not take it for the schema Node clients see.

Context: #408 also adds a ratchet test that compares every tool's declared parameter names against the keys its handler reads, on both layers. It found 16 stale JSON-RPC parameters in total, tracked in #407; this PR clears five of them, so after it merges the corresponding entries come out of that test's known list (the test fails on stale entries, so whoever merges second handles that).

mixelpixx:
Merged, thank you, with the one adjustment from my earlier comment: the `add_net` `netClass -> class` hunk was reverted, because the zod schema (what Node clients see) declares `netClass` and since #408 the handler reads `netClass` with `class` as a fallback, so `netClass` is the canonical name on both layers. The other four renames are in as you wrote them, and their entries came out of the known list in `tests/test_schema_param_names.py` (16 -> 11 stale JSON-RPC parameters remain; #407 tracks them if you want more of the same). CHANGELOG entry added on main with credit.

PR #393, opened 2026-08-26
fix(schematic): guard kicad-skip imports so one missing package can't kill all 216 tools

### LINKED ISSUE #389: One missing optional dependency (kicad-skip) kills every tool at import
Root-cause pattern behind #361 (and half of #264): six modules import the kicad-skip package unguarded at module top level, and kicad_interface.py imports them all at startup — so a single absent pip package (typically installed into the wrong interpreter) takes down the entire backend with 'No module named skip' before the MCP handshake.

The six sites:
- python/commands/library_schematic.py:8
- python/commands/component_schematic.py:6
- python/commands/connection_schematic.py:6
- python/commands/pin_locator.py:18
- python/commands/schematic.py:9
- python/commands/schematic_analysis.py:18

Fix shape: guard the import (try/except ImportError with a module flag), and have the affected command handlers return a structured error naming the missing package and the exact pip command for the configured interpreter — the same self-explaining style #370 brought to setup-macos --verify. The server then starts with every tool that does not need kicad-skip working, and the ones that do fail with an actionable message instead of killing the process.

Setup-side prevention is separate and already improving (#370 for macOS; setup-windows installs via the detected interpreter). This issue is about the server failing better when prevention was skipped.

AmirF194:
Six modules import `from skip import Schematic` at their own top level. Two of them, `library_schematic.py` and `schematic.py`, sit on the import chain `kicad_interface` -> `schematic_handlers` -> `library_schematic`/`schematic`, which runs at `kicad_interface.py:39`, before the file's own try/except import guard at line 335 is even reached. So a missing kicad-skip did not hit the intended `sys.exit(1)` with a clean JSON error at all: it raised an unhandled `ModuleNotFoundError` straight out of the module-level import, taking down every one of the 216 tools including the 210 that never touch kicad-skip.

Invariant: importing `kicad_interface` (or any of its command modules) must not require kicad-skip to be installed. Only the handful of tools that actually call into it may fail, and only with a structured, actionable error.

Fix, by module:

- `library_schematic.py`, `schematic_analysis.py`: `Schematic` was imported but never referenced anywhere in the file. Removed.
- `pin_locator.py`: same, never referenced in production code, but the existing test suite patches `commands.pin_locator.Schematic` as a seam, so the import is guarded (falls back to `None`) instead of removed.
- `component_schematic.py`, `connection_schematic.py`: `Schematic` is used only as a parameter type hint, never instantiated. Guarded with a fallback to `typing.Any`, matching this file's own existing `WIRE_MANAGER_AVAILABLE` pattern for `PinLocator`/`WireManager`.
- `schematic.py`: the one module that actually calls `Schematic(...)`. `create_schematic` and `load_schematic` now raise the existing `SchematicLoadError` with a new `dependency_missing` kind, naming the exact `pip install kicad-skip` command, instead of failing at import time.

Verification (clean container, kicad-skip genuinely absent via `sys.modules["skip"] = None`, checked on Python 3.9 and 3.12):
- New test file added: 4 tests fail on main with the unhandled `ModuleNotFoundError` above, pass on this branch. Covers the full `kicad_interface` import, the actionable error from `SchematicManager`, and that a tool needing no kicad-skip still works.
- Full existing suite: 2154 passed, 47 skipped (unchanged from main; skips are all pre-existing environment gates like `kicad-cli`/real `pcbnew`).
- `black --check` and `mypy` both clean.
- Not verified: real kicad-skip behavior end to end (this fix only concerns what happens when the package is absent).

Fixes #389.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- **A missing kicad-skip no longer kills all 216 tools at startup** (#389).
+  Six modules imported `from skip import Schematic` at their own top level.
+  Two of them sit on the import chain `kicad_interface` -> `schematic_handlers`
+  -> `library_schematic`/`schematic`, which runs before the process reaches
+  its own try/except import guard, so a missing kicad-skip raised an
+  unhandled `ModuleNotFoundError` at startup instead of the intended clean
+  JSON error. Three of the six modules never actually used the import and
+  had it removed; the other three guard it now. `create_schematic` and
+  `load_schematic`, the two tools that genuinely need kicad-skip, raise a
+  `SchematicLoadError` naming the exact `pip install kicad-skip` command
+  instead of taking the rest of the server down with them.
+
 ## [2.7.0] - 2026-08-20
 
 ### New Tools
--- a/python/commands/component_schematic.py
+++ b/python/commands/component_schematic.py
@@ -3,7 +3,11 @@ import uuid
 from pathlib import Path
 from typing import Any, List, Optional
 
-from skip import Schematic
+try:
+    from skip import Schematic
+except ImportError:
+    # kicad-skip is optional; fall back to Any so these type hints stay valid.
+    Schematic = Any
 
 logger = logging.getLogger(__name__)
 
--- a/python/commands/connection_schematic.py
+++ b/python/commands/connection_schematic.py
@@ -3,7 +3,11 @@ import os
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
-from skip import Schematic
+try:
+    from skip import Schematic
+except ImportError:
+    # kicad-skip is optional; fall back to Any so these type hints stay valid.
+    Schematic = Any
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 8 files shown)


AmirF194:
Checking back in after a week. Right now one missing optional package (kicad-skip) throws at import time and takes down all 216 tools with it; this scopes the import failure to just the schematic tools that actually need it. Should be a quick look.

mixelpixx:
Merged, thank you; this closes #389. The crash chain you traced (`kicad_interface` -> `schematic_handlers` -> `library_schematic`/`schematic`, all before the guard at line 335) checks out, and subprocess tests with `sys.modules["skip"] = None` are the right way to prove it, since the conftest stub would otherwise hide the failure.

One wording change on merge: the CHANGELOG line said "all 216 tools". The server registers 233 (173 of them indexed by `search_tools`), so it now says "every tool"; the number in #389's title was my own miscount.

Two lazy imports remain outside the invariant: `schematic_handlers.py` inside `_build_hierarchical_pad_net_map`, and `wire_connectivity.py` inside `get_connections_for_net`. They cannot crash startup, but with kicad-skip missing they surface a raw traceback instead of the `SchematicLoadError` message. I will fold those into the `sync_schematic_to_board` fix that is next in the queue, since one of them lives in that function.

AmirF194:
Good catch on the tool count, I copied 216 from the issue title without rechecking it against search_tools. Makes sense to fold the two lazy-import spots into the sync_schematic_to_board fix since one already lives in that function.

PR #394, opened 2026-08-27
perf: batch net connectivity per sheet — list_schematic_nets O(nets × sheet) → one pass

markszente:
## Problem

`list_schematic_nets` recomputes connectivity from scratch for every net in the schematic. On a flat 119-component / 44-net schematic it ran for **90+ seconds** (and was cancelled by the user before finishing).

Profiling one `get_connections_for_net('GND')` call (2.1 s total) showed where it goes:

- `sexpdata.loads` runs **23 times per net** — `_load_sexp`'s docstring says "Load and cache" but there was no cache, and every per-net call re-parsed the same 384 KB `.kicad_sch`
- `pin_locator.get_symbol_pins` runs 155× per net, re-locating all pins for all components — every call site constructs a fresh `PinLocator()`, so its instance caches die between nets
- the handler loop then repeats all of this once per net; `generate_netlist` in `connection_schematic.py` has the identical loop

## Fix

- **`_load_sexp` now actually caches** the parsed tree, keyed by `(mtime_ns, size)` so on-disk edits still invalidate. `_discover_sub_sheets` and `_parse_hierarchical_labels_sexp` now go through it too.
- **New `get_connections_for_nets(schematic, path, net_names, locator=None)`** resolves a batch of nets with one pass per sheet: sexp parse, wire graph, virtual connections, and component pin world positions (`_build_sheet_pin_index`) are computed once and shared; each net then costs only its BFS + a membership check against the pin index. Sub-sheets are also loaded once per batch instead of once per net. `get_connections_for_net` / `_process_single_sheet` remain as single-net wrappers, so existing callers and behavior (multi-unit filtering per #293, ±1 IU tolerance, `SchematicLoadError` propagation) are unchanged.
- **`count_pins_on_net` accepts an optional shared `locator`**, and **`PinLocator.get_all_symbol_pins` memoises per `(schematic, reference)`** — it previously re-found the symbol and re-ran the per-pin transform on every call (44 nets × 119 symbols in the handler loop).
- `_handle_list_schematic_nets` and `generate_netlist` use the batched API with one shared locator. Netlist net order is now sorted (previously set-iteration order).

## Results

On the 119-component / 44-net schematic (KiCad bundled Python 3.9):

| | before | after |
|---|---|---|
| per-net connections loop | 62.1 s | 1.31 s (**47×**) |
| full `list_schematic_nets` handler path | ~75 s | **2.81 s** |

A/B equivalence check against the pre-fix implementation on the same schematic: all 44 nets return **identical** `(component, pin)` sets, and pin counts are unchanged.

## Testing

- Full Python suite: **2159 passed**, 38 skipped (environmental: missing 7-Zip/kicad-cli/Linux library paths)
- A/B comparison script (old module loaded side-by-side) confirming identical output on a real 384 KB production schematic

https://claude.ai/code/session_01PvUVwFTAdx3z4gW7jnY5w9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`list_schematic_nets` and `generate_netlist` no longer scale as
+  O(nets × sheet size)** — net connectivity was recomputed from scratch for
+  every net in the list: each per-net query re-parsed the schematic
+  s-expression tree (dozens of `sexpdata.loads` calls per net), rebuilt the
+  wire adjacency graph, and re-located every component pin through a fresh
+  `PinLocator`, whose caches died with it. On a flat 119-component, 44-net
+  schematic, `list_schematic_nets` took over 90 seconds. The per-sheet work is
+  now done once per request and shared across all nets
+  (`get_connections_for_nets`), `_load_sexp` actually caches the parsed tree
+  (keyed by file mtime and size, as its docstring always claimed), and
+  `PinLocator.get_all_symbol_pins` memoises per (schematic, reference). The
+  same schematic now lists in under 3 seconds, with identical results —
+  `get_connections_for_net` remains as a single-net wrapper for existing
+  callers.
+
 ## [2.7.0] - 2026-08-20
 
 ### New Tools
--- a/python/commands/connection_schematic.py
+++ b/python/commands/connection_schematic.py
@@ -420,7 +420,7 @@ class ConnectionManager:
             }
         """
         try:
-            from commands.wire_connectivity import get_connections_for_net
+            from commands.wire_connectivity import get_connections_for_nets
 
             netlist: Dict[str, Any] = {"nets": [], "components": []}
 
@@ -449,8 +449,12 @@ class ConnectionManager:
                             net_names.add(label.value)
 
             sch_path_str = str(schematic_path) if schematic_path else ""
-            for net_name in net_names:
-                connections = get_connections_for_net(schematic, sch_path_str, net_name)
+            # Batched: every sheet is parsed and pin-located once for the whole
+            # net list, instead of once per net.
+            connections_by_net = get_connections_for_nets(
+                schematic, sch_path_str, sorted(net_names)
+            )
+            for net_name, connections in connections_by_net.items():
                 if connections:
                     netlist["nets"].append({"name": net_name, "connections": connections})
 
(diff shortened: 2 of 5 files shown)


mixelpixx:
Thanks; the profile is convincing and the hoist into `_build_sheet_pin_index` is a line-for-line copy of the existing pin-transform logic, so I want this in. Two things before merge.

1. A committed equivalence test. The A/B run in the description is exactly what belongs in the repo: the same net list through `get_connections_for_net` (per net) and `get_connections_for_nets` (batched), on a fixture with a hierarchical sub-sheet, a multi-unit symbol and a power symbol, asserting identical output. Without it the next change to `_build_sheet_pin_index` has nothing to fail against, and the PR currently adds no test at all.

2. `_SEXP_CACHE` is a module-level global in a Python process that lives for the whole session, keyed by `(mtime_ns, size)`, and the schematic writers in this repo rewrite `.kicad_sch` in place (`open(path, "w")`, no temp-and-rename). A coordinate edit that leaves the file size unchanged and lands inside the filesystem's timestamp granularity serves the previous parse to the next `list_schematic_nets` call, silently. Either scope the cache to a request (the way `PinLocator` already scopes its caches to an instance, which the `locator=` parameter suggests anyway) or invalidate it from the write path. A test that rewrites the file with a forced-identical `os.stat` result and asserts the second read is fresh would pin whichever you choose.

One merge-order note: #393 touches the top of `pin_locator.py` and `connection_schematic.py` (import guards) and is merging first; yours touches function bodies in the same files, so expect a small rebase once it lands.

mixelpixx:
One more thing, now that the fork CI run has been approved and finished: it is red. Code Quality fails because black would reformat `python/commands/wire_connectivity.py`, and the Python Tests (ubuntu-24.04, 3.12) job exits 1; the run is https://github.com/mixelpixx/KiCAD-MCP-Server/actions/runs/33078628428. Running `pre-commit run --all-files` and the full `pytest tests/` locally before the next push will show both. Not a judgement on the change itself, just what has to be green before it can merge.

mixelpixx:
Two updates while this waits on the equivalence test and the cache scoping.

1. #409 merged on 2026-09-06 and removed the unused `from skip import Schematic as SkipSchematic` import inside `get_connections_for_net` in python/commands/wire_connectivity.py, the function this PR hoists, so the branch now conflicts there as well as in CHANGELOG.md. The resolution is to drop that import on your side too.

2. To keep the queue moving: if the test and the cache change are not up by 2026-09-16, I will rebase the hoist onto main myself, add the test, scope the cache per request, and land it with you as co-author. The profile you posted is convincing and I want the speed-up in the next release either way.

mixelpixx:
Closing in favour of #429, which carries your commit unchanged in authorship (cherry-picked onto current main, with you as co-author on the follow-up). The 2026-09-16 date from the last update passed without the requested changes, and the speed-up is too good to leave sitting. The follow-up commit does what the reviews asked. The parse cache is now scoped to one call instead of the whole process: a process-wide cache handed shared trees to every caller in the module and could miss an edit that kept the size and mtime, and a test in #429 shows that case. The pin memo is opt-in, because ConnectionManager keeps one PinLocator for the life of the worker and an always-on memo there would never refresh. It also adds the equivalence tests. On KiCad's own hierarchical demos, your batched path returns exactly the same pins for every net as main's per-net code, about ten times faster. Thank you for the profiling; it made the review easy.

PR #395, opened 2026-08-28
refactor: share KiCad text file helpers

### LINKED ISSUE #380: Consolidate the four private _read_text/_write_text copies into python/utils
The 2026-08-06 batch (#362, #364, #365, #366) each shipped a private, near-identical pair of _read_text/_write_text helpers implementing newline-preserving reads and atomic os.replace writes. Byte-for-byte the logic is the same; four copies mean the next fix (for example, preserving BOM, or fsync-before-replace) has to be applied four times or silently diverges.

Move one canonical pair into python/utils (sexpr_format.py already hosts the shared S-expression helpers; a small fileio module is also fine) and point the four call sites at it. Pure refactor, no behavior change intended — the existing per-module tests already pin the atomicity and newline behavior, so they double as the regression net.

Related: #362 also defines private _match_paren/_iter_children that duplicate the shared match_paren/iter_child_offsets in sexpr_format.py (landed via #365) — same consolidation, same PR.

kkkhs:
## What
Fixes #380. Consolidates the duplicated KiCad text read/write helpers used by the Python file-editing tools.

## Fix
Adds `utils.file_io` for newline-preserving reads and atomic writes, then points the four private helper copies at it while keeping the shared paren walkers in `utils.sexpr_format`.

## Test
No new cases; existing add-symbol-property, backannotation, library-table, and pin-type regression tests cover the moved helpers.

### DIFF
--- a/python/commands/add_symbol_property.py
+++ b/python/commands/add_symbol_property.py
@@ -2,13 +2,15 @@
 
 from __future__ import annotations
 
-import os
 import re
 from collections.abc import Iterator, Sequence
 from pathlib import Path
 from typing import Any
 
-from utils.sexpr_format import escape_sexpr_string
+from utils.file_io import read_text_preserve_newline as _read_text
+from utils.file_io import write_text_atomic as _write_text
+from utils.sexpr_format import escape_sexpr_string, iter_child_offsets
+from utils.sexpr_format import match_paren as _match_paren
 
 # A property belongs to the symbol itself, so it has to sit at the same nesting
 # level as (property "Reference" ...). Inside the parent block that is depth 2:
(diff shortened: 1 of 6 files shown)


mixelpixx:
Merged as-is, thank you; this closes #380. It was still marked draft, but nothing in it looked unfinished: all four copies are consolidated, the moves onto the shared `match_paren` / `iter_child_offsets` are correct, and the monkeypatch target change in `test_add_symbol_property` is exactly what dropping the local `os` import requires.

One behaviour change worth stating, since the description did not: two of the four sites (`add_symbol_property`, `backannotate_footprints`) now clean up the `.mcp-tmp` file on a failed write, which the other two already did. That is an upgrade, so it went in unchanged. CHANGELOG entry added on main with credit.

PR #396, opened 2026-09-01
docs: correct documentation to match v2.7.0, and gate the generated tool inventory in CI

scorp508:
## Summary

The documentation had drifted well out of step with what the server actually
ships. `docs/TOOL_INVENTORY.md` listed 138 tools while the server registers 229,
so roughly 40 percent of the callable surface was undocumented. Several other
documents described a tool-gating design that was removed over a year ago, and
two documented tools do not exist at all.

This PR makes the docs match the code, and adds a generator plus a CI gate so
the inventory cannot silently drift again.

Two commits:

1. `docs: add tool inventory generator and regenerate TOOL_INVENTORY.md`
2. `docs: correct documentation to match v2.7.0`

## Why a generator

`docs/TOOL_INVENTORY.md` is now produced by `scripts/generate-tool-inventory.mjs`,
which derives the list from the two authoritative sources rather than from hand
maintenance:

- `src/tools/*.ts` for the tools that are actually registered and callable
- `dist/tools/registry.js` for the metadata `search_tools` exposes

It preserves existing hand-written descriptions so curated prose survives
regeneration, and emits any tool file it does not recognise at the end of the
document rather than dropping it silently.

Two npm scripts:

- `npm run docs:tools` regenerates the document
- `npm run docs:tools:check` verifies it is current and exits 1 on drift

## Corrections in the second commit

- **Tool discovery docs.** `ROUTER_QUICK_START.md` described the router as a gate
  that hides tools behind `execute_tool`. That design was disabled in `3d9497e`
  (it made the model hallucinate tool schemas) and `execute_tool` was deleted in
  `963a39c`. All 229 tools are registered directly and visible; what remains is a
  search catalogue over 169 of them in 15 categories. The document is rewritten
  around that reality, and its category table is taken from `getRegistryStats()`.

- **Tools that do not exist.** `SCHEMATIC_TOOLS_REFERENCE.md` documented `add_wire`
  and `add_schematic_connection`. Neither appears anywhere in `src/`. The correct
  tool is `add_schematic_wire`. The same document carried a stale tool count and a
  hardcoded `/home/chris/MCP/...` path.

- **README figures.** The architecture section advertised `Tools (122),
  Resources (8)` and 14 categories. The actual figures are 229 tools, 23 resources
  (`src/resources/`) and 15 categories.

- **v2.7.0 library tooling.** `LIBRARY_INTEGRATION.md` did not mention the
  library-table maintenance or symbol validation tools added in v2.7.0.

- **Per-process logging.** The logging docs still pointed at a single
  `kicad_interface.log`. v2.7.0 writes one file per server process,
  `kicad_interface-<pid>.log`. `python/kicad_interface.py` and `setup-windows.ps1`
  now say so, and the setup script reports the most recently written file.

## CI

Adds `npm run docs:tools:check` to the existing `typescript-tests` job, after
`npm run build` (the check reads `dist/`). The pre-commit hook only protects
contributors who have installed it, so the inventory was not actually gated.

Worth noting: `tests-ts/readme-counts.test.ts` already pins the README headline to
the registry, and that line stayed correct. The unpinned architecture bullet about
1,500 lines below it is what drifted to 122/8/14. That is the argument for the
generated-plus-gated approach.

`.gitattributes` pins `*.md text eol=lf`, so the byte comparison is stable on the
Windows runners in the matrix.

## Testing

- `npm run docs:tools:check` passes
- `npx vitest run` passes: 73 tests across 8 files
- `pre-commit` passes on all changed files (black, isort, prettier, flake8, mypy, eslint)
- `ci.yml` validated as parseable YAML and step ordering confirmed

## Checklist

- [x] All tests pass locally
- [x] Documentation updated
- [x] No merge conflicts

## Companion PRs: #397 and #398

The follow-up noted below is now open as #397
(`fix: stop discovery tools telling clients to call deleted execute_tool`).

A third, unrelated fix is open as #398 (`fix: stop PCB_VIA::GetWidth assert
dialog, without breaking KiCad 8`).

`src/tools/router.ts` and `src/tools/registry.ts` still tell the model to use
`execute_tool` at runtime, in the `get_category_tools` and `search_tools`
responses, even though that tool no longer exists. A single `search_tools` reply
can currently contain both "call directly, no execute_tool needed" and "use
execute_tool with the tool name to run it".

That is a behaviour change rather than a docs change, so it is kept out of this
PR. The two are **independent**:

- This PR is `docs:` and touches `docs/`, `README.md`, other `*.md` files and
  `.github/workflows/ci.yml`. It changes no runtime behaviour.
- #397 is `fix:` and is confined to `src/` and `tests-ts/` (TypeScript).
- #398 is `fix:` and is confined to `python/commands/` and `tests/` (Python).

conflict and no ordering requirement. Any of them can merge first. I verified
this with a real `git merge-tree` simulation between the branches, which
reported no conflict. All three were cut from `aa53d52`
(`chore: release v2.7.0`).


mixelpixx:
Merged, thank you. Resolution notes, since several PRs landed on main between your base and the merge: README counts were re-derived to 233 registered / 173 indexed / 16 categories (`set_net_color` from #376 and the three Digi-Key tools from #368 arrived after your count), the Routing section keeps both your three additions and `set_net_color`, and `docs/TOOL_INVENTORY.md` was regenerated with your script after the merge so `docs:tools:check` passes on main's first run. All five factual claims I spot-checked were accurate, and the CI step is additive inside the existing TypeScript job, so fork-PR gating is unchanged.

The generator has no unit test of its own. Not blocking, but if you add one later, description preservation and the unknown-file fallback are the two behaviours worth pinning, because they are what stop a regeneration from silently dropping prose.

PR #397, opened 2026-09-01
fix: stop discovery tools telling clients to call deleted execute_tool

scorp508:
## Summary

The three discovery tools (`list_tool_categories`, `get_category_tools`,
`search_tools`) tell the client how to run what it just found. That text is read
by a model, so a stale instruction is not cosmetic: it sends the model after a
tool that does not exist.

They still instruct clients to call `execute_tool`, which is not registered on
the server:

- `c656000` added the gated router with a single `execute_tool` entry point.
- `3d9497e` disabled the gate because routing calls through one indirect entry
  point made the model hallucinate tool schemas.
- `963a39c` deleted `execute_tool` itself.

The response strings were never updated. The result is self-contradictory
within a single reply: `searchTools()` labels direct tools
"call directly, no execute_tool needed" while the router note in the same
response says to use `execute_tool` for routed ones. A model following either
half is sent to a nonexistent tool.

All 144 registry tools are callable directly by name, so the corrected text
simply says so.

## Changes

- `src/tools/router.ts` (5 edits)
  - File header now records the discovery-only status and the
    `3d9497e` -> `963a39c` history, so the next reader does not "restore" the
    removed path.
  - `list_tool_categories` description and note.
  - `get_category_tools` per-tool description and note.
  - `search_tools` note.
- `src/tools/registry.ts` (1 edit)
  - `searchTools()` direct-tool text: "call directly, no execute_tool needed"
    becomes "call directly by name".
- `tests-ts/no-stale-tool-references.test.ts` (new)

## About the test

The guard asserts on **runtime responses**, not on source text.

A source grep would flag the file header in `router.ts` that legitimately
explains this history by name, which would pressure a future author to delete
the explanation in order to make CI pass. What must stay clean is what actually
reaches a client.

The test registers the router against a small recording stand-in for
`McpServer`, invokes the real handlers, and asserts:

1. `execute_tool` is not registered anywhere.
2. Exactly the three read-only discovery tools are registered.
3. No discovery response body mentions `execute_tool`, covering the matched,
   direct-tool, and no-match branches.
4. Every tool name advertised in a response is actually registered via
   `server.tool(`.

The stand-in's command function throws if called, which also proves the
discovery tools stay read-only and never reach into KiCAD.

## Verification

- Guard confirmed non-vacuous: reverting the source fix and re-running it fails
  with `expected '{"category": "export", ...' not to contain 'execute_tool'`.
  The fix was then restored.
- `npm run build` clean.
- Full suite: 77 tests across 9 files pass (73 across 8 before this PR).
- `npm run lint:ts`: 0 errors. The 9 warnings are pre-existing and in files this
  PR does not touch.
- `pre-commit` passes on all changed files.

No behaviour change beyond the corrected text: the discovery tools remain
read-only.

## Relationship to #396 and #398

This is the follow-up named in the "Known follow-up, not in this PR" section of
#396 (`docs: correct documentation to match v2.7.0, and gate the generated tool
inventory in CI`). A third, unrelated fix is open as #398 (`fix: stop
PCB_VIA::GetWidth assert dialog, without breaking KiCad 8`).

The three are deliberately separate and **independent**:

- #396 is a `docs:` change. It touches `docs/`, `README.md`, `*.md` and
  `.github/workflows/ci.yml`, and changes no runtime behaviour.
- This PR is a `fix:` change confined to `src/` and `tests-ts/` (TypeScript).
- #398 is a `fix:` change confined to `python/commands/` and `tests/` (Python).

conflict and no ordering requirement. Any of them can merge first. I verified
this with a real `git merge-tree` simulation between the branches, which
reported no conflict.

All three branches were cut from `aa53d52` (`chore: release v2.7.0`).

### DIFF
--- a/src/tools/registry.ts
+++ b/src/tools/registry.ts
@@ -397,7 +397,7 @@ export function searchTools(query: string): SearchResult[] {
       matches.push({
         category: "direct",
         tool: toolName,
-        description: `${toolName} (direct tool — call directly, no execute_tool needed)`,
+        description: `${toolName} (direct tool — call directly by name)`,
       });
     }
   }
--- a/src/tools/router.ts
+++ b/src/tools/router.ts
@@ -1,7 +1,16 @@
 /**
- * Router Tools for KiCAD MCP Server
+ * Tool discovery for the KiCAD MCP Server.
  *
- * Provides discovery and execution of routed tools
+ * These three tools browse and search the registry in `registry.ts`. They do
+ * not execute anything on the caller's behalf. Every tool is registered
+ * directly with the server and is callable by name, so discovery is a search
+ * catalogue rather than a gate.
+ *
+ * The original gated design routed calls through a single `execute_tool`. It
+ * was disabled in 3d9497e because indirect execution made the model
+ * hallucinate tool schemas, and `execute_tool` was deleted in 963a39c. Response
+ * text here must therefore never tell a client to call `execute_tool`; see
+ * tests-ts/no-stale-tool-references.test.ts, which enforces that.
  */
 
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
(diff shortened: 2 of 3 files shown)


mixelpixx:
Merged, thank you. Confirmed the history (963a39c deleted `execute_tool` and the discovery strings were never updated) and that the new test checks runtime output rather than source text, so the historical comment survives it. CHANGELOG entry added on main with credit.

PR #398, opened 2026-09-01
fix: stop PCB_VIA::GetWidth assert dialog, without breaking KiCad 8

scorp508:
## Summary

Vias trigger a modal wxWidgets debug alert on KiCad 9 and 10:

```
pcb_track.cpp(387): assert "false" failed in GetWidth():
Warning: PCB_VIA::GetWidth called without a layer argument
```

It fires once per via, so listing traces or running `add_gnd_stitching_vias` on
a real board buries the user in dialogs.

KiCad 9 moved vias onto per-layer padstacks. `PCB_VIA` gained
`GetWidth( PCB_LAYER_ID )` and the `GetFrontWidth()` wrapper, and the inherited
no-argument `GetWidth()` was kept only to satisfy the parent class:

```
pcbnew/pcb_track.cpp  9.0:381   wxCHECK_MSG( false, m_padStack.Size( ALL_LAYERS ).x,
pcbnew/pcb_track.cpp 10.0:387     "Warning: PCB_VIA::GetWidth called without a layer argument" );
```

The check returns the ALL_LAYERS size, so **the value was never wrong**. The
defect is the dialog, not the geometry.

## Why not just call GetFrontWidth()

Because it does not exist on KiCad 8, which this repo still tests in CI
(`Integration Tests (Linux + KiCAD)` covers 8.0, 9.0 and 10.0):

```
pcbnew/pcb_track.h  8.0 : PCB_VIA only inherits  int GetWidth() const     (line 107)
pcbnew/pcb_track.h  9.0 : 430  int GetWidth() const override
                          433  int GetWidth( PCB_LAYER_ID aLayer ) const
                          437  int GetFrontWidth() const { return GetWidth( F_Cu ); }
```

Calling `GetFrontWidth()` unconditionally would trade a cosmetic warning on
9/10 for an `AttributeError` on 8. Passing `F_Cu` explicitly raises `TypeError`
there for the same reason.

That is not hypothetical: **two call sites were already broken on KiCad 8**,
both passing `pcbnew.F_Cu` (`add_via` and `copy_routing_pattern`). This PR
fixes those as well.

## Changes

`python/commands/routing.py` adds `_via_front_width()`, which probes for the
accessor rather than branching on a version string, and routes all four
via-width reads through it:

| Call site | Was | Problem |
|---|---|---|
| `get_traces_and_vias` (`"diameter"`) | `GetWidth()` | assert dialog on 9/10 |
| `add_gnd_stitching_vias` (obstacle radius) | `GetWidth()` | assert dialog on 9/10 |
| `add_via` (`"size"`) | `GetWidth(F_Cu)` | broken on KiCad 8 |
| `copy_routing_pattern` | `GetWidth(F_Cu)` | broken on KiCad 8 |

## Test double correction (please read)

`tests/test_add_gnd_stitching_vias.py` builds vias from `MagicMock`, which
auto-creates `GetFrontWidth`. `int()` of the resulting mock is `1`, so the new
shim silently returned a **1 nm** via width, every obstacle radius in that
suite collapsed to the drill size, and **all 24 tests still passed**.

That is worth flagging on its own: the suite was not sensitive to via width at
all. The double now pins `GetFrontWidth` explicitly.

`tests/test_via_front_width.py` is new and deliberately uses plain classes
instead of `MagicMock`, because attribute presence is the thing under test. The
KiCad 9/10 fake **raises** from `GetWidth()`, so taking the asserting path is a
test failure rather than an invisible dialog.

## Verification

- Reproduced the assert from the command line against KiCad 10.0.6, and
  confirmed all three accessors return the same value
  (`GetWidth(F_Cu)` = `GetFrontWidth()` = `GetWidth()` = 600000).
- Full Python suite: **2164 passed, 36 skipped**.
- The single failure, `tests/test_ipc_open_board_path.py`, reproduces unchanged
  on clean `main` (`aa53d52`) in a separate worktree. It is pre-existing and
  unrelated.
- `pre-commit` passes on all changed files (black, isort, flake8, mypy).

## Relationship to #396 and #397

Third of three independent branches cut from `aa53d52` (`chore: release
v2.7.0`):

- #396 is `docs:` and touches `docs/`, `README.md` and `.github/workflows/ci.yml`.
- #397 is `fix:` and touches `src/tools/` and `tests-ts/` (TypeScript).
- This PR is `fix:` and touches `python/commands/` and `tests/` (Python).

conflict and no ordering requirement. Any of them can merge first.

### DIFF
--- a/python/commands/routing.py
+++ b/python/commands/routing.py
@@ -702,7 +702,7 @@ class RoutingCommands:
                         "y": position["y"],
                         "unit": position["unit"],
                     },
-                    "size": via.GetWidth(pcbnew.F_Cu) / 1000000,
+                    "size": _via_front_width(via) / 1000000,
                     "drill": via.GetDrill() / 1000000,
                     "from_layer": from_layer,
                     "to_layer": to_layer,
@@ -959,7 +959,7 @@ class RoutingCommands:
                                 },
                                 "net": track.GetNetname(),
                                 "netCode": track.GetNetCode(),
-                                "diameter": track.GetWidth() / scale,
+                                "diameter": _via_front_width(track) / scale,
                                 "drill": track.GetDrillValue() / scale,
                             }
                         )
@@ -1404,7 +1404,7 @@ class RoutingCommands:
                 # Create new via
                 new_via = pcbnew.PCB_VIA(self.board)
                 new_via.SetPosition(pcbnew.VECTOR2I(pos.x + offset_x, pos.y + offset_y))
-                new_via.SetWidth(via.GetWidth(pcbnew.F_Cu))
+                new_via.SetWidth(_via_front_width(via))
                 new_via.SetDrill(via.GetDrillValue())
                 new_via.SetViaType(via.GetViaType())
 
@@ -2119,7 +2119,7 @@ class RoutingCommands:
                 is_via = False
             if is_via:
                 pos = track.GetPosition()
-                width = track.GetWidth()
+                width = _via_front_width(track)
                 drill = 0
                 try:
                     drill = track.GetDrill()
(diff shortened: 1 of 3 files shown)


mixelpixx:
Merged, thank you. Verified the four call sites against main, and the KiCad 8, 9 and 10 integration jobs were green on the PR head. The MagicMock finding (24 stitching-via tests silently running with 1 nm vias) was worth a PR on its own; that class of test-double error is exactly why the fakes in this repo are moving to plain classes. CHANGELOG entry added on main with credit.

PR #401, opened 2026-09-05
fix(process): detect KiCad launched via PATH on Linux

famez:
get_process_info() required /kicad or /pcbnew (with leading slash) in the ps line, which excluded processes whose argv[0] is the bare command name (common when launched from a desktop launcher or terminal via PATH).

Extract the base name and also accept it when it matches "kicad" or "pcbnew" exactly, avoiding false matches against our own Python scripts.

Launching kicad from UI:

Command_output

Before the patch:

Kicad_not_running

"Shows kicad not running".

After the patch:

MCP-kicad-afer

Shows "kicad running".

### DIFF
--- a/python/utils/kicad_process.py
+++ b/python/utils/kicad_process.py
@@ -291,17 +291,28 @@ class KiCADProcessManager:
                         and "kicad_interface.py" not in line
                         and "grep" not in line
                     ):
-                        # More specific check: must have /pcbnew or /kicad in the path
-                        if "/pcbnew" in line or "/kicad" in line or "KiCad.app" in line:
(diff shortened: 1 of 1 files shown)


mixelpixx:
Merged, thank you. The change is additive and confined to the Linux/Darwin branch, so Windows detection is untouched. There is no test for `get_process_info`'s `ps` parsing at all yet (the two existing references mock the method away); if you feel like a follow-up, one test feeding a synthetic `ps aux` line with a bare `kicad` argv[0] through a mocked `subprocess.run` would pin this. CHANGELOG entry added on main with credit.

PR #411, opened 2026-09-07
fix(component): exchange the physical footprint in edit_component, not just the FPID

### LINKED ISSUE #399: Missing tool: update_footprint_from_library (PCB equivalent of KiCad GUI's 'Update Footprint from Library')
## Problem

There is no MCP tool to update the **physical pad geometry** of a PCB footprint from the KiCad library — the equivalent of KiCad PCB Editor → right-click → *Update Footprint from Library*.

The existing `edit_component` tool accepts a `footprint` parameter, but it only calls `module.SetFPID(fpid)`, which updates the footprint ID string (library:name) without replacing the pads, courtyard, silkscreen, or any other graphic elements from the library. This means that after calling `edit_component` with a new footprint, the component still has the physical geometry of its original footprint.

### Observed in `python/commands/component.py` (lines 503–515)

```python
if footprint:
    if ":" in footprint:
        lib_name, fp_name = footprint.split(":", 1)
        fpid = pcbnew.LIB_ID(lib_name, fp_name)
        module.SetFPID(fpid)
    else:
        ...
        module.SetFPID(fpid)
```

Only `SetFPID` is called — no `FootprintLoad()`, no pad replacement.

### Symptoms

After `edit_component` changes e.g. `Diode_SMD:D_SOD-323` → `Package_TO_SOT_SMD:SOT-23`:
- The pad count and positions remain those of the old SOD-323 footprint
- DRC reports `lib_footprint_mismatch` (expected — ID says SOT-23 but pads are SOD-323)
- DRC also reports unconnected pads (SOT-23 has 3 pads; only 2 are present on the board)
- Freerouting routes to the old pad positions (SOD-323), not the library ones

## Expected behavior

A new tool `update_footprint_from_library` (or an option on `edit_component`) that:

1. Loads the footprint from the library using `pcbnew.FootprintLoad(library_path, fp_name)`
2. Copies the loaded footprint's pads, courtyard, silkscreen, etc. to the existing component
3. Preserves the component's position, rotation, and reference/value fields
4. Optionally: also updates the footprint ID if a new `footprint` string is given

This mirrors what KiCad does internally in `PCB_EDIT_FRAME::ExchangeFootprint()`.

## Workaround

Currently the only workaround is to open KiCad PCB Editor in GUI, select the affected components, right-click → *Update Footprint from Library*, save, and then reload the board in the MCP session before re-routing. This breaks fully automated PCB workflows.

## Note

`update_symbol_from_library` already exists for **schematic symbols** — a PCB counterpart would complete the parity.

## Environment

- KiCad 10.0
- Windows 11
- KiCAD-MCP-Server (current main)

AmirF194:
Did it, following the shape you laid out. `edit_component`'s footprint branch only called `SetFPID`, so the pads, courtyard and silkscreen stayed whatever the old footprint had, matching what you and the reporter both described.

Loads the new footprint from the library and swaps it in for the old one: reference, value, position and orientation carry over, and each new pad picks up the net of the old pad with the same number, so a resistor keeps its connections when it becomes a bigger resistor. A pad number with no match on the old footprint (your 2-pad-to-3-pad example) is left unset rather than guessed at.

Tests mock `pcbnew` the way the rest of this suite does, since there's no real board here to check DRC against. `test_edit_component_footprint_exchange.py` covers the swap itself, net reassignment by pad number, the unmatched-pad case, the bare-footprint-name path that keeps the existing library, and the flip-after-add ordering (matches the existing `place_component` comment about `Flip()` needing board context).

Fixes #399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,15 @@ All notable changes to the KiCAD MCP Server project are documented here.
 
 ### Bug Fixes
 
+- **`edit_component`'s footprint swap now actually replaces the footprint** (#399,
+  reported by @mixelpixx). Passing a new `footprint` rewrote the FPID library-ID
+  string via `SetFPID` and stopped there, so the pads, courtyard and silkscreen
+  stayed whatever the old footprint had. KiCad then reports `lib_footprint_mismatch`
+  plus unconnected pads once the pad counts differ. The handler now loads the new
+  footprint from the library and exchanges it in place, matching KiCad's own
+  `PCB_EDIT_FRAME::ExchangeFootprint()`: reference, value, position and orientation
+  carry over, and each new pad picks up the net of the old pad with the same
+  number.
 - **A missing kicad-skip no longer kills every tool at startup** (#389, @AmirF194).
   Six modules imported `from skip import Schematic` at their own top level.
   Two of them sit on the import chain `kicad_interface` -> `schematic_handlers`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,7 @@ All notable changes to the KiCAD MCP Server project are documented here.
 ### Bug Fixes
 
 - **`edit_component`'s footprint swap now actually replaces the footprint** (#399,
-  reported by @mixelpixx). Passing a new `footprint` rewrote the FPID library-ID
+  reported by @joseluu). Passing a new `footprint` rewrote the FPID library-ID
   string via `SetFPID` and stopped there, so the pads, courtyard and silkscreen
   stayed whatever the old footprint had. KiCad then reports `lib_footprint_mismatch`
   plus unconnected pads once the pad counts differ. The handler now loads the new
--- a/python/commands/component.py
+++ b/python/commands/component.py
@@ -501,18 +501,52 @@ class ComponentCommands(PlacementOptimizerCommands):
             if value:
                 module.SetValue(value)
             if footprint:
-                # For KiCAD 9.x compatibility, use SetFPID instead of SetFootprintName
                 # Parse footprint string (format: "Library:Footprint")
                 if ":" in footprint:
                     lib_name, fp_name = footprint.split(":", 1)
-                    fpid = pcbnew.LIB_ID(lib_name, fp_name)
-                    module.SetFPID(fpid)
                 else:
                     # If no library specified, keep existing library
                     current_fpid = module.GetFPID()
                     lib_name = current_fpid.GetLibNickname().GetUTF8()
-                    fpid = pcbnew.LIB_ID(lib_name, footprint)
-                    module.SetFPID(fpid)
+                    fp_name = footprint
+
+                library_path = self.library_manager.get_library_path(lib_name)
+                if not library_path:
+                    return {
+                        "success": False,
+                        "message": "Library not found",
+                        "errorDetails": f"Could not find footprint library: {lib_name}",
+                    }
+
+                new_module = pcbnew.FootprintLoad(library_path, fp_name)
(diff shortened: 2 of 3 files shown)


mixelpixx (changes requested):
Reviewed, and this is the right shape: load from the library, carry reference, value, position and orientation, reconnect pads by number, flip only after `board.Add` (the KiCad 9 ordering the tests pin). CI is green across the KiCad 8/9/10 integration jobs, and the full suite on the merged tree locally shows no change to the known failure set. Two things before merge, both small:

1. **CHANGELOG attribution.** The entry says "reported by @mixelpixx"; #399 was filed by @joseluu, I only triaged it. Please credit them.

2. **Bare footprint name with no usable old nickname.** The new branch resolves a bare name by reusing `module.GetFPID().GetLibNickname()`. When that nickname is empty (a footprint placed without a library, or an imported board) `get_library_path("")` returns None and the call fails with "Library not found" even though the footprint exists in the tables. `place_component` already solves this with `self.library_manager.find_footprint(name)` (component.py:57), which searches every library and returns the path and name; falling back to it when the nickname is empty, or when `get_library_path` returns None, keeps the two tools consistent. A test for that path would round it out.

Three carry-overs that KiCad's own `PCB_EDIT_FRAME::ExchangeFootprint` does and this branch does not. I would take them here or as a follow-up issue I file, your choice:

- `new_module.SetPath(module.GetPath())`. The KIID_PATH is what links a footprint to its schematic symbol; without it, "Update PCB from Schematic" matching by UUID no longer recognises the exchanged footprint and places a new one next to it.
- `new_module.SetAttributes(module.GetAttributes())`. Keeps DNP and the exclude-from-BOM / exclude-from-POS flags; `set_footprint_type` (component.py:581-601) already uses SetAttributes, so the SWIG call is known to work on 8/9/10.
- `new_module.SetLocked(module.IsLocked())`.

Fields beyond reference and value, and group membership, are further out and can wait.

One note, not blocking: the test module sets `side_effect` and `return_value` on the shared session-wide `pcbnew` MagicMock (tests/conftest.py installs one for the whole session), and `reset_mock()` does not clear those, so `LIB_ID.side_effect` outlives each test. The suite passes in either order today, so this is hygiene rather than a defect; `monkeypatch.setattr` on the two attributes would contain it.

This closes the edit_component half of #399. The bulk "update every footprint from its library" tool the issue also asks for stays open, and with this branch it becomes a loop over `board.GetFootprints()` through the same path.

AmirF194:
Both fixed. Credited @joseluu in the CHANGELOG entry, and the bare-name path now falls back to find_footprint() when the nickname is empty or doesn't resolve, same as place_component, with a test for it.

Also took the three carry-overs: SetPath, SetAttributes and SetLocked now run alongside the existing reference/value/position/orientation copy, plus a test asserting all three. Left the MagicMock hygiene note for later since it's not causing a failure today.

Pushed to the same branch (c803bb3). New test count is 10/10 for this file, full unit suite still 839/839.

mixelpixx:
Merged as ac716d1. Thanks for the fast turnaround and for taking the carry-overs: SetPath in particular is what keeps Update PCB from Schematic matching the swapped footprint by UUID instead of placing a second one. Verified on the merged tree (main with #391 and this): 2368 passed, the only failures being the four environmental ones on this machine; CI on the PR head was green on all checks including KiCad 8, 9 and 10. The other half of #399, a bulk update over every footprint sharing an FPID, stays open.

AmirF194:
Good to know SetPath is the one that actually matters for Update PCB from Schematic, that's useful context I didn't have. Glad it's holding up on the merged tree.

PR #416, opened 2026-09-11
fix(windows): include KiCad 10 in UI launcher and footprint library defaults

DieterMayerOSS:
## Problem

On Windows with KiCad 9 and KiCad 10 installed side by side, two code paths still hard-code only the 9.0/8.0 install locations:

- `KiCADProcessManager.get_executable_path()` (`python/utils/kicad_process.py`) never checks `C:\Program Files\KiCad\10.0`, so `launch_kicad_ui` opens the **KiCad 9** GUI even though the server itself is running on the KiCad 10 bundled Python (the TypeScript-side detection in `src/server.ts` already sorts versions descending and correctly picks 10.0).
- `list_footprint_libraries()` (`python/commands/footprint.py`) only scans the 9.0/8.0 `share\kicad\footprints` directories, so footprint discovery reads the old 9.0 libraries.

## Fix

Prepend the 10.0 locations to both fallback lists (Program Files, Program Files (x86), and the per-user `Documents\KiCad\10.0\footprints` path), keeping the existing 9.0/8.0 entries as fallbacks.

Verified locally on Windows 11 with KiCad 9.0 and 10.0.4 installed: `get_executable_path()` now resolves `C:\Program Files\KiCad\10.0\bin\pcbnew.exe`.

### DIFF
--- a/python/commands/footprint.py
+++ b/python/commands/footprint.py
@@ -504,10 +504,12 @@ class FootprintCreator:
     def list_footprint_libraries(self, search_paths: Optional[List[str]] = None) -> Dict[str, Any]:
         """List all .pretty libraries and their footprints."""
         default_paths = [
+            r"C:\Program Files\KiCad\10.0\share\kicad\footprints",
             r"C:\Program Files\KiCad\9.0\share\kicad\footprints",
             r"C:\Program Files\KiCad\8.0\share\kicad\footprints",
             "/usr/share/kicad/footprints",
             "/usr/local/share/kicad/footprints",
+            os.path.expanduser("~/Documents/KiCad/10.0/footprints"),
             os.path.expanduser("~/Documents/KiCad/9.0/footprints"),
         ]
         paths = search_paths or default_paths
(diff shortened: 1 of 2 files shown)


mixelpixx:
Merged as fbbf325, thank you. Your diagnosis was right for Program Files installs. Two things were added in the merge:

- Both sites now consult `utils.kicad_roots.kicad_install_roots()` before the fixed paths. That is the shared discovery (registry, Program Files, and custom roots such as `C:\KiCad\10.0`) that #286 introduced for the other modules. On a machine with KiCad under a custom root, like the one this was reviewed on, neither fixed list matched at all.
- `list_footprint_libraries` assigns by library name, so a later directory replaces an earlier one. With 10.0 prepended, the 9.0 libraries were scanned after it and still won every name clash. The list now runs oldest version first, with per-user libraries last, so the newest install wins and a user's own library still overrides both.

`tests/test_windows_kicad10_discovery.py` pins both, and the PR also gained a CHANGELOG entry. While reviewing this I found three more places that only know the 9.0 configuration directory: global library registration, the PCM third-party variable, and the cairo DLL preload. They are filed as #425.

PR #417, opened 2026-09-14
fix(freerouting): report failed SES imports instead of success; import boards with duplicate refs

outstanda:
## Summary

`autoroute` and `import_ses` can answer `success: true` while nothing was imported.

`pcbnew.ImportSpecctraSES` / `ExportSpecctraDSN` return `False` on failure. The four call
sites checked `result is not True and result != 0`, and since `False == 0` in Python a
failed call passed the check. The tool reported success, gave a track count, and the board
was unchanged.

The failure this was hiding in my case: `ImportSpecctraSES` **aborts the whole import** when
the SES `(placement ...)` block names a component reference that is not on the board.
Freerouting always writes that block, and the DSN export renames duplicate references —
e.g. several `REF**` footprints become `REF**_1`, `REF**_2`... — so any board with
non-unique references (legacy/imported boards, footprints placed without annotation) could
not be imported at all. Verified on KiCad 10.0.6: the same SES returns `False` with the
placement block, and without it imports normally (board tracks 166 → 418).

- `_api_ok(result)` — success only for `True` (or `0` for builds that return an int). Used by
  all four export/import checks.
- `_strip_ses_placement(ses_text)` — removes the `(placement ...)` block (paren-matched).
  Freerouting never moves components, so the block carries nothing the import needs.
  `import_ses` applies it together with the existing #246 net-name reconciliation (same temp
  file, same fall-back-to-original on any error; the "Reconciled N net names" log now only
  fires when names were actually remapped). `autoroute` applies it to its staged SES before
  import.

## Tests

`tests/test_ses_import_failures.py`:
- `_api_ok` accepts `True`/`0`, rejects `False`, `None`, `1`, a `MagicMock`
- `_strip_ses_placement` removes only the placement block, keeps routes and paren balance,
  no-op without the block
- `import_ses` returns `success: False` when pcbnew returns `False`
- `import_ses` hands pcbnew a file without the placement block and leaves the caller's SES
  untouched

Full Python suite: 2375 passed, 38 skipped. `black --check`, `mypy`, `flake8`, `isort` clean on
the touched files.

Note: this PR and its two siblings (Java detection, hole edge clearance) each add an entry at
the top of `CHANGELOG.md` → Bug Fixes, so the second and third merge will need a trivial
changelog conflict resolution.

Docs: `docs/FREEROUTING_GUIDE.md` (`import_ses`) notes that the placement block is skipped and failed imports return `success: false`.

## CI run locally

- **python-tests:** `pytest --cov=python` on Python **3.9, 3.10, 3.11 and 3.12** — 2375 passed, 38 skipped on each; `black --check python/` and `mypy python/` on 3.12 — clean


outstanda:
@mixelpixx ready for review when you get to it. This is one of three independent Freerouting fixes (with #418 and #419); they merge cleanly in any order, only the CHANGELOG entries overlap.

mixelpixx:
Merged as 7c5b15f, thank you. The core diagnosis was exact. On KiCad 10.0.5, `ImportSpecctraSES` returns a plain `False` on failure, and the old check let it through. It was worse than described: a failed import has already removed the board's unlocked tracks in memory, and main then saved that board and reported success.

One behaviour change went into the merge. `import_ses` no longer strips the whole placement block. KiCad does apply SES placement: an entry that moves and flips R1 lands it at the SES position on B.Cu. Freerouting's GUI can move parts, so stripping would silently discard those moves while importing routes drawn for them. `import_ses` now drops only the entries whose reference is missing from the board or duplicated on it, and lists them in `placementSkipped`. `autoroute` keeps your full strip, because its headless run never moves parts, but it now works on a copy so a `keepArtifacts` run keeps the SES Freerouting wrote.

The block matcher now ignores parentheses inside quoted names. Both failure messages now say that nothing was saved and that the board should be reopened before saving.

On the cause: on KiCad 10.0.5, `ExportSpecctraDSN` itself fails for a board with two `REF**` footprints, so a renamed `REF**_1` never reaches an SES there. The CHANGELOG attributes the renaming to your 10.0.6. Verified with the full suite on the merged tree and CI on the three PRs.

PR #418, opened 2026-09-14
fix(freerouting): require the Java version the JAR was built for; find JDKs on macOS

outstanda:
## Summary

Two ways `autoroute` fails to find a usable Java:

1. **Wrong version requirement.** The runtime check accepts any Java 21+, but current
   Freerouting releases need more: `freerouting-2.4.1.jar` is compiled for **Java 25**
   (class-file major 69). With Java 21 installed, `check_freerouting` reports
   `ready: true`, and every run then fails with `UnsupportedClassVersionError ... class
   file version 69.0, this version of the Java Runtime only recognizes class file versions
   up to 65.0`. The Docker fallback image (`eclipse-temurin:21-jre`) has the same Java 21 and cannot load the JAR either.
2. **macOS lookup stops at an unusable `java`.** `_find_java` returned the first hit of
   `shutil.which("java")`. On macOS `/usr/bin/java` is a stub that exists even with no JRE
   installed ("Unable to locate a Java Runtime"), and Homebrew's `openjdk` is keg-only (not
   on `PATH`), so a correctly installed JDK was never picked up.

- `_jar_required_java(jar_path)` reads the JAR's `Main-Class` from the manifest and returns
  `class major - 44` (never below 21; falls back to 21 if the JAR can't be read).
- `_java_version_ok(java_exe, required=21)` compares against that requirement.
- `_find_java(required=21)` tries `JAVA_HOME`, `PATH`, Homebrew `openjdk`
  (`/opt/homebrew`, `/usr/local`), `/usr/bin/java`, `/usr/local/bin/java`, and returns the
  first that is new enough — or the first one found, so the error can say "found but too old".
- `_resolve_execution_mode`, `_build_freerouting_cmd` and `check_freerouting` use the JAR's
  requirement; the Docker image becomes `eclipse-temurin:<required>-jre`; error messages name
  the actual version.
- `check_freerouting` adds `java.required_version` and `java.version_ok`. `java_21_ok` is kept
  for existing callers and now means "meets the JAR's requirement".
- The unused `DOCKER_IMAGE` constant is removed; `docs/FREEROUTING_GUIDE.md` and the
  `autoroute` tool description (which still said "Java 11+") are updated.

Checked against the real `freerouting-2.4.1.jar` on macOS (Homebrew openjdk 26, no system
JRE): `required_version: 25`, Java found at `/opt/homebrew/opt/openjdk/bin/java`,
`execution_mode: direct`, Docker image `eclipse-temurin:25-jre`.

## Tests

`tests/test_freerouting.py`:
- `_jar_required_java` on generated JARs (major 69 → 25, 65 → 21), missing/garbage JAR → 21
- `_docker_image` follows the requirement (never below 21)
- `_resolve_execution_mode` error names "Java 25+" for a 2.4-style JAR
- `_find_java` skips the macOS stub for a working Homebrew JDK, prefers `JAVA_HOME`, falls back
  to the first executable found when none is new enough
- `_java_version_ok` rejects Java 21 when 25 is required
- The autouse `_find_java` monkeypatches in `test_freerouting.py` / `test_autoroute_best_of_n.py`
  now accept the new optional argument.

Full Python suite: 2378 passed, 38 skipped. `black --check`, `mypy`, `flake8`, `isort` clean;
`npm run build` OK, `npm run lint:ts` unchanged (9 pre-existing warnings),
`docs:tools:check` up to date.

Note: this PR and its two siblings each add an entry at the top of `CHANGELOG.md` → Bug Fixes,
so later merges need a trivial changelog conflict resolution.

## CI run locally

- **python-tests:** `pytest --cov=python` on Python **3.9, 3.10, 3.11 and 3.12** — 2378 passed, 38 skipped on each; `black --check python/` and `mypy python/` on 3.12 — clean


outstanda:
@mixelpixx ready for review when you get to it. This is one of three independent Freerouting fixes (with #417 and #419); they merge cleanly in any order, only the CHANGELOG entries overlap.

mixelpixx:
Merged as 551efc0, thank you. Reading the requirement from the JAR is the right approach: the Main-Class major version minus 44, with 21 as the fallback. `java_21_ok` stays compatible for existing callers.

Your CI run failed on all eight Python jobs for a reason outside the logic. `actions/setup-java` exports `JAVA_HOME`, and `test_finds_via_which` did not clear it, so `_find_java` returned the `JAVA_HOME` java first. That is fixed in the merge, along with two Windows problems:

- The `JAVA_HOME` candidate was built as `os.path.join(JAVA_HOME, "bin", "java")`. That path never exists on Windows, where the binary is `java.exe`, even though the new error message tells users to set `JAVA_HOME`. It now goes through `shutil.which`, which applies PATHEXT.
- `test_java_home_is_preferred` hard-coded a POSIX path. It now uses a real file named the way the platform names it.

Also, `check_freerouting` now reports `required_version` as None when there is no JAR to read it from. The guide's pull commands now show `25-jre`.

PR #419, opened 2026-09-14
fix(freerouting): keep autorouted tracks out of the edge clearance around board holes

outstanda:
## Summary

`autoroute` places tracks inside the board's copper-to-edge clearance around mounting holes.

KiCad exports internal Edge.Cuts circles (mounting holes, round cut-outs) to the Specctra DSN
as plain structure keepouts (`(keepout "" (polygon signal 0 ...))`). Freerouting keeps routes
away from those by the ordinary track clearance only — 0.2 mm by default — not by the board's
copper-to-edge clearance (KiCad default 0.5 mm). Routed tracks therefore run ~0.2 mm from
holes, and DRC reports `copper_edge_clearance` for each. On a mounting hole that is a real
manufacturing problem (screw heads and washers short those tracks).

Measured on a 190 × 37 mm 2-layer board with 10 Edge.Cuts holes, KiCad 10.0.6 +
Freerouting 2.4.1:

| run | `copper_edge_clearance` violations |
|---|---|
| current export | 8 (all at holes, 0 at the outer outline) |
| current export + `--router.copper_to_edge_clearance_um=500` | 8 — SES byte-identical, the setting does not apply to these keepouts |
| this PR | **0**, no new clearance violations |

- `_grow_hole_keepouts(dsn_text, grow_um)` enlarges **circular** structure keepouts only
  (≥ 12 distinct vertices, all within 5 % of one radius around the bounding-box centre).
  Rectangles and other shapes are untouched. The circle is polygonised, so its edges sit
  inside its vertices; the new radius is `(r + grow) / cos(π/n)` so every edge clears
  `r + grow`, not just the vertices (without this, DRC still found 0.4985 mm vs 0.5 mm).
- `FreeroutingCommands._apply_edge_clearance(dsn_path)` computes
  `grow = board copper-to-edge clearance − DSN default clearance` and rewrites the DSN.
  It does nothing if `grow <= 0`, if the DSN is not in `(unit um)`, or on any error (logged,
  DSN left as exported).
- Called after the DSN export in both `autoroute` and `export_dsn`.

## Tests

`tests/test_dsn_hole_edge_clearance.py`:
- every polygon edge of a grown hole keepout clears `r + grow` (fails with vertex-only scaling)
- rectangular keepouts and non-positive growth leave the DSN unchanged
- `_apply_edge_clearance` uses board edge clearance minus DSN clearance; non-`um` DSNs are left
  alone

Full Python suite: 2374 passed, 38 skipped. `black --check`, `mypy`, `flake8`, `isort` clean.

Note: this PR and its two siblings each add an entry at the top of `CHANGELOG.md` → Bug Fixes,
so later merges need a trivial changelog conflict resolution.

Docs: `docs/FREEROUTING_GUIDE.md` (`export_dsn`) notes that hole keepouts are enlarged to the edge clearance.

## CI run locally

- **python-tests:** `pytest --cov=python` on Python **3.9, 3.10, 3.11 and 3.12** — 2374 passed, 38 skipped on each; `black --check python/` and `mypy python/` on 3.12 — clean


outstanda:
@mixelpixx ready for review when you get to it. This is one of three independent Freerouting fixes (with #417 and #418); they merge cleanly in any order, only the CHANGELOG entries overlap.

mixelpixx:
Merged as 1e7980a, thank you. I checked it against a real KiCad 10.0.5 DSN. Holes come out as signal keepouts, while the outline is a boundary path and rule areas are layer polygons, so only holes can match. The measured edge distances for a 1.35 mm hole, a 5 mm hole and a hole in a round board came within a few micrometres of the target.

The only additions in the merge are CHANGELOG credit and one caveat: the growth is measured against the default net class's clearance, so a class with a smaller clearance can still come slightly closer. Two refinements, if you want to follow up:

- The 5% roundness test also accepts a square with large corner radii, which is then treated as a circle and comes out about 3 µm short.
- Very small holes come out about 4 µm short, because the vertex count includes KiCad's repeated start vertex.

Sizing the scale from the largest vertex radius and the smallest edge distance would make both exact. The conflict with #417 resolved itself once #417 and #418 were in. Routing now runs in this order: DSN export, then the hole keepouts grow, then Freerouting runs, then the placement block is stripped from a copy, then the SES is imported.

